PR #1, opened 2020-07-08
Issue976

jendrikseipp:
https://git-scm.com/docs/gitignore


.gitignore
```diff
@@ -0,0 +1,12 @@
+/**/*.pyc
```
FlorianPommerening, line 1: Didn't we decide to exclude `/**/__pycache__` instead?

jendrikseipp:
Yes, I changed that now.

.gitignore
```diff
@@ -0,0 +1,12 @@
+/**/__pycache__/
+/**/output.sas
+/**/sas_plan*
```
maltehelmert, line 3: I generally prefer keeping ignores as local as possible to minimize false positives. In this case, I would limit these to the root directory (which contains fast-downward.py) and to ignore sas_plan and sas_plan.[1-9]* rather than all files everywhere whose base name begins with sas_plan.

PatrickFerber, line 3: If we want to match the files/directories anywhere in our repositories, I would skip the leading '/'. Less is more.
**/foo matches any file/directory foo in our repository. 
/**/foo matches any file/directory found when starting to look at the directory root (is the same)

(https://git-scm.com/docs/gitignore)

PatrickFerber: I agree.

.gitignore
```diff
@@ -0,0 +1,12 @@
...
+/builds/
+/experiments/issue[0-9]*/.venv/
+/experiments/issue[0-9]*/data/
+/experiments/issue[0-9]*/*-microbenchmark/.obj/
+/experiments/issue[0-9]*/*-microbenchmark/benchmark
+/experiments/issue[0-9]*/*-microbenchmark/Makefile.depend
+/misc/.tox/
+/misc/autodoc/downward-xmlrpc.secret
+/src/CMakeLists.txt.user
```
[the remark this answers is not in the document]
FlorianPommerening: It is created for example by QtCreator when opening Fast Downward as a project. I would be in favor of keeping the ignore but I could understand treating this as something outside of the main code. If we remove it, I'll try to set up a local ignore file for it.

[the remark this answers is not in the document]
silvansievers: I don't think we should accommodate ignoring files of IDEs.

FlorianPommerening:
By the way, this pull request is targeted at the main repository, not at your fork. In the meeting this morning Silvan suggested to do pull requests targeted at our own forks for now and do the push to the main repository explicitly. Did you want to use this issue to experiment with this?

.gitignore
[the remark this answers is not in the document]
gfrances: I guess you already know, but I usually put those rules in `~/.gitignore`, and then in `~/.gitconfig`:
```
[core]
    excludesfile = ~/.gitignore
```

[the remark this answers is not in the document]
maltehelmert: That's also what the gitignore doc suggests for things like these.

.gitignore
```diff
@@ -0,0 +1,12 @@
+/**/__pycache__/
+/**/output.sas
+/**/sas_plan*
+/builds/
+/experiments/issue[0-9]*/.venv/
+/experiments/issue[0-9]*/data/
+/experiments/issue[0-9]*/*-microbenchmark/.obj/
+/experiments/issue[0-9]*/*-microbenchmark/benchmark
+/experiments/issue[0-9]*/*-microbenchmark/Makefile.depend
```
maltehelmert, line 9: Note that these don't have the same semantics as in our previous ignore file: we used to have zero or more digits. Now we have a digit followed by any characters. That's why I said sas_plan.[1-9]* above rather than sas_plan.[0-9]* because these numbers never begin with a 0, and [0-9]* to me looks like it wants to be a regex, which it isn't. Looking at the gitignore doc, I think there is no way to express "zero or more digits" or anything of the sort in the supported syntax. Given that, I think it's cleaner/more idiomatic not to use a half-baked solution that cares about the first digit but not the others and instead use the not-at-all-baked "issue*"  and "sas_plan.*" instead of the versions with the first-digit range.

Not very important as such, but we also want to establish conventions.

maltehelmert: Note also that it might be nicer to not have these "microbenchmark" rules here and instead have specific .gitignore files only in the relevant issue directories. In hg we didn't have that option. I'm not advocating it necessarily, just mentioning the option. 

I assume the "data" and ".venv" ones perhaps make sense globally, although personally I wouldn't actually want to ignore these files: I want to know if I have "data" directories in my repositories because that's a hint I want to clean them up, not least considering their massive size. Basically, "git status" will tell me which data directories I still have to deal with, I don't want them to hang around in my clone forever. (It's a common impulse to ignore everything that shouldn't be committed, but it's not a dichotomy.) Just a thought, feel free to keep ignoring data and .venv.

FlorianPommerening: I see a large advantage in ignoring `data`: After running an experiment, they will contain a large amount of files and if they are not ignored, git operations (like `git status`) will be much slower and have their output cluttered. I understand that you see this as a reason and reminder to not keep the data around there, but so far, I usually did multiple iterations of experiments during which I wanted to keep the experiment data and still work with the repository.

jendrikseipp: I changed the patterns as you suggest. I agree with Florian about `data`. Maybe it's best not to change too many things now. We can always change them when we have more experience with Git.

jendrikseipp:
> Did you want to use this issue to experiment with this?

I'd say this is a good opportunity to test this.

FlorianPommerening:
> > Did you want to use this issue to experiment with this?
> 
> I'd say this is a good opportunity to test this.

If it works out, please update http://www.fast-downward.org/ForDevelopers/Git or bring up the discussion again (if we need any protection against wrong configurations or commit messages).

.gitignore
maltehelmert: Makes sense regarding `data`.

PR #5, opened 2020-07-10
Issue940

silvansievers:


.github/workflows/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
PatrickFerber, line 1: Was this added by the "act" tool? If yes, remove this file, if not, what was the reason?

.github/workflows/README.md
```diff
@@ -0,0 +1,236 @@
+![](https://github.com/nektos/act/wiki/img/logo-150.png)
```
PatrickFerber, line 1: this again seems to be added by the act tool which we are not using. Please remove it.

.github/workflows/README.md
```diff
@@ -0,0 +1,236 @@
...
+
```
PatrickFerber, line 2: I cannot add a note to a not text file. Please also remove the archive of "act"

.github/workflows/ubuntu-20.04.yml
```diff
@@ -0,0 +1,212 @@
+---
```
PatrickFerber, line 1: When merging your branch with mine, we should make the empty lines consistent. I prefer you style here with an empty line between sequential steps and two empty lines between jobs.

.github/workflows/ubuntu-20.04.yml
```diff
@@ -0,0 +1,212 @@
...
+name: ubuntu-20.04
```
PatrickFerber, line 2: Copy file for ubuntu18.04. adapt compiler versions.
I think I prefer having twice almost the same file, than having twice the same content in a single file. 
If there is a nice way to define different compiler/python versions for two runs-on, this could be a solution, but for now, I would keep it as it is.

.github/workflows/ubuntu-20.04.yml
```diff
@@ -0,0 +1,212 @@
...
+          DOWNWARD_COIN_ROOT: /home/runner/work/downward/lib/coin
+        run: |
+          export PATH="$(pwd):$PATH"  # Add uncrustify and VAL to PATH.
+          cd misc/
+          python_version=${{ matrix.python-version }}
+          if [[ $python_version=="3.6" ]]; then
+            tox -e py36,translator,search,style
+          elif [[ $python_version=="3.8" ]]; then
+            tox -e py38,translator,search,style
+          else
```
PatrickFerber, line 207: We could remove the style test from one of the versions. The results should be the same.

.github/workflows/ubuntu-20.04.yml
jendrikseipp: I think we should have a single workflow file for both Ubuntu versions. Otherwise, it will be tedious to keep them in sync. GitHub Actions are quite flexible: e.g., you can execute part of the steps only for specific OS-compiler pairs.

PatrickFerber: Ubuntu is now in a single file with exclusion.

PR #8, opened 2020-07-17
Issue964

silvansievers:


src/search/utils/logging.h
```diff
@@ -4,36 +4,51 @@
...
+  stdout.
 
-  Usage:
-        utils::g_log << "States: " << num_states << endl;
+  Internal class encapsulated by LogProxy.
 */
 class Log {
-private:
-    bool line_has_started = false;
+    std::ostream &stream;
```
PatrickFerber, line 34: Looks a bit strange that we store a reference to a stream, but hard code it to std::cout. I know we want to have configurable streams in the future, but who knows when we do this and adding then the stream options would not take a lot of effort.
Thus, I am uncertain, if we should simplify the code to just use std::cout as before.

src/search/merge_and_shrink/merge_scoring_function_dfp.cc
```diff
@@ -130,10 +134,13 @@ static shared_ptr<MergeScoringFunction>_parse(options::OptionParser &parser) {
         "greedy=false),label_reduction=exact(before_shrinking=true,"
         "before_merging=false),max_states=50000,threshold_before_merge=1)\n}}}");
 
+    add_merge_scoring_function_options_to_parser(parser);
+
```
PatrickFerber, line 138: Here you add a configurable verbosity (via the logger) to classes which did not previosuly have it. Just because it would be good here or is there a reason why this is needed in the current patch, e.g. you previosuly just passed Verbosity::Silent to a function, but now would need to create a new LogProxy object for each function call.

PatrickFerber: from below this seems to be exactly the case. Can you still confirm, please.

src/search/merge_and_shrink/merge_selector.h
```diff
@@ -1,22 +1,31 @@
...
 namespace merge_and_shrink {
 class FactoredTransitionSystem;
+
 class MergeSelector {
 protected:
+    mutable utils::LogProxy log;
+
     virtual std::string name() const = 0;
-    virtual void dump_specific_options() const {}
+    virtual void dump_specific_options() const = 0;
```
PatrickFerber, line 23: Why is this changed?

src/search/merge_and_shrink/merge_selector.h
```diff
@@ -1,22 +1,31 @@
...
     std::vector<std::pair<int, int>> compute_merge_candidates(
         const FactoredTransitionSystem &fts,
         const std::vector<int> &indices_subset) const;
 public:
-    MergeSelector() = default;
+    explicit MergeSelector(const options::Options &options);
```
PatrickFerber, line 28: here you add new functionality (configuring the mergeselector), don't you?
If yes, I accept, just want to understand that nothing strange happens here.

src/search/merge_and_shrink/merge_selector.h
PatrickFerber: I cannot comment on unchanged lines. In the concrete implementation of dump_specific_options(), why did you nto forward the LogProxy to the called dump functions? Even if you want to log their output always, if later the stream can be changed, we would overlook changing those calls.

src/search/merge_and_shrink/merge_tree_factory.h
```diff
@@ -25,8 +27,10 @@ class MergeTreeFactory {
 protected:
     std::shared_ptr<utils::RandomNumberGenerator> rng;
     UpdateOption update_option;
+    mutable utils::LogProxy log;
+
     virtual std::string name() const = 0;
-    virtual void dump_tree_specific_options() const {}
+    virtual void dump_tree_specific_options() const = 0;
```
PatrickFerber, line 33: again, this is unrelated to the issue, isn't it?

src/search/merge_and_shrink/shrink_strategy.h
```diff
@@ -6,14 +6,18 @@
...
 
+namespace utils {
+class LogProxy;
+}
+
 namespace merge_and_shrink {
 class Distances;
 class TransitionSystem;
 
 class ShrinkStrategy {
```
PatrickFerber, line 17: From the previous modifications, why did you not add the log options for the parser for the ShrinkStrategies?

src/search/merge_and_shrink/merge_scoring_function_dfp.cc
silvansievers: Yes, this is unrelated to this issue. I don't remember why I did this. Probably to have some kind of test case, I guess. I could also revert the changes done in M&S.

src/search/merge_and_shrink/merge_selector.h
silvansievers: Yes, see previous reply.

src/search/merge_and_shrink/merge_selector.h
silvansievers: I'm not sure I understand your comment regarding "later changing this". I think the reason why I gave the ScoringFunction and MergeSelector classes their own log instances is to avoid using hard-coded "silent loggers" as it was previously done in MergeScoringFunctionMIASM. I thought it would be better if they could be configured separately instead. But I'm not convinced if that is indeed the best solution... And you are right again that this technically doesn't belong into this issue. Should I revert these two changed parts?

PatrickFerber: My comment about unchanged lines was regarding the overridden version MergeSelectorScoreBasedFiltering::dump_specific_options.

You added a LogProxy to MergeSelectorScoreBasedFiltering, but then call another dump function without passing the logger in MergeSelectorScoreBasedFiltering::dump_specific_options.

PatrickFerber: The issue is to provide the option to use local loggers. Your changes are more than required, thus, I would revert them.

Do you intend to do in the near future more changes to the logging in M&S (e.g. add the option in more place or use the log level in more outputs)? If yes, it would be easy to move your changes from here to that new issue.

src/search/merge_and_shrink/shrink_strategy.h
silvansievers: I think because with the exception of a single line in the bucket-based strategy, shrink strategies don't log anything. Every component of M&S prints their options at the start, but I'm not sure if that justifies adding their logger. Maybe it actually does.

src/search/state_registry.cc
```diff
@@ -103,7 +103,7 @@ int StateRegistry::get_state_size_in_bytes() const {
     return get_bins_per_state() * sizeof(PackedStateBin);
 }
 
-void StateRegistry::print_statistics() const {
-    utils::g_log << "Number of registered states: " << size() << endl;
-    registered_states.print_statistics();
+void StateRegistry::print_statistics(utils::LogProxy &log) const {
+    log << "Number of registered states: " << size() << endl;
+    registered_states.print_statistics(log);
```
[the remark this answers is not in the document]
silvansievers: I don't understand what you mean - which hash set? you mean the registered states aka. closed list (and therefore hash set)?

src/search/merge_and_shrink/merge_selector.h
silvansievers: On the one hand, yes, but thinking about it a bit more, I think it is important that the two parts of the code which already use the verbosity option now switch to the new mechanism. Otherwise, we don't see how the new mechanism works. This is why I changed the search engines and merge-and-shrink to use the new mechanism.

src/search/state_registry.cc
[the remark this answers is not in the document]
PatrickFerber: exactly. The statistics of the registered_states (which is a hashset) might not be too interesting in general. (but by the description of the issue, we might decide to ignore this).

src/search/merge_and_shrink/shrink_strategy.h
PatrickFerber: For the strategy which logs, is that logging required or could be skip it?

src/search/merge_and_shrink/merge_selector.h
PatrickFerber: If we keep the code, then I suggest to use the local logger from MergeSelectorScoreBasedFiltering also in the `dump_options` calls written in 
```void MergeSelectorScoreBasedFiltering::dump_specific_options() const {
    for (const shared_ptr<MergeScoringFunction> &scoring_function
         : merge_scoring_functions) {
        scoring_function->dump_options();
    }
}```

src/search/merge_and_shrink/merge_scoring_function_dfp.cc
silvansievers: As discussed, we decided to make the changes less invasive and replace the currently hard-coded silent verbosity of the MIASM scoring function to a hard-coded silent logger.

src/search/merge_and_shrink/shrink_strategy.h
silvansievers: It always does it as it is some kind of warning. 

I had another look at the thing and I think that not giving shrink strategies their own logger is good because they don't pre-compute anything. They are basically stateless objects, which means whoever wants to use them should decide from the outside if they want to log or not. The same is not true for the merge strategies: here, only the *factories* have a logger because they might do some pre-computation and log there. But the merge strategies themselves don't have a log and do not log at all. This also very nicely aligns with our decision to not pass or have loggers in merge selectors or merge scoring functions, which are all stateless objects that can be used in different contexts.

src/search/state_registry.cc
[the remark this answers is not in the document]
silvansievers: I think this is unrelated to this issue.

src/search/merge_and_shrink/shrink_strategy.h
PatrickFerber: ok

PatrickFerber (approved):
I went again quickly over the Diff and it looks good

PR #11, opened 2020-07-24
Issue927

silvansievers:


src/search/algorithms/equivalence_relation.h
```diff
@@ -27,120 +30,61 @@ class Block {
...
     BlockListIter it_intersection_block;
 public:
-    bool empty() const;
-    ElementListIter insert(int element);
-    void erase(ElementListIter it);
-    ElementListIter begin() {return elements.begin();}
-    ElementListIter end() {return elements.end();}
-    ElementListConstIter begin() const {return elements.begin();}
-    ElementListConstIter end() const {return elements.end();}
+    bool empty() const {
```
FlorianPommerening, line 33: I think we inlined these methods when we used a templated block class. Now that the class is no longer templated, I'd move the implementations back to the cc file unless you see a significant hit to performance.

src/search/algorithms/equivalence_relation.h
```diff
@@ -27,120 +30,61 @@ class Block {
...
+    ElementListConstIter begin() const {
+        return elements.begin();
+    }
+
+    ElementListConstIter end() const {
+        return elements.end();
+    }
 };
 
+
```
FlorianPommerening, line 62: remove empty line again?

src/search/merge_and_shrink/merge_and_shrink_algorithm.cc
```diff
@@ -84,6 +84,9 @@ void MergeAndShrinkAlgorithm::dump_options() const {
                      << shrink_threshold_before_merge << endl;
         utils::g_log << endl;
 
+        shrink_strategy->dump_options();
```
FlorianPommerening, line 87: Is this output short enough to remain here long-term?

src/search/merge_and_shrink/labels.h
```diff
@@ -1,44 +1,63 @@
 #ifndef MERGE_AND_SHRINK_LABELS_H
 #define MERGE_AND_SHRINK_LABELS_H
 
+// TODO: rename this file to global_labels.h
```
FlorianPommerening, line 4: I'd do this now. preserving the diff for the review would be a reason to postpone this but in this case the classes in this file changed so much that the diff isn't useful anyway.

src/search/merge_and_shrink/labels.h
```diff
@@ -1,44 +1,63 @@
...
-    int max_size; // the maximum number of labels that can be created
+class GlobalLabels {
+    std::vector<int> label_costs;
+    int max_num_labels; // the maximum number of labels that can be created
 public:
-    explicit Labels(std::vector<std::unique_ptr<Label>> &&labels);
-    void reduce_labels(const std::vector<int> &old_label_nos);
-    bool is_current_label(int label_no) const;
-    int get_label_cost(int label_no) const;
+    GlobalLabels(std::vector<int> &&label_costs, int max_num_labels);
```
FlorianPommerening, line 40: `max_num_labels` sounds like something that should be the responsibility of this class rather than something that is passed in as a parameter. Can you compute it in the constructor based on the size of `label_costs`?

FlorianPommerening (commented):
I had a few minutes to start the review. I'm not done yet but thought, I'd submit what I have so far.

src/search/merge_and_shrink/fts_factory.cc
```diff
@@ -302,52 +307,72 @@ void FTSFactory::build_transitions_for_operator(OperatorProxy op) {
...
-    int num_labels = task_proxy.get_operators().size();
 
     // Collect all irrelevant labels for this variable.
-    vector<int> irrelevant_labels;
-    for (int label_no = 0; label_no < num_labels; ++label_no) {
-        if (!is_relevant(var_no, label_no)) {
-            irrelevant_labels.push_back(label_no);
+    LabelGroup irrelevant_labels;
+    int cost = INF;
+    for (int label : labels) {
```
FlorianPommerening, line 351: rename to `global_label`?

Also, I think the code would be more readable if you separate the 2 loops. The loop to compute the cost could just loop over `irrelevant_labels` after the first loop is done. You could even use `std::min_element` instead of the second loop.

FlorianPommerening: Speaking of `std::min_element`: I recently found this and wasn't aware of it earlier. I think it could be used in a lot of places to simplify things but I find it a bit annoying that the call always gets so long because you pass in iterators. We could add overloads to the namespace `algorithm` that take a collection and just call `begin()` and `end()` on it.

src/search/merge_and_shrink/fts_factory.cc
```diff
@@ -420,9 +443,9 @@ vector<unique_ptr<Distances>> FTSFactory::create_distances(
     assert(num_variables >= 1);
     result.reserve(num_variables * 2 - 1);
 
-    for (int var_no = 0; var_no < num_variables; ++var_no) {
+    for (int var_id = 0; var_id < num_variables; ++var_id) {
```
FlorianPommerening, line 446: rename to `var`?

src/search/merge_and_shrink/fts_factory.cc
```diff
@@ -402,10 +425,10 @@ vector<unique_ptr<MergeAndShrinkRepresentation>> FTSFactory::create_mas_represen
     assert(num_variables >= 1);
     result.reserve(num_variables * 2 - 1);
 
-    for (int var_no = 0; var_no < num_variables; ++var_no) {
-        int range = task_proxy.get_variables()[var_no].get_domain_size();
+    for (int var_id = 0; var_id < num_variables; ++var_id) {
```
FlorianPommerening, line 428: rename to `var`?

src/search/merge_and_shrink/fts_factory.cc
```diff
@@ -373,18 +398,16 @@ vector<unique_ptr<TransitionSystem>> FTSFactory::create_transition_systems(const
...
     result.reserve(num_variables * 2 - 1);
 
-    for (int var_no = 0; var_no < num_variables; ++var_no) {
-        TransitionSystemData &ts_data = transition_system_data_by_var[var_no];
-        /* Construct the label equivalence relation from the previously
-           computed label groups. */
-        ts_data.label_equivalence_relation =
-            utils::make_unique_ptr<LabelEquivalenceRelation>(
-                labels, ts_data.label_groups);
+    for (int var_id = 0; var_id < num_variables; ++var_id) {
```
FlorianPommerening, line 401: rename to `var`?

FlorianPommerening (commented):
I ran out of time for today, so the review is not complete yet but I left some comments.

src/search/merge_and_shrink/transition_system.cc
```diff
@@ -328,63 +396,67 @@ void TransitionSystem::apply_abstraction(
...
-      storing these affected groups). Furthermore, rather than computing
-      locally equivalent labels from scratch, we did not per default add a new
-      group for every label, but checked for an existing equivalent label
-      group. In issue539, it turned out that this incremental fashion of
-      computation does not accelerate the computation.
+      local labels, and the local labels themselves if do not represent any
+      global labels anymore. Also collect the transitions of all reduced
+      labels. Add a new local label for every new label and assign the
+      collected transitions to it. Recompute the cost of all affected local
+      labels and re-compute locally equivalent labels.
```
salome-eriksson, line 420: I first misunderstood this comment in regards to what happens for equivalent labels. I didn't understand that we add the new global label to the same label group that the old global labels were in, which in turn means that the affected local label cannot become empty. I'm not sure if the comment can be improved to make this clearer or if it's just because I'm not familiar enough with the details of M&S ;).

src/search/merge_and_shrink/transition_system.cc
```diff
@@ -328,63 +396,67 @@ void TransitionSystem::apply_abstraction(
...
         vector<vector<Transition>> new_transitions;
         new_transitions.reserve(label_mapping.size());
-        unordered_set<int> affected_group_ids;
+        unordered_set<int> affected_local_labels;
         for (const pair<int, vector<int>> &mapping: label_mapping) {
-            const vector<int> &old_label_nos = mapping.second;
-            assert(old_label_nos.size() >= 2);
-            unordered_set<int> seen_group_ids;
+            const vector<int> &old_labels = mapping.second;
+            assert(old_labels.size() >= 2);
```
salome-eriksson, line 452: Why do we have this assert here but not in the only_equivalent_labels case?

src/search/merge_and_shrink/transition_system.cc
```diff
@@ -328,63 +396,67 @@ void TransitionSystem::apply_abstraction(
...
+            unordered_set<int> seen_local_labels;
             set<Transition> new_label_transitions;
```
salome-eriksson, line 454: I assume you use set here because you want the transitions ordered. Could it be more efficient to use an unordered_set and sort it once at the end? (I have absolutely no intuition)

src/search/merge_and_shrink/transition_system.cc
```diff
@@ -394,42 +466,68 @@ void TransitionSystem::apply_label_reduction(
         assert(label_mapping.size() == new_transitions.size());
 
         /*
-           Apply all label mappings to label_equivalence_relation. This needs
+           Apply all label mappings to the label mappings. This needs
```
salome-eriksson, line 469: That sentence sounds weird now :).

src/search/merge_and_shrink/transition_system.cc
```diff
@@ -394,42 +466,68 @@ void TransitionSystem::apply_label_reduction(
...
-        // Go over all affected group IDs and remove their transitions if the
-        // group is empty.
-        for (int group_id : affected_group_ids) {
-            if (label_equivalence_relation->is_empty_group(group_id)) {
-                utils::release_vector_memory(transitions_by_group_id[group_id]);
+        // Finally remove transitions of all local labels that do not represent
+        // any global label anymore.
+        for (int local_label : affected_local_labels) {
+            if (local_to_global_labels[local_label].empty()) {
+                utils::release_vector_memory(local_label_to_transitions[local_label]);
```
salome-eriksson, line 523: Could we also do that in the loop that recomputes the cost of all affected local labels? Not sure if we'd want that though...

src/search/algorithms/equivalence_relation.h
```diff
@@ -1,20 +1,23 @@
...
-typedef std::list<int>::iterator ElementListIter;
-typedef std::list<int>::const_iterator ElementListConstIter;
-typedef std::list<Block>::iterator BlockListIter;
-typedef std::list<Block>::const_iterator BlockListConstIter;
 class EquivalenceRelation;
 
+using ElementListIter = std::list<int>::iterator;
+using ElementListConstIter = std::list<int>::const_iterator;
+using BlockListIter = typename std::list<Block>::iterator;
+using BlockListConstIter = typename std::list<Block>::const_iterator;
```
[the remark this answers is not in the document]
FlorianPommerening: I think it is because the type `const_iterator` is nested in `std::list<Block>`.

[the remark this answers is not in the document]
maltehelmert: I don't think that would explain why is it not needed for the first two definitions, which are the same as the last two with `int` instead of `Block`.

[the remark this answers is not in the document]
silvansievers: I think this is a left-over of the intermediate stage where we actually templated the whole class. I assume it can go (again).

src/search/algorithms/equivalence_relation.h
silvansievers: They were inlined in the old implementation (and never in the implementation file), we just didn't touch this. Since refining equivalence relations indeed can be a bottle neck, as we found out, I would actually leave it like this (or make it one-liners again if you prefer).

src/search/merge_and_shrink/factored_transition_system.cc
```diff
@@ -97,8 +97,7 @@ bool FactoredTransitionSystem::is_component_valid(int index) const {
     if (compute_goal_distances && !distances[index]->are_goal_distances_computed()) {
         return false;
     }
-    return transition_systems[index]->are_transitions_sorted_unique() &&
-           transition_systems[index]->in_sync_with_label_equivalence_relation();
+    return transition_systems[index]->is_valid();
```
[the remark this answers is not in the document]
silvansievers: Previously, we would always assert `are_transitions_sorted_unique` and `in_sync_with_label_equivalence_relation`. The latter is now "replaced" by asserting that the label mappings make sense. The other test remained.

src/search/merge_and_shrink/fts_factory.cc
```diff
@@ -102,17 +105,19 @@ FTSFactory::FTSFactory(const TaskProxy &task_proxy)
 FTSFactory::~FTSFactory() {
 }
 
-vector<unique_ptr<Label>> FTSFactory::create_labels() {
-    vector<unique_ptr<Label>> result;
-    int num_ops = task_proxy.get_operators().size();
+unique_ptr<GlobalLabels> FTSFactory::create_labels() {
```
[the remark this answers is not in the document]
silvansievers: I think the intention is to construct everything in a one-shot way. The same logic applies to `TransitionSystem` etc., which we construct here and move there. I wouldn't change this.

src/search/merge_and_shrink/fts_factory.cc
```diff
@@ -141,26 +146,25 @@ void FTSFactory::build_state_data(VariableProxy var) {
     }
 }
 
-void FTSFactory::initialize_transition_system_data(const Labels &labels) {
+void FTSFactory::initialize_transition_system_data(const GlobalLabels &global_labels) {
     VariablesProxy variables = task_proxy.get_variables();
-    int num_labels = task_proxy.get_operators().size();
     transition_system_data_by_var.resize(variables.size());
     for (VariableProxy var : variables) {
         TransitionSystemData &ts_data = transition_system_data_by_var[var.get_id()];
```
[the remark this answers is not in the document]
silvansievers: In contrast to your last comment, here we could change things, if you think they make sense, because all of this is internal to this factory file.

src/search/merge_and_shrink/fts_factory.cc
```diff
@@ -302,52 +307,72 @@ void FTSFactory::build_transitions_for_operator(OperatorProxy op) {
...
+        vector<int> &local_label_to_cost =
+            transition_system_data_by_var[var_id].local_label_to_cost;
+        assert(local_label_to_transitions.size() == local_to_global_labels.size());
+        assert(local_label_to_transitions.size() == local_label_to_cost.size());
         bool found_locally_equivalent_label_group = false;
-        for (size_t group_id = 0; group_id < existing_transitions_by_group_id.size(); ++group_id) {
-            const vector<Transition> &group_transitions = existing_transitions_by_group_id[group_id];
+        for (size_t local_label = 0; local_label < local_label_to_transitions.size(); ++local_label) {
+            const vector<Transition> &group_transitions = local_label_to_transitions[local_label];
             if (transitions == group_transitions) {
```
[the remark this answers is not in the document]
silvansievers: Yes. One item of the meta issue for refactoring M&S reads "consider speeding up the computation of locally equivalent labels through sorting and hashing" :-)

src/search/merge_and_shrink/fts_factory.cc
```diff
@@ -302,52 +307,72 @@ void FTSFactory::build_transitions_for_operator(OperatorProxy op) {
...
     int num_states = variable.get_domain_size();
-    int num_labels = task_proxy.get_operators().size();
 
     // Collect all irrelevant labels for this variable.
-    vector<int> irrelevant_labels;
-    for (int label_no = 0; label_no < num_labels; ++label_no) {
-        if (!is_relevant(var_no, label_no)) {
-            irrelevant_labels.push_back(label_no);
+    LabelGroup irrelevant_labels;
+    int cost = INF;
```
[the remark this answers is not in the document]
silvansievers: I'm not sure I agree. I think `numeric_limits<int>::max()` is very technical and has no semantics. I think having a defined variable for this is better.

src/search/merge_and_shrink/label_reduction.cc
```diff
@@ -60,40 +59,35 @@ void LabelReduction::initialize(const TaskProxy &task_proxy) {
...
-                // only consider non-reduced labels
-                int cost = labels.get_label_cost(label_no);
-                equivalent_label_nos[cost].push_back(label_no);
-                ++num_labels;
-            }
+    for (const equivalence_relation::Block &block : relation) {
+        unordered_map<int, vector<int>> cost_to_equivalent_labels;
+        for (int label : block) {
+            assert(label < next_new_label);
+            assert(global_labels.is_current_label(label));
```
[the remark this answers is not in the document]
silvansievers: You are right with your answer to your question. I also agree; I actually wanted to try to get rid of `is_current_label` as a public method but simply forgot to do so.

src/search/merge_and_shrink/label_reduction.cc
```diff
@@ -108,7 +102,7 @@ void LabelReduction::compute_label_mapping(
 }
 
 equivalence_relation::EquivalenceRelation
-*LabelReduction::compute_combinable_equivalence_relation(
+LabelReduction::compute_combinable_equivalence_relation(
     int ts_index,
```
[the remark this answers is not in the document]
silvansievers: General refactoring. There is no need to use raw pointers here.

src/search/merge_and_shrink/labels.h
silvansievers: See my answer at `fts_factory.cc`.

src/search/merge_and_shrink/labels.h
```diff
@@ -1,44 +1,63 @@
...
+    void reduce_labels(const std::vector<int> &old_labels);
```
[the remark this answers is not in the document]
silvansievers: Because it is **this** class which creates the new label (i.e., the int).

src/search/merge_and_shrink/merge_and_shrink_algorithm.cc
silvansievers: Yes. The whole method here is also just `dump_options`, which is done once for M&S. (It used to be there but I think was lost when refactoring in an earlier issue.)

src/search/merge_and_shrink/transition_system.cc
```diff
@@ -35,40 +32,34 @@ static void normalize_given_transitions(vector<Transition> &transitions) {
...
+    ++current_label;
 }
 
 GroupAndTransitions TSConstIterator::operator*() const {
     return GroupAndTransitions(
-        label_equivalence_relation.get_group(current_group_id),
-        transitions_by_group_id[current_group_id]);
+        local_to_global_labels[current_label],
+        local_label_to_transitions[current_label],
+        local_label_to_cost[current_label]);
```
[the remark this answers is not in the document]
silvansievers: I could add one, but then `is_current_label` couldn't become private, I assume.

src/search/algorithms/equivalence_relation.h
```diff
@@ -1,20 +1,23 @@
 #ifndef ALGORITHMS_EQUIVALENCE_RELATION_H
```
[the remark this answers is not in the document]
silvansievers: I would leave it in algorithms for the time being. It is still independent of M&S and possibly useful other cases.

src/search/merge_and_shrink/transition_system.cc
```diff
@@ -328,63 +396,67 @@ void TransitionSystem::apply_abstraction(
...
-    assert(in_sync_with_label_equivalence_relation());
+    assert(is_valid());
 }
 
 void TransitionSystem::apply_label_reduction(
     const vector<pair<int, vector<int>>> &label_mapping,
     bool only_equivalent_labels) {
-    assert(are_transitions_sorted_unique());
-    assert(in_sync_with_label_equivalence_relation());
-
```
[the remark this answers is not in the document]
silvansievers: We don't have LabelEquivalenceRelation anymore. The above `is_valid()` asserts `are_transitions_sorted_unique` and ensures that the local to global and global to local label mappings are consistent, which is what the previous `in_sync_with_label_equivalence_relation` did.

src/search/merge_and_shrink/transition_system.cc
silvansievers: I think we once decided to have every logical step in a separate loop/place. Although you are right, of course.

src/search/merge_and_shrink/transition_system.cc
silvansievers: I also don't have intuition for such questions. I added this as a note to the code but wouldn't want to try it in this issue.

FlorianPommerening: FWIW, my intuition is that if you don't have too many many duplicates, it is cheaper to add everything to a vector and then use sort/unique/erase to remove duplicates. This doesn't require the more complex data structures of a set or a hash set. But I also wouldn't change it now.

src/search/merge_and_shrink/fts_factory.cc
silvansievers: For this and other places: we consistently use `var` for `VariableProxy` and `var_id` for the int retrieved from `VariableProxy::get_id`, so I would like to stick with `var_id` here and at other analogous places.

src/search/merge_and_shrink/distances.cc
```diff
@@ -39,9 +38,8 @@ bool Distances::is_unit_cost() const {
       that the actual shortest-path algorithms (e.g.
       compute_goal_distances_general_cost) do.
     */
-    for (GroupAndTransitions gat : transition_system) {
-        const LabelGroup &label_group = gat.label_group;
-        if (label_group.get_cost() != 1)
+    for (const LocalLabelInfo &local_labe_info : transition_system) {
```
maltehelmert, line 41: Typo in variable name?

src/search/merge_and_shrink/factored_transition_system.cc
```diff
@@ -113,9 +112,10 @@ void FactoredTransitionSystem::apply_label_mapping(
     const vector<pair<int, vector<int>>> &label_mapping,
     int combinable_index) {
     assert_all_components_valid();
-    for (const auto &new_label_old_labels : label_mapping) {
-        assert(new_label_old_labels.first == labels->get_size());
-        labels->reduce_labels(new_label_old_labels.second);
+    for (const auto &entry : label_mapping) {
+        assert(entry.first == labels->get_size());
```
maltehelmert, line 116: Have you done tests in debug mode?

src/search/merge_and_shrink/label_reduction.cc
```diff
@@ -60,41 +59,34 @@ void LabelReduction::initialize(const TaskProxy &task_proxy) {
...
+            int cost = labels.get_label_cost(label);
+            cost_to_equivalent_labels[cost].push_back(label);
+            ++num_labels;
         }
-        for (auto it = equivalent_label_nos.begin();
-             it != equivalent_label_nos.end(); ++it) {
-            const vector<int> &label_nos = it->second;
-            if (label_nos.size() > 1) {
+        for (auto it = cost_to_equivalent_labels.begin();
+             it != cost_to_equivalent_labels.end(); ++it) {
```
maltehelmert, line 79: Is there a reason why this isn't a range-based loop?

src/search/merge_and_shrink/labels.cc
```diff
@@ -12,50 +12,64 @@
...
+    while (current_index < label_costs.size()
+           && label_costs[current_index] == -1) {
+        ++current_index;
     }
 }
 
-void Labels::reduce_labels(const vector<int> &old_label_nos) {
+void LabelsConstIterator::operator++() {
+    ++current_index;
+    next_valid_index();
```
maltehelmert, line 31: Change this to `advance_to_next_valid_index()` or similar? This sounds more like a method that would return the next valid index.

src/search/merge_and_shrink/labels.cc
```diff
@@ -12,50 +12,64 @@
...
       Even though we currently only support exact label reductions where
       reduced labels are of equal cost, to support non-exact label reductions,
       we compute the cost of the new label as the minimum cost of all old
       labels reduced to it to satisfy admissibility.
     */
     int new_label_cost = INF;
-    for (size_t i = 0; i < old_label_nos.size(); ++i) {
-        int old_label_no = old_label_nos[i];
-        int cost = get_label_cost(old_label_no);
+    for (size_t i = 0; i < old_labels.size(); ++i) {
```
maltehelmert, line 48: Is there a reason why this is not a range-based loop?

src/search/merge_and_shrink/labels.h
```diff
@@ -4,45 +4,65 @@
...
+class LabelsConstIterator {
+    const std::vector<int> &label_costs;
+    std::size_t current_index;
+
+    void next_valid_index();
 public:
-    explicit Label(int cost_)
-        : cost(cost_) {
+    LabelsConstIterator(const std::vector<int> &label_costs, bool end);
+    void operator++();
```
maltehelmert, line 15: I think `operator++` normally shouldn't return `void`.

src/search/merge_and_shrink/labels.h
```diff
@@ -4,45 +4,65 @@
...
+
+    int operator*() const {
+        return int(current_index);
```
maltehelmert, line 18: I think C-style cases are discouraged, and we should use something like `static_cast`?

src/search/merge_and_shrink/labels.h
```diff
@@ -4,45 +4,65 @@
...
 
 /*
-  This class serves both as a container class to handle the set of all labels
-  and to perform label reduction on this set.
+  This class serves both as a container class to handle the set of all
+  labels and to perform label reduction on this set.
+
+  Labels are identified via integers indexing label_costs, which store their
+  costs. When using label reductions, reduced labels are set to -1 in label
+  costs.
```
maltehelmert, line 36: "which store" => "which stores"? I wouldn't use plural to refer to a variable, even if that variable is a collection.  I wouldn't say "reduced labels" to denote label IDs that become unused. In my view, label reduction is like abstraction; if I map labels A and B to the new label C (or to the label A), then in all cases "A" and "B" are reduced. But I think you mean here label IDs that have become obsolete. I would use a word like "labels that become unused".

I would not write "in label costs" to refer to the variable "label_costs". Better to use "label_costs".

src/search/merge_and_shrink/labels.h
```diff
@@ -4,45 +4,65 @@
...
 */
 class Labels {
-    std::vector<std::unique_ptr<Label>> labels;
-    int max_size; // the maximum number of labels that can be created
+    std::vector<int> label_costs;
+    int max_num_labels; // The maximum number of labels that can be created.
+    int num_active_labels; // The current number of active (non-reduced) labels.
```
maltehelmert, line 41: OK, if you use "active", I suggest "inactive" above.

src/search/merge_and_shrink/labels.h
```diff
@@ -4,45 +4,65 @@
...
-    int get_label_cost(int label_no) const;
-    void dump_labels(utils::LogProxy &log) const;
+    Labels(std::vector<int> &&label_costs, int max_num_labels);
+    void reduce_labels(const std::vector<int> &old_labels);
+    int get_label_cost(int label) const;
+    void dump_labels() const;
+
     int get_size() const {
-        return labels.size();
+        return label_costs.size();
```
maltehelmert, line 49: It's maybe not obvious what `get_size()` is there for/what the return value means; does it need an explanation? (Intuitively, it could mean either what it means here or the number of active labels.)

maltehelmert: Perhaps avoid the method name altogether and use one that resonates with `get_num_active_labels`, e.g. using `total` or whatever. But perhaps it doesn't need changing.

src/search/merge_and_shrink/labels.h
```diff
@@ -4,45 +4,65 @@
...
+    }
+
+    LabelsConstIterator begin() const {
+        return LabelsConstIterator(label_costs, false);
     }
-    int get_max_size() const {
-        return max_size;
+
+    LabelsConstIterator end() const {
+        return LabelsConstIterator(label_costs, true);
```
maltehelmert, line 65: I think a more "normal" implementation/interface for the iterators would pass 0 and the end position to LabelsConstIterator to the iterator instead of false or true. And I think an even more idiomatic (and efficient) implementation would avoid having the iterators hang on to label_costs and only pass in an iterator to label_costs instead. But again, perhaps this kind of polishing is not necessary.

src/search/merge_and_shrink/merge_and_shrink_algorithm.cc
```diff
@@ -83,6 +83,9 @@ void MergeAndShrinkAlgorithm::dump_options() const {
             << shrink_threshold_before_merge << endl;
         log << endl;
 
+        shrink_strategy->dump_options(log);
+        log << endl;
```
maltehelmert, line 87: Is this here intentionally? Is it related to the issue? I don't mind adding it if you want to add it. Should it be mentioned in the change log then? More generally, I think you don't have a change log entry yet?

maltehelmert: (BTW, I'm not convinced that adding a blank line is great. With structured logging formats this looks weird.)

src/search/merge_and_shrink/transition_system.h
```diff
@@ -41,42 +39,43 @@ struct Transition {
...
-    GroupAndTransitions(const LabelGroup &label_group,
-                        const std::vector<Transition> &transitions)
+using LabelGroup = std::vector<int>;
+
+class LocalLabelInfo {
+    LabelGroup label_group;
+    std::vector<Transition> transitions;
+    int cost;
+
+    friend class TransitionSystem;
```
maltehelmert, line 49: `friend` is generally a pretty bad code smell. Is it necessary?

src/search/merge_and_shrink/transition_system.h
```diff
@@ -94,51 +93,50 @@ class TransitionSystem {
...
 
     int num_states;
     std::vector<bool> goal_states;
     int init_state;
 
     /*
-      Check if two or more labels are locally equivalent to each other, and
-      if so, update the label equivalence relation.
+      Check if two or more local labels are locally equivalent to each other,
+      and if so, merge them and store their transitions only once.
```
maltehelmert, line 113: Is "locally equivalent" necessary, or would "equivalent" fit better? Is there a meaningful other (non-local) way in which "local labels" can be equivalent?

maltehelmert: (If not, perhaps the method name might then also change to `compute_equivalent_local_labels` or some such.)

If you prefer the comment in the current form, we can leave it as is.

src/search/merge_and_shrink/transition_system.h
```diff
@@ -94,51 +93,50 @@ class TransitionSystem {
...
 
-    const std::vector<Transition> &get_transitions_for_group_id(int group_id) const {
-        return transitions_by_group_id[group_id];
-    }
-
     // Statistics and output
     int compute_total_transitions() const;
     std::string get_description() const;
+
+    // Consistency
```
maltehelmert, line 121: I'm not sure I like this line as a comment, especially since "consistency" is also a heuristic property, and also `dump_label_mapping` doesn't sound like something I'd search for under "Consistency".

src/search/merge_and_shrink/transition_system.cc
```diff
@@ -245,32 +234,41 @@ void TransitionSystem::compute_locally_equivalent_labels() {
     /*
       Compare every group of labels and their transitions to all others and
       merge two groups whenever the transitions are the same.
+
+      Note that there can be empty local label groups after applying label
+      reduction when combining labels which are combinable for this transition
+      system.
```
maltehelmert, line 240: This looks like an inefficient implementation, but so does the old one. (More of a comment for the future.) Is this sometimes a bottleneck?

src/search/merge_and_shrink/transition_system.cc
```diff
@@ -328,108 +327,114 @@ void TransitionSystem::apply_abstraction(
...
+                // we would need to recompute the cost of the local label here.
+            }
+        }
     } else {
         /*
-          Go over all mappings, collect transitions of old groups and
-          remember all affected group IDs. This needs to happen *before*
-          updating label_equivalence_relation, because after updating it,
-          we cannot find out the group ID of reduced labels anymore.
+          Go over all label reductions (given by label_mapping). For each
```
maltehelmert, line 379: I'm not a great fan of the terminology here ("all label reductions"). In the terminology of our papers, the whole label mapping defines *one* label reduction transformation. But if you prefer it this way, no objections.

src/search/merge_and_shrink/factored_transition_system.cc
silvansievers: I will do.

src/search/merge_and_shrink/merge_and_shrink_algorithm.cc
silvansievers: This was removed accidentally in another issue. I don't think that there should be a change log entry. The empty lines are an attempt to structuring, but yes, with the prefix printed, it is probably nicer (and a lot easier) to not use any empty lines for structuring.

silvansievers: But I wouldn't touch this now -- this is done at several places, and I don't dislike it everywhere . For example, having an empty line between the output of each iteration of M&S doesn't seem that bad to me, but we can certainly discuss this.

src/search/merge_and_shrink/transition_system.h
silvansievers: It allows `TransitionSystem` to modify the private members `label_group` , `transitions`, and `cost`. The alternative would be to add three non-const getters in addition to the existing const getters. Do you prefer that? Mir ist's egal.

src/search/merge_and_shrink/transition_system.cc
silvansievers: I might be and the meta issue (https://issues.fast-downward.org/issue567) for refactoring M&S is aware of this. Item C reads:
C. TransitionSystem::compute_locally_equivalent_labels(): make computation
faster through sorting transitions (O(n log n)) or hashing (O(n)). Sorting is
probably better, because we do not need to consider all transitions in the best
case. [This doesn't seem to be a huuuuuge bottleneck, though. According to
experiments (not on the tracker), it can be up to 10% of merge-and-shrink
construction time in some configs.]

src/search/merge_and_shrink/merge_and_shrink_algorithm.cc
maltehelmert: I think with this kind of logging format, it's weird with or without the prefix line. But feel free to leave this as is if it's not new to this issue.

src/search/merge_and_shrink/transition_system.cc
maltehelmert: OK, up to 10% isn't necessarily a deal (although also I wouldn't necessarily trust old numbers like these). In any case, doesn't appear relevant now.

src/search/merge_and_shrink/transition_system.h
maltehelmert: OK, I have to check how the class is used in general to understand better what the benefits/drawbacks of this would be. Give some time to clone the code and look through this.

maltehelmert: It looks like the only non-const non-friend use of `LocalLabelInfo` is in `fts_factory.cc`. In `TransitionSystem` I also only found one non-const use, which changes three things at the same time. Would it make sense to update this change in a (non-const) method of LocalLabelInfo? Then we wouldn't need friend any more. With friend, we cannot be sure of the class invariants of `LocalLabelInfo` without checking all of `TransitionSystem`, which is a very large class and only messes with the internal details of `LocalLabelInfo` in a single place.

silvansievers: I pushed a commit to remove the friend declaration. It adds several methods to LocalLabelInfo. Do you prefer this version over the alternative of turning LocalLabelInfo into a struct? For me, either way is fine.

src/search/merge_and_shrink/transition_system.cc
```diff
@@ -28,47 +25,95 @@ ostream &operator<<(ostream &os, const Transition &trans) {
...
-        ++current_group_id;
+void LocalLabelInfo::apply_same_cost_label_mapping(
+    int new_label, const vector<int> &old_labels) {
+    assert(is_consistent());
+    remove_labels(old_labels);
+    label_group.push_back(new_label);
+    assert(is_consistent());
+}
+
+void LocalLabelInfo::remove_labels(const vector<int> &old_labels) {
```
silvansievers, line 62: One could consider sorting the given old_labels in the method instead of outside.

src/search/merge_and_shrink/transition_system.cc
```diff
@@ -28,47 +25,95 @@ ostream &operator<<(ostream &os, const Transition &trans) {
...
+  Class for representing groups of labels with equivalent transitions in a
+  transition system. See also documentation for TransitionSystem.
+
+  The local label is in a consistent state if label_group and transitions
+  are sorted and unique.
+*/
+LocalLabelInfo::LocalLabelInfo(
+    LabelGroup &&label_group,
+    std::vector<Transition> &&transitions,
+    int cost)
```
silvansievers, line 38: Instead of computing cost outside, we could pass the Labels instance and compute cost in the constructor (by calling recompute_cost). This would mean, for one case in FTSFactory, to iterate twice over the label group: once here to compute the cost, and once outside to compute the mapping label_to_local_label.

src/search/merge_and_shrink/transition_system.cc
```diff
@@ -28,47 +25,95 @@ ostream &operator<<(ostream &os, const Transition &trans) {
...
+        cost = min(cost, label_cost);
+    }
+    assert(is_consistent());
 }
 
-void TSConstIterator::next_valid_index() {
-    while (current_group_id < label_equivalence_relation.get_size()
-           && label_equivalence_relation.is_empty_group(current_group_id)) {
-        ++current_group_id;
+void LocalLabelInfo::apply_same_cost_label_mapping(
```
silvansievers, line 54: This does not more than calling recompute_cost and adding a label; so maybe the name isn't ideal?

src/search/merge_and_shrink/transition_system.cc
```diff
@@ -196,18 +240,24 @@ unique_ptr<TransitionSystem> TransitionSystem::merge(
...
-            vector<int> &new_labels = bucket.second;
+            LabelGroup &new_labels = bucket.second;
             if (new_transitions.empty()) {
                 dead_labels.insert(dead_labels.end(), new_labels.begin(), new_labels.end());
             } else {
                 sort(new_transitions.begin(), new_transitions.end());
-                label_groups.push_back(move(new_labels));
-                transitions_by_group_id.push_back(move(new_transitions));
+                sort(new_labels.begin(), new_labels.end());
+                int new_local_label = local_label_infos.size();
```
silvansievers, line 254: This is one of the places where we would no longer need to sort outside if the constructor did it.

src/search/merge_and_shrink/transition_system.cc
```diff
@@ -196,18 +240,24 @@ unique_ptr<TransitionSystem> TransitionSystem::merge(
...
+                int cost = INF;
+                for (int label : new_labels) {
+                    cost = min(ts1.labels.get_label_cost(label), cost);
+                    label_to_local_label[label] = new_local_label;
+                }
```
silvansievers, line 259: This is one of the places where we wouldn't need to compute the cost, but only update the mapping, if the constructor computed cost.

silvansievers (commented):
I wasn't quite sure if it is cleaner to expect that label groups and transitions are sorted outside of LocalLabelInfo so that that one can expect consistent data, or if LocalLabelInfo itself should take care of this. Similar for computing costs.

Speaking of costs: the consistency test of LocalLabelInfo could also ensure that cost equals the minimum cost of all labels in label_group; however, that would require to pass Labels to the consistency method, which might be not so nice.

maltehelmert:
Shall we talk quickly over Zoom?

maltehelmert:
Shall we talk quickly over Zoom?

src/search/merge_and_shrink/transition_system.h
```diff
@@ -53,27 +64,38 @@ class LocalLabelInfo {
         : label_group(move(label_group)),
           transitions(move(transitions)),
           cost(cost) {
+        assert(utils::is_sorted_unique(this->label_group) &&
+               utils::is_sorted_unique(this->transitions));
```
maltehelmert, line 68: Is this `assert(is_consistent())`? Then I'd use this.

PR #12, opened 2020-08-06
Issue960

overactor:


jendrikseipp:
Hey Rik, just so you don't stumble over this later: the two latest commits don't have the [issue960] prefix.

src/search/lp/lp_solver.cc
```diff
@@ -304,6 +378,23 @@ void LPSolver::solve() {
...
+void LPSolver::write_lp(string filename) const {
+    try {
+        lp_solver->writeLp(filename.c_str());
+    } catch (CoinError &error) {
+        handle_coin_error(error);
+    }
+}
+
+void LPSolver::print_failure_analysis() const {
+    cout << "isAbandoned: " << lp_solver->isAbandoned() << endl;
```
FlorianPommerening, line 390: I'd change the strings from `isAbandoned` to `abandoned`, `isProvenOptimal` to `proven optimal` and so on.

src/search/lp/lp_solver.h
```diff
@@ -73,6 +76,28 @@ struct LPVariable {
...
+
+    utils::NamedVector<LPVariable> variables;
+    utils::NamedVector<LPConstraint> constraints;
+
+public:
+    explicit LinearProgram(LPObjectiveSense sense, utils::NamedVector<LPVariable> &&variables, utils::NamedVector<LPConstraint> &&constraints, std::string objective_name = "")
+        : sense(sense), variables(std::move(variables)), constraints(std::move(constraints)), objective_name(objective_name) {
+    }
+
+    utils::NamedVector<LPVariable> &get_variables();
```
FlorianPommerening, line 91: Please add some documentation explaining that variables and constraints can be named but doing so will have an impact on performance, so this should only be done for debugging purposes.

src/search/utils/collections.h
```diff
@@ -151,6 +152,89 @@ int estimate_unordered_map_bytes(int num_entries) {
     // See comments for _estimate_hash_table_bytes.
     return _estimate_hash_table_bytes<std::unordered_map<Key, Value>>(num_entries);
 }
+
+template<typename T>
```
FlorianPommerening, line 156: Please add some documentation explaining the purpose of this class.

src/search/utils/collections.h
```diff
@@ -151,6 +152,89 @@ int estimate_unordered_map_bytes(int num_entries) {
...
+        return elements.end();
+    }
+
+    void clear() {
+        elements.clear();
+        names.clear();
+    }
+
+    void resize(int count, T value) {
+        elements.resize(count, value);
```
FlorianPommerening, line 235: Maybe add a comment here to explain why `names` is not resized. Also, I'd re-order the methods, so that `reserve`, `resize`, `size`, and `clear` are together and `has_names` is together with the other name-related methods.

FlorianPommerening (changes requested):

experiments/issue960/v1.py
```diff
@@ -15,9 +15,9 @@
 BENCHMARKS_DIR = os.environ["DOWNWARD_BENCHMARKS"]
 REVISIONS = ["issue960-base", "issue960-v1"]
 CONFIGS = [
-    IssueConfig("opcount-seq", ["--search", "astar(operatorcounting([state_equation_constraints()]))"]),
-    IssueConfig("diverse-pot", ["--search", "astar(diverse_potentials(random_seed=1729)"]),
-    IssueConfig("lm_count", ["--search", "astar(lmcount(lm_rhw(), admissible=true, optimal=true)"]),
+    IssueConfig("opcount-seq-lmcut-soplex", ["--search", "astar(operatorcounting([state_equation_constraints(), lmcut_constraints()], lpsolver=soplex))"]),
```
FlorianPommerening, line 18: If you re-do the experiments, I'd run them with both CPLEX and SoPlex. For the operator-counting heuristic, the config including `lmcut_constraints` is better and for the landmark heuristic, I'd also use the one with h^m landmarks (the one with `lm_merged(..., lm_hm(m=1))`).

src/search/utils/collections.h
```diff
@@ -151,6 +152,98 @@ int estimate_unordered_map_bytes(int num_entries) {
     // See comments for _estimate_hash_table_bytes.
     return _estimate_hash_table_bytes<std::unordered_map<Key, Value>>(num_entries);
 }
+
+    /**
+     * NamedVector is a vector-like collection with optional names to be associated with each element.
+     * It is optimized to have minimal overhead when there are no names.
+     * Any name which is not specified is assumed to be an empty string.
```
jendrikseipp, line 159: This comment doesn't follow our convention. See "Comments" under http://www.fast-downward.org/ForDevelopers/CodingConventions

src/search/utils/collections.h
```diff
@@ -151,6 +152,98 @@ int estimate_unordered_map_bytes(int num_entries) {
...
+                // All unspecified names are empty by default.
+                return;
+            }
+            names.resize(index + 1, "");
+        }
+        names[index] = name;
+    }
+
+    std::string get_name(int index) const {
+        int names_num = names.size();
```
jendrikseipp, line 199: I'd use num_names instead of names_num.

src/search/utils/collections.h
```diff
@@ -151,6 +152,98 @@ int estimate_unordered_map_bytes(int num_entries) {
...
+        return elements.end();
+    }
+
+    void clear() {
+        elements.clear();
+        names.clear();
+    }
+
+    void reserve(int capacity) {
+        // No space is reserved in the names vector because it is kept at minimal length and space is only used when necessary.
```
jendrikseipp, line 238: Break comment line, please (here and below).

src/search/utils/collections.h
```diff
@@ -9,6 +9,7 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
+#include <iostream>
```
jendrikseipp, line 12: Order includes alphabetically, please.

src/search/utils/collections.h
```diff
@@ -151,6 +152,98 @@ int estimate_unordered_map_bytes(int num_entries) {
...
     return _estimate_hash_table_bytes<std::unordered_map<Key, Value>>(num_entries);
 }
+
+    /**
+     * NamedVector is a vector-like collection with optional names to be associated with each element.
+     * It is optimized to have minimal overhead when there are no names.
+     * Any name which is not specified is assumed to be an empty string.
+     */
+template<typename T>
+class NamedVector {
```
jendrikseipp, line 162: I think this new class should be moved to its own file. I think it fits best under algorithms/named_vector.h, given the other files in that directory, even though this is more a data structure than an algorithm.

jendrikseipp (changes requested):
I only looked at NamedVector so far. I'll have a look at the other files later. The code looks good, as usual I have some nitpicks :-)

jendrikseipp (changes requested):
Some more comments.

src/search/lp/lp_solver.h
```diff
@@ -73,6 +78,31 @@ struct LPVariable {
...
+    LPObjectiveSense sense;
+    std::string objective_name;
+
+    utils::NamedVector<LPVariable> variables;
+    utils::NamedVector<LPConstraint> constraints;
+    double infinity;
+
+public:
+    explicit LinearProgram(LPObjectiveSense sense, utils::NamedVector<LPVariable> &&variables, utils::NamedVector<LPConstraint> &&constraints, std::string objective_name = "")
+        : sense(sense), objective_name(objective_name), variables(std::move(variables)), constraints(std::move(constraints)) {
```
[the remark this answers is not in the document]
FlorianPommerening: No, the objective name is a name you can pick for the linear expression in the objective function. For example "total_cost" in operator-counting LPs or "initial_h_value"/"average_h_value" for a potential heuristic LP. Although, now that I see this line again, maybe it would be more consistent with the names of variables and constraints if we use `set_objective_name()` rather than passing the value in the constructor.

src/search/lp/lp_solver.cc
```diff
@@ -69,13 +70,61 @@ void LPConstraint::insert(int index, double coefficient) {
...
       objective_coefficient(objective_coefficient) {
 }
 
+utils::NamedVector<LPVariable> &LinearProgram::get_variables() {
+    return variables;
+}
+
+utils::NamedVector<LPConstraint> &LinearProgram::get_constraints() {
+    return constraints;
+}
```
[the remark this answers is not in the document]
FlorianPommerening: As far as I know, we use them in cases where the LP is created once and then modified (in the LM-count heuristic for example).

src/search/utils/collections.h
overactor: I think it is a holdover from debugging code, I will remove it.

src/search/lp/lp_solver.h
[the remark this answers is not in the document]
overactor: We set the names of variables and constraints upon initializing the lp_solver as well by passing in the NamedVectors at that point. I do suppose the variable and constraint names can be changed after initialization and the objective name currently cannot.

src/search/utils/collections.h
overactor: I think I'll wait for a second opinion on this before moving this class. @FlorianPommerening what do you think?

src/search/lp/lp_solver.cc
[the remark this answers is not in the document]
overactor: Yes, both are currently used.

src/search/lp/lp_solver.h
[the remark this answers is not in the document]
jendrikseipp: In any case, I think the `objective_name` warrants some documentation along the lines of what Florian wrote.

src/search/lp/lp_solver.cc
```diff
@@ -304,6 +381,23 @@ void LPSolver::solve() {
...
+    try {
+        lp_solver->writeLp(filename.c_str());
+    } catch (CoinError &error) {
+        handle_coin_error(error);
+    }
+}
+
+void LPSolver::print_failure_analysis() const {
+    cout << "abandoned: " << lp_solver->isAbandoned() << endl;
+    cout << "proven optimal:  " << lp_solver->isProvenOptimal() << endl;
```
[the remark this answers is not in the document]
overactor: It took me a while to see what you meant by this.

src/search/landmarks/landmark_cost_assignment.h
```diff
@@ -46,14 +46,16 @@ class LandmarkEfficientOptimalSharedCostAssignment : public LandmarkCostAssignme
...
+    lp::LinearProgram lp;
+    // We keep an additional copy of the constraints around to avoid some effort with recreating the vector (see issue443).
     std::vector<lp::LPConstraint> lp_constraints;
-    std::vector<lp::LPConstraint> non_empty_lp_constraints;
 public:
     LandmarkEfficientOptimalSharedCostAssignment(const std::vector<int> &operator_costs,
                                                  const LandmarkGraph &graph,
                                                  lp::LPSolverType solver_type);
 
+    lp::LinearProgram initial_program();
```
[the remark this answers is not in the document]
overactor: It can be private, but not const or static.

src/search/lp/lp_solver.h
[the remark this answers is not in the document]
FlorianPommerening: I meant inside of `NamedVector` we do not pass the name as an optional parameter to `push_back` anymore but switched to a method where we use `set_name` after we add the actual content. I think it would make sense to do the same here. It also makes it easier to implement things in a way that names are only used in some cases.

src/search/utils/collections.h
FlorianPommerening: I'm fine with that. As Jendrik said, the name "algorithms" is not that good of a fit, but this is already the case for the things that are there. (It should probably be "algorithms and data structures".)

src/search/DownwardFiles.cmake
```diff
@@ -440,7 +448,6 @@ fast_downward_plugin(
     SOURCES
         search_engines/lazy_search
     DEPENDS ORDERED_SET SUCCESSOR_GENERATOR
-    DEPENDENCY_ONLY
```
FlorianPommerening, line 443: Did you remove this accidentally? Looks unrelated to the patch.

src/search/landmarks/landmark_cost_assignment.cc
```diff
@@ -126,12 +126,17 @@ LandmarkEfficientOptimalSharedCostAssignment::LandmarkEfficientOptimalSharedCost
     const LandmarkGraph &graph,
     lp::LPSolverType solver_type)
     : LandmarkCostAssignment(operator_costs, graph),
-      lp_solver(solver_type) {
+      lp_solver(solver_type),
+      lp(initial_lp()) {}
```
FlorianPommerening, line 130: I think we usually add a line break after `{` in situations like this. Also, names of functions should be verbs, so I'd change this to
```
      lp(build_initial_lp()) {
}
```

src/search/lp/lp_solver.cc
```diff
@@ -158,11 +214,36 @@ void LPSolver::load_problem(LPObjectiveSense sense,
...
         */
-        if (sense == LPObjectiveSense::MINIMIZE) {
+        if (lp.get_sense() == LPObjectiveSense::MINIMIZE) {
             lp_solver->setObjSense(1);
         } else {
             lp_solver->setObjSense(-1);
         }
+
+        if (!lp.get_objective_name().empty()) {
+            lp_solver->setObjName(lp.get_objective_name());
```
FlorianPommerening, line 224: Is it possible to set a name for an objective and not for variables and constraints? In this case `OsiNameDiscipline` will not be set below.

src/search/lp/lp_solver.h
```diff
@@ -1,6 +1,8 @@
 #ifndef LP_LP_SOLVER_H
 #define LP_LP_SOLVER_H
 
+#include "../algorithms/named_vector.h"
+#include "../utils/collections.h"
```
FlorianPommerening, line 5: Do we still need the second include?

FlorianPommerening (changes requested):
I added a few more comments.

src/search/lp/lp_solver.h
```diff
@@ -73,6 +78,33 @@ struct LPVariable {
...
+    LPObjectiveSense sense;
+    std::string objective_name;
+
+    named_vector::NamedVector<LPVariable> variables;
+    named_vector::NamedVector<LPConstraint> constraints;
+    double infinity;
+
+public:
+    // objective_name is the name of the objective function used when writing the lp to a file.
+    explicit LinearProgram(LPObjectiveSense sense, named_vector::NamedVector<LPVariable> &&variables, named_vector::NamedVector<LPConstraint> &&constraints)
```
jendrikseipp, line 91: Doesn't need to be explicit. Also, please break the line.

src/search/lp/lp_solver.h
```diff
@@ -73,6 +78,33 @@ struct LPVariable {
...
+        : sense(sense), objective_name(""), variables(std::move(variables)), constraints(std::move(constraints)) {
```
jendrikseipp, line 92: No need to invoke the default constructor for objective_name.

jendrikseipp (changes requested):
Two small nitpicks, otherwise looks good to me!

src/search/algorithms/named_vector.h
```diff
@@ -0,0 +1,103 @@
...
+#include <string>
+#include <vector>
+
+namespace named_vector {
+/*
+  NamedVector is a vector-like collection with optional names to be associated with each element.
+  It is optimized to have minimal overhead when there are no names.
+  Any name which is not specified is assumed to be an empty string.
+ */
+template<typename T>
```
FlorianPommerening, line 13: Please document that the intended use of `NamedVector` is for debugging, i.e., where we wanted names for debugging that can be set only in debugging mode without causing an overhead when they are not set. (This could expand or replace the second sentence in documentation).

Also, please document that accessing the name of an invalid index is an error (see below).

src/search/algorithms/named_vector.h
```diff
@@ -0,0 +1,103 @@
...
+
+    T &operator[](int index) {
+        return elements[index];
+    }
+
+    bool has_names() const {
+        return !names.empty();
+    }
+
+    void set_name(int index, const std::string &name) {
```
FlorianPommerening, line 39: Getting and setting a name for an invalid index should be an error. You can use `utils::in_bounds` to check the index and `utils::ExitCodes::SearchCriticalError` (I think) for exiting. Look for other places where exit codes from `utils::ExitCodes` are used to see the usage.

src/search/algorithms/named_vector.h
```diff
@@ -0,0 +1,103 @@
...
+  NamedVector is a vector-like collection with optional names to be associated with each element.
+  It is optimized to have minimal overhead when there are no names.
+  Any name which is not specified is assumed to be an empty string.
+ */
+template<typename T>
+class NamedVector {
+    std::vector<T> elements;
+    std::vector<std::string> names;
+public:
+    NamedVector() = default;
```
FlorianPommerening, line 18: Since we do not change any of the constructors, we should leave out all of them.

This is called the "rule of zero" and is a bit of a controversial topic apparently (https://www.fluentcpp.com/2019/04/23/the-rule-of-zero-zero-constructor-zero-calorie/).

src/search/algorithms/named_vector.h
```diff
@@ -0,0 +1,103 @@
...
+    template<typename ... _Args>
+    void emplace_back(_Args && ... __args) {
+        elements.emplace_back(std::forward<_Args>(__args) ...);
+    }
+
+    void push_back(const T &element) {
+        elements.push_back(element);
+    }
+
+    T &operator[](int index) {
```
FlorianPommerening, line 31: Please add a `const` version of this because otherwise a `const NamedVector` cannot be used for much.

src/search/algorithms/named_vector.h
```diff
@@ -0,0 +1,103 @@
...
+        names.clear();
+    }
+
+    void reserve(int capacity) {
+        /* No space is reserved in the names vector because it is kept at minimal length
+           and space is only used when necessary. */
+        elements.reserve(capacity);
+    }
+
+    void resize(int count, T value) {
```
FlorianPommerening, line 95: Please add a default value to `value` so it behaves in the same way as `vector::resize`.

FlorianPommerening (changes requested):
I talked to Malte about this and he had some additional requests for changes.

src/search/algorithms/named_vector.h
maltehelmert: I would assert rather than test and exit. This is more consistent with other parts of the code and avoids an unnecessary dependency of the algorithms (which are not meant to be specific to Fast Downward) on Fast Downward's exit code.

src/search/algorithms/named_vector.h
maltehelmert: I don't think the rule of zero is that controversial, or else it wouldn't be so prominent in the Core Guidelines. Of course there are different opinions by someone on *anything*, but I find the discussed Scott Meyer example unconvincing, at least from the second-hand discussion. Firstly, it shows that there is an issue if you *violate* the rule of zero by introducing an explicit destructor, but of course the point of the rule is to follow it, not violate it. And these issues are caught by compiler warnings.

But this is beside the main point. The recommendations are either the rule of zero or the rule of five. The code we currently have is exactly the kind of code Scott Meyer warns against when misusing the rule of 0 because (if I see it correctly) it implicitly defines move assignment as copy assignment. If I understand Meyer's argument correctly, he thinks the rule of 5 should be followed rather than the rule of 0 because if you start with the rule of 0 and then (by not paying attention) add *some* of the 5 methods but not all, you add in a dangerous spot. This is exactly where the current version of the code is.

Following either the rule of 0 or the rule of 5 would fix that. I don't feel very strongly about which of the two we follow, but the rule of 0 certainly seems to be what the C++ community is moving towards and what we have started using more frequently in the code.

FlorianPommerening: I think "controversial" may have come across in the wrong way: I just meant the different recommendations of Meyers and the core guidelines. I didn't want to imply that we are not agreeing on what rule to follow.

maltehelmert: PS: Also the page that Florian linked ended by saying that this is not controversial:

> To conclude I’ll leave the closing word to Arne Mertz, that summarized the debate with a rule that everyone agrees on, the “Rule of All or Nothing”:
>    As long as you can, stick to the Rule of Zero, but if you have to write at least one of the Big Five, default the rest.

...and this is kind-of what the Core Guidelines say as well (C.20 + C.21), although C.21 doesn't explicitly mention "defaulting" but rather "defining". But defaulting is of course the simplest way to define something where it has the right semantics.

PR #15, opened 2020-11-18
restore individual action costs in plan output file

emilkeyder-invitae:
.

### DIFF
--- a/src/search/plan_manager.cc
+++ b/src/search/plan_manager.cc
@@ -58,7 +58,8 @@ void PlanManager::save_plan(
     OperatorsProxy operators = task_proxy.get_operators();
     for (OperatorID op_id : plan) {
         cout << operators[op_id].get_name() << " (" << operators[op_id].get_cost() << ")" << endl;
-        outfile << "(" << operators[op_id].get_name() << ")" << endl;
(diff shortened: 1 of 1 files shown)


src/search/plan_manager.cc
```diff
@@ -58,7 +58,8 @@ void PlanManager::save_plan(
     OperatorsProxy operators = task_proxy.get_operators();
     for (OperatorID op_id : plan) {
         cout << operators[op_id].get_name() << " (" << operators[op_id].get_cost() << ")" << endl;
-        outfile << "(" << operators[op_id].get_name() << ")" << endl;
+        outfile << "(" << operators[op_id].get_name() << ") ; " << "("
+		<< operators[op_id].get_cost() << ")" << endl;
```
maltehelmert, line 62: Hi Emil, this is mostly based on oral tradition, but I *think* when the IPC plan file format was introduced, it did not allow lines that are a mixture of actions and comments; comments must be their own lines. So I would be somewhat reluctant to merge this. But perhaps we could control the output format with an option or something like this.

emilkeyder-invitae: Hi Malte, I was trying to push this to our internal fork and pushed it to your repo accidentally, please ignore. Sorry for the trouble.

PR #18, opened 2021-01-22
Issue988

ClemensBuechner:
This issue wants to make the `LandmarkGraph` class constant after creation. The logic should not change, only the responsibilities. The following things changed so far:
- moved status of landmark to the `LandmarkStatusManager`
- compute landmark status on demand, only store which landmarks are reached in a state
- moved computation of LM-count value to the heuristic itself
- give `LandmarkGraph` access to the `TaskProxy` for which it was created
- changed `LandmarkGraph::dump()` output to DOT format to make visualizing the graph easy


src/search/landmarks/landmark_cost_assignment.h
```diff
@@ -11,31 +13,35 @@ class OperatorsProxy;
...
     virtual ~LandmarkCostAssignment() = default;
 
-    virtual double cost_sharing_h_value() = 0;
+    virtual double cost_sharing_h_value(const GlobalState &state) = 0;
 };
 
 class LandmarkUniformSharedCostAssignment : public LandmarkCostAssignment {
     bool use_action_landmarks;
 public:
     LandmarkUniformSharedCostAssignment(const std::vector<int> &operator_costs,
```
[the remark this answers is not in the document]
ClemensBuechner: What's the problem here?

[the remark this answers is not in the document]
silvansievers: If we need to break lines here, I think we prefer to make line break after the opening parenthesis so that the next line starts with `const std::vector<int>` on a one-level indentation compared to the constructor. Then all further parameters are not as extremely indented as they are now.

src/search/landmarks/landmark_graph.h
```diff
@@ -48,19 +45,17 @@ class LandmarkNode {
...
 
-    utils::HashSet<FactPair> forward_orders;
     std::set<int> first_achievers;
     std::set<int> possible_achievers;
 
     int get_id() const {
         return id;
     }
 
+    // Possibly should not be changable
```
silvansievers, line 58: I would prefer marking such suggestions/ideas with a TODO.

src/search/landmarks/landmark_status_manager.h
```diff
@@ -7,24 +7,43 @@ namespace landmarks {
...
-    bool update_lm_status(const GlobalState &global_state);
+    void update_lm_status(const GlobalState &global_state);
+    bool dead_end_exists();
 
     void set_landmarks_for_initial_state(const GlobalState &initial_state);
     bool update_reached_lms(const GlobalState &parent_global_state,
                             OperatorID op_id,
                             const GlobalState &global_state);
+
+    /*
```
silvansievers, line 33: I think we should mark this with a TODO.

src/search/landmarks/landmark_status_manager.cc
```diff
@@ -106,69 +117,80 @@ bool LandmarkStatusManager::update_reached_lms(const GlobalState &parent_global_
...
-            return true;
+    LandmarkNode *node = lm_graph.get_lm_for_index(id);
+    if (node->is_true_in_state(state)) {
+        return false;
+    } else if (node->is_goal()) {
+        return true;
+    } else {
+        /* For all A ->_gn B, if B is not reached and A currently not
+           true, since A is a necessary precondition for actions
+           achieving B for the first time, it must become true again. */
```
silvansievers, line 180: Is this new text or did it exist before? Just asking out of curiosity.

Note that the style for long comments is
/*
  text
  text
*/

i.e., slightly differently than it is here.

ClemensBuechner: The style guide sais it's fine if "short comments (2-3 lines)" are commented in this fashion, but I don't have a strong opinion on this, so I will change it.

silvansievers: Ok, me neither, I just thought the limit was 2 lines. Never mind then.

ClemensBuechner: This text is new.

maltehelmert: I think these days, I always use the style that Silvan cites also if it's only two lines, and I think the code is moving towards this convention in general. I think the exception in the style guide is there primarily for people that felt uncomfortable about the extra lines back when we first discussed style, but over time most people have become comfortable with this way of commenting. This is a common thing. When we first debate such things, opinions are often very strong and we debate heavily over what should or shouldn't be allowed, but over time we often fall into patterns that make some of the old discussions redundand. So I have a preference for the version with start/end on separate lines, but I don't feel strongly about it.

PR #23, opened 2021-02-09
Issue383

ClemensBuechner:
Issue383 denotes a bug where the `LandmarkFactory::approximate_reasonable_orderings()` function is returned arbitrarily when a landmark holds in the initial state. This leads to non-deterministic behaviour and partly ignores useful information (i.e., all reasonable orderings we could find but are not found due to early termination).

This bug cannot be fixed without changing the logic of the code. The following changes are part of this pull request:
- Landmarks are considered `reached` when they hold in a state for the first time. 
- Previously, landmarks were only `reached` when also all predecessors in the landmark-graph were `reached`, but this check was non-deterministic. Now, this is checked in the `needed_again` criterion and it is deterministic.
- Reasonable orders are only added if the involved landmarks cannot be reached simultaneously (i.e., they have no common possible achiever).


src/search/landmarks/landmark_factory.cc
```diff
@@ -389,6 +397,18 @@ void LandmarkFactory::approximate_reasonable_orders(
     }
 }
 
+bool LandmarkFactory::have_common_achiever(
+    const LandmarkNode *node_a, const LandmarkNode *node_b) const {
+    set<int> achievers_a = node_a->possible_achievers;
```
thomaskeller79, line 402: Is it on purpose that these are copied? I think this should be a reference instead.

src/search/landmarks/landmark_factory.cc
```diff
@@ -339,14 +339,18 @@ void LandmarkFactory::approximate_reasonable_orders(
...
-            return;
+        if (obedient_orders && node_p->is_true_in_state(initial_state))
+            continue;
 
         if (!obedient_orders && node_p->is_true_in_goal) {
             for (auto &node2_p : lm_graph->get_nodes()) {
                 if (node2_p == node_p || node2_p->disjunctive)
                     continue;
-                if (interferes(task_proxy, node2_p.get(), node_p.get())) {
+                if (node_p->is_true_in_state(initial_state)
```
thomaskeller79, line 349: I find it nicer to either catch all three "continue" cases by themselves or all at once (and I prefer to catch them all at once). The 2-to-1 split seems arbitrary.

src/search/landmarks/landmark_factory.cc
```diff
@@ -378,7 +382,11 @@ void LandmarkFactory::approximate_reasonable_orders(
             for (LandmarkNode *node : interesting_nodes) {
                 if (node == node_p.get() || node->disjunctive)
                     continue;
-                if (interferes(task_proxy, node, node_p.get())) {
+                if (node->is_true_in_state(initial_state)
```
thomaskeller79, line 385: Same comment as before

src/search/landmarks/landmark_factory.cc
```diff
@@ -339,14 +339,18 @@ void LandmarkFactory::approximate_reasonable_orders(
...
         if (!obedient_orders && node_p->is_true_in_goal) {
             for (auto &node2_p : lm_graph->get_nodes()) {
                 if (node2_p == node_p || node2_p->disjunctive)
                     continue;
-                if (interferes(task_proxy, node2_p.get(), node_p.get())) {
+                if (node_p->is_true_in_state(initial_state)
+                    && node2_p->is_true_in_state(initial_state))
+                    continue;
+                if (interferes(task_proxy, node2_p.get(), node_p.get())
+                    && !have_common_achiever(node2_p.get(), node_p.get())) {
```
thomaskeller79, line 353: Is this the only case that could introduce an ordering a -> b where a and b can be achieved simultaneously? If not, I think it would make more sense to resolve this issue differently, e.g., by allowing landmark factories to introduce orderings that can be achieved simultaneously and by offering an option that removes all orderings that are not strict.

Note also that we lose information here that could be exploited: if we have (fact) landmarks a and b and an operator o_1 that achieves both simultaneously and operators o_2 and o_3 that achieve both independently, we'd actually like to express that we have to apply o_1 or o_2 and o_3.

ClemensBuechner: I agree, checking this here is probably the wrong place. 
Looking at this right now, I would remove the check completely from this issue. When implementing it, I thought it was needed for ensuring that LM-count is goal-aware. With the now deterministic update of `reached`, however, this should not be a problem (I think). In order to make this patch about fixing the bug in the first place (opposed to introducing this new feature), I will remove the check for common achievers again.

src/search/landmarks/landmark_factory.cc
ClemensBuechner: A similar comment to what I wrote about the common achievers applies here: should we only fix the bug or make the implementation more reasonable (no pun intended)?
Option 1: Keep the check `node_p->is_true_in_state(initial_state)` where it was, right at the start of the loop, and replace the `return` with a `continue`.
Option 2: Do it like it's implemented in this patch to allow orders A->B where B is true initially.

On the one side, option 1 does preserve the semantics of the current code, but does not terminate too early anymore. On the other hand, I see no good argument for disregarding orderings A->B where B is initially true, so we might as well change to option 2 in this issue.

PR #24, opened 2021-02-09
[issue1003] Adapt Github actions to actually test what we want to test.

silvansievers:


.github/workflows/windows.yml
```diff
@@ -1,71 +1,95 @@
...
-name: windows-latest
+name: windows
 
 on: [push, pull_request]
+
 env:
   ARCH: "x64"
-  VC: "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvarsall.bat"
+  VC2019: "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvarsall.bat"
+  VC2016: "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Enterprise\\VC\\Auxiliary\\Build\\vcvarsall.bat"
```
PatrickFerber, line 9: Rename VC2016 to VC2017?

.github/workflows/windows.yml
```diff
@@ -1,71 +1,95 @@
...
+          CC: cl
+          CXX: cl
         run: |
-          "${{ env.VC }}" ${{ env.ARCH }} & python build.py ${{ matrix.build-version }}
+          IF "${{ matrix.windows-version }}" == "windows-2016" (
+            "${{ env.VC2016 }}" ${{ env.ARCH }} & python build.py ${{ matrix.build-version }}
+          )
+          IF "${{ matrix.windows-version }}" == "windows-2019" (
+            "${{ env.VC2019 }}" ${{ env.ARCH }} & python build.py ${{ matrix.build-version }}
+          )
```
PatrickFerber, line 39: I take a look, if we can improve this, because, I have to call those scripts frequently in my issue

.github/workflows/ubuntu.yml
```diff
@@ -3,27 +3,24 @@ name: ubuntu
...
+          - {cc: clang-11, cxx: clang++-11}
         python-version: [3.6]
         exclude:
           - ubuntu-version: ubuntu-18.04
-            compiler-version: gcc-10
+            compiler-version: {cc: gcc-10, cxx: g++-10}
           - ubuntu-version: ubuntu-18.04
-            compiler-version: clang-11
+            compiler-version: {cc: clang-11, cxx: clang++-11}
     steps: # each - is a new sequentially run step
```
PatrickFerber, line 24: Would be nice, if we do not need to repeat the whole dictionary here, but could name the dicts, but it it is not possible, we can live with it (Silvan already said he looked into it and found no way)

.github/workflows/mac.yml
```diff
@@ -45,12 +45,10 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: ${{ matrix.python-version }}
-          architecture: x64
```
jendrikseipp, line 48: Can we remove the architecture line above as well?

.github/workflows/mac.yml
```diff
@@ -45,12 +45,10 @@ jobs:
         uses: actions/setup-python@v1
```
jendrikseipp, line 45: Do we really need to set up Python again?

.github/workflows/mac.yml
```diff
@@ -45,12 +45,10 @@ jobs:
...
         with:
           python-version: ${{ matrix.python-version }}
-          architecture: x64
 
       - name: setup
         run: |
           pip3 install pytest tox
```
jendrikseipp, line 51: pytest is not needed here. tox will install it inside its venvs.

.github/workflows/mac.yml
```diff
@@ -45,12 +45,10 @@ jobs:
...
-          #brew install valgrind # TODO: does not work
           mkdir builds
 
       - name: download-planner
```
jendrikseipp, line 54: I'd use `tox -e py,search` below. This will run the exitcodes test and run the standard configs, which both have some OS-dependent code. I don't think we need to test the Python translator code on macOS.

.github/workflows/ubuntu.yml
```diff
@@ -3,28 +3,27 @@ name: ubuntu
...
+        # name the compiler versions so that we don't have to copy them here.
         exclude:
           - ubuntu-version: ubuntu-18.04
-            compiler-version: gcc-10
+            compiler-version: {cc: gcc-10, cxx: g++-10}
           - ubuntu-version: ubuntu-18.04
-            compiler-version: clang-11
-    steps: # each - is a new sequentially run step
+            compiler-version: {cc: clang-11, cxx: clang++-11}
+    steps: # Eeach - is a new sequentially run step
```
jendrikseipp, line 26: Typo: Each.

Also, I'd use `"-"` instead of `-`. I always stumble over this, when I read this sentence.

.github/workflows/ubuntu.yml
```diff
@@ -37,23 +36,16 @@ jobs:
...
-          export CC=${{ matrix.compiler-version }}
-          if [[ $CC == gcc* ]]; then
-            export CXX="$(echo "${CC}" | sed "s/cc/++/g")";
-          elif [[ $CC == clang* ]]; then
-            export CXX="$(echo "${CC}" | sed "s/clang/clang++/g")";
-          else
-            echo "Unknown compiler version";
-            exit 1;
-          fi
+          sudo apt-get -y install ${{ matrix.compiler-version.cc }}
```
jendrikseipp, line 39: I'd combine the two apt calls.

.github/workflows/ubuntu.yml
```diff
@@ -37,23 +36,16 @@ jobs:
         run: |
           sudo apt-get update
```
jendrikseipp, line 37: The update step can take really long on GitHub Actions. I'd simply remove it.

.github/workflows/ubuntu.yml
```diff
@@ -37,23 +36,16 @@ jobs:
...
+          sudo apt-get -y install ${{ matrix.compiler-version.cc }}
           mkdir /home/runner/work/downward/lib
 
       # We only want to setup osi if both LP solvers are set, hence
       # we execute the following three steps only if both secrets
       # are set.
       - name: setup-cplex
         env:
+          CC: ${{ matrix.compiler-version.cc }}
+          CXX: ${{ matrix.compiler-version.cxx }}
```
jendrikseipp, line 48: I think you could also set an environment variable that applies to all steps in a single place: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

.github/workflows/ubuntu.yml
```diff
@@ -156,17 +154,17 @@ jobs:
       - name: archive-files
```
jendrikseipp, line 154: Some of the `make` calls use the -j argument. I'd use -j2 for all of them since there are usually two CPUs available. (Can be double-checked with `python3 -c "import multiprocessing as m; print('CPUs:', m.cpu_count())"`)

jendrikseipp, line 154: I think the `ls` call for VAL can be removed.

.github/workflows/ubuntu.yml
```diff
@@ -156,17 +154,17 @@ jobs:
...
-        if: ${{ matrix.compiler-version }} == gcc
+        if: ${{ matrix.compiler-version.cc == 'gcc' }}
         uses: actions/upload-artifact@v1
         with:
           name: compiled-planner-${{ matrix.ubuntu-version }}
-          path: /home/runner/work/downward/${{ matrix.compiler-version }}.tar.gz
+          path: /home/runner/work/downward/${{ matrix.compiler-version.cc }}.tar.gz
 
 
   test:
```
jendrikseipp, line 170: Why do we need to specify and install the compiler for the tests?

jendrikseipp, line 170: I'd remove the "apt-get update" call below.

jendrikseipp, line 170: Instead of the if/else block at the end of the file, you can use "tox -e py,..." and tox will use the available Python version.

.github/workflows/windows.yml
```diff
@@ -1,71 +1,95 @@
...
+          CC: cl
+          CXX: cl
         run: |
           cd misc/
-          "${{ env.VC }}" ${{ env.ARCH }} & tox -e search,translator
+          IF "${{ matrix.windows-version }}" == "windows-2016" (
+            "${{ env.VC2017 }}" ${{ env.ARCH }} & tox -e search,translator
+          )
+          IF "${{ matrix.windows-version }}" == "windows-2019" (
+            "${{ env.VC2019 }}" ${{ env.ARCH }} & tox -e search,translator
```
jendrikseipp, line 92: I'd use "tox -e py,search" in both cases.

jendrikseipp (changes requested):
Nice work! Sending the buildbot to retirement is now one step closer :-)

.github/workflows/mac.yml
silvansievers: I think we need to install a python 3 version so that we have pip3 available.

.github/workflows/mac.yml
silvansievers: yes

.github/workflows/mac.yml
silvansievers: Does the translator include the py test? If yes, I don't think it hurts to run the translator tests here, too.They are fast, I tihnk.

.github/workflows/ubuntu.yml
silvansievers: I think because some tox tests (search?) also check if the code is built. If CMAKE doesn't find compilers, it complains. Maybe we would not need to specify a spefific version, but it also doesn't hurt to always use the same setup.

.github/workflows/windows.yml
silvansievers: But how does it use the right python version then? If tox installs its own python environment?

.github/workflows/mac.yml
jendrikseipp: The "py" tests call the "testenv" section of tox, so in our case:

```
  pytest driver/tests.py misc/tests/test-exitcodes.py
  bash -c "python3 misc/autodoc/autodoc.py --dry-run > /dev/null"
```

I'm fine with running the translator tests on macOS, so "tox -e py,translator,search".

.github/workflows/mac.yml
jendrikseipp: What surprises me is that it's done twice (if I'm not overlooking anything).

.github/workflows/ubuntu.yml
jendrikseipp: I see. Can you add this info as a comment so we don't wonder about this in the future?

.github/workflows/mac.yml
silvansievers: Once for the first job, once for the second job. Jobs are run independently of each other.

.github/workflows/mac.yml
silvansievers: The py tests fail because we don't have validate installed and don't know how to do so. Unless you have a better idea, I will stick with translator and search tests.

jendrikseipp: Ah, thanks! I think it would be good to add this as a comment.

.github/workflows/mac.yml
jendrikseipp: That makes sense. I missed that these are two separate jobs.

.github/workflows/ubuntu.yml
```diff
@@ -3,28 +3,27 @@ name: ubuntu
...
         ubuntu-version: [ubuntu-18.04, ubuntu-20.04]
-        compiler-version: [gcc, gcc-10, clang, clang-11]
+        compiler-version:
+          - {cc: gcc, cxx: g++}
+          - {cc: gcc-10, cxx: g++-10}
+          - {cc: clang, cxx: clang++}
+          - {cc: clang-11, cxx: clang++-11}
         python-version: [3.6]
+        # Unfortunately, we couldn't figure out if there is a way to
+        # name the compiler versions so that we don't have to copy them here.
```
jendrikseipp, line 20: I just noticed that the debug builds are missing in this workflow.

.github/workflows/ubuntu.yml
```diff
@@ -150,23 +148,22 @@ jobs:
...
+          tar cfz ${{ matrix.compiler-version.cc }}.tar.gz downward lib
 
       - name: upload-files
-        if: ${{ matrix.compiler-version }} == gcc
+        if: ${{ matrix.compiler-version.cc == 'gcc' }}
         uses: actions/upload-artifact@v1
         with:
           name: compiled-planner-${{ matrix.ubuntu-version }}
-          path: /home/runner/work/downward/${{ matrix.compiler-version }}.tar.gz
+          path: /home/runner/work/downward/${{ matrix.compiler-version.cc }}.tar.gz
```
jendrikseipp, line 166: I just pushed the current main branch to a private repo and even though only the macOS and Windows build steps succeeded, 2.1 GB of artifacts were uploaded. These artifacts are actually retained for 90 days and the limit is 2 GB. Not sure where the 2.1 GB come from. The debug and release Windows tarballs come in at 200 MB and the macOS tarballs are both 50 MB in size. In any case, we have to delete the artifacts after a job is done, I think. (That's not a solution for jobs with failing tests though.) Or we try to find a solution that doesn't store artifacts in the first place.

.github/workflows/ubuntu.yml
```diff
@@ -205,29 +211,16 @@ jobs:
...
-          python_version=${{ matrix.python-version }}
-          if [[ $python_version=="3.6" ]]; then
-            tox -e py36,translator,search,style
-          elif [[ $python_version=="3.8" ]]; then
-            tox -e py38,translator,search,style
-          else
-            echo "Unknown Python version";
-            exit 1;
-          fi
+          tox -e py,translator,search,style
```
jendrikseipp, line 224: I think the search and style tests should only be run once, not thrice. They can take up quite some time.

jendrikseipp (changes requested):

.github/workflows/ubuntu.yml
silvansievers: No, for every matrix strategy, build for debug and release. See below.

.github/workflows/ubuntu.yml
silvansievers: Using a private repo to run these actions is not a good idea, because also the runtime limit will be reached after a few runs at last. I don't think this is a problem for us.

With the ubuntu artifact, we could certainly only upload include the driver instead of the whole code, but I doubt this will reduce size a lot. We upload the entire installation of external libraries which probably consume a lot of space. It seems to me that deleting artifacts is the right solution.

.github/workflows/ubuntu.yml
silvansievers: What do you mean? Are they part of the py test?

jendrikseipp: No, the tox call is excecuted for 18.04-3.6, 20.04-3.6 and 20.04-3.8 if I'm not overlooking anything.

.github/workflows/ubuntu.yml
```diff
@@ -187,6 +184,15 @@ jobs:
         with:
```
jendrikseipp, line 184: I'd also exclude the combination Ubuntu 20.04 with Python 3.6. I think it makes sense to test both Ubuntu versions only with their default Python version.

.github/workflows/ubuntu.yml
jendrikseipp: > Using a private repo to run these actions is not a good idea, because also the runtime limit will be reached after a few runs at last. I don't think this is a problem for us.

I agree. This is the default, however. At the moment users would have to disable GitHub Actions for all private repos manually.

Regarding artifacts, I see several TODOs:
1. We should set the retention time to the minimum: 1 day. This is done by adding `retention-days: 1` to the upload step (https://github.blog/changelog/2020-10-08-github-actions-ability-to-change-retention-days-for-artifacts-and-logs/).
2. We should delete artifacts after the tests finish (ideally for failed tests as well).
3. We should try to make the tarballs smaller. Currently they are so large (> 700 MB sometimes) that uploading them takes 30s-60s. At the bottom of this page (https://github.com/aibasel/downward/actions/runs/558175941) you can see the artifact sizes for Windows builds.

I wonder whether the whole artifact dance is really needed. Can't we just use an if-statement to only run the tests for a subset of the matrix combinations?

.github/workflows/ubuntu.yml
PatrickFerber: As Jendrik said, you can do this jsut at the top of the script. You can take a look at my issue1005

.github/workflows/windows.yml
PatrickFerber: Found a nicer way. See issue1005. If you want you can change it, but I do not require it, as I will change it soon after :)

.github/workflows/ubuntu.yml
jendrikseipp: My bad, you're right. But wouldn't it make sense to add the "release" vs. "debug" split to the build matrix as for Windows and macOS?

.github/workflows/ubuntu.yml
FlorianPommerening: I have not really looked into this but from dealing with the IPC containers I remember that the size of the object files has a large impact. So calling `./build.py` and then zipping `builds` created a much larger package than just zipping `builds/release/bin`.

.github/workflows/ubuntu.yml
PatrickFerber: Initially, we split the compilation of release and debug, because we could parallize this well(16min -> 2x8min). The setup for the compilation is negligible. The uploading and downloading of the artifacts takes some time, but is still a speedup (+2-4min)
For Ubuntu, we also install the LP solvers. This takes quite some time (+8 minutes), thus, our impression is that the speed up from parallelization is not making up for the additional wasted resources.

(In issue1005 we add LP support to windows and I removed the parallelization there too)

jendrikseipp: Ah, I see, thanks!

.github/workflows/ubuntu.yml
```diff
@@ -3,28 +3,27 @@ name: ubuntu
 
 on: [push, pull_request]
```
jendrikseipp, line 4: I think all jobs should immediately abort if the target repo is not "aibasel/downward". Then users who want to run (a subset of) the tests can do so manually.

.github/workflows/ubuntu.yml
jendrikseipp: Also, we're currently missing the clang-tidy and valgrind tests.

.github/workflows/ubuntu.yml
```diff
@@ -205,29 +211,16 @@ jobs:
...
       - name: test
         env:
+          CC: ${{ matrix.compiler-version.cc }}
+          CXX: ${{ matrix.compiler-version.cxx }}
           DOWNWARD_CPLEX_ROOT: /home/runner/work/downward/lib/ibm/ILOG/CPLEX_Studio129/cplex
           DOWNWARD_SOPLEX_ROOT: /home/runner/work/downward/lib/soplex-3.1.1
           DOWNWARD_COIN_ROOT: /home/runner/work/downward/lib/coin
         run: |
           export PATH="$(pwd):$PATH"  # Add uncrustify and VAL to PATH.
           cd misc/
```
jendrikseipp, line 223: One more thing: I think all tests that don't require compilation (style, clang-tidy, translator) should be run before the compilation. The style and clang-tidy tests are the ones that most frequently fail and it makes no sense to spend all the cycles for useless compilations. Quicker feedback will also speed up our development process.

.github/workflows/ubuntu.yml
silvansievers: Do you know how to do this? I didn't find any events that care about the repository. Also, I don't think the solution to abort is nice because it will still run actions on forks and create lots of emails saying that an action failed.

.github/workflows/ubuntu.yml
silvansievers: > I wonder whether the whole artifact dance is really needed. Can't we just use an if-statement to only run the tests for a subset of the matrix combinations?

We use them for testing different python versions without installing and compiling everything just for testing several python versions. That is, we upload one compiled version and use it with several tests. We can get rid of this if we are fine with running the same setup steps for testing different pythons. Or maybe we don't need to test running the planner with all python versions?

.github/workflows/ubuntu.yml
silvansievers: I don't understand how this goes with our goal of supporting Python 3.9. We decided to test all Python versions within their lifespan. Currently, we do this by testing the oldest and latest version supported in Fast Downward, which is 3.6 and 3.8. With the compilers, we test the default compiler on the older Ubuntu and the default and the newest version on the newer Ubuntu release. This just mimics this behavior.

.github/workflows/ubuntu.yml
silvansievers: Thanks to our quick discussion on discord, I now understand what the py test does. But why do we perform exactly the exit code and autodoc test with different python versions? I think I don't understand what exactly we want to test with different python versions: all of the tox tests? all of the python scripts (i.e., translator and driver)? both? Figuring this out could also potentially help with the artifacts problem.

.github/workflows/ubuntu.yml
silvansievers: That makes sense. However, I would like to figure out first what exactly we want to test under which setting. Do we need the tests you mentioned here only once, with any python version? Does it matter at all?

.github/workflows/ubuntu.yml
PatrickFerber: I found a page that says we can add an if condition to the jobs. I tested it on my work in branch GHATest (use != for not equal). This creates an entry for the action, but the action is denoted as skipped and no mail is send

```
jobs:
  compile: # name of the job. Jobs run in parallel unless specified otherwise.
    name: compile
    if: github.repository == 'PatrickFerber/downward'
```
https://github.community/t/stop-github-actions-running-on-a-fork/17965

silvansievers: Thanks for looking this up. I found something along those lines before. I just don't think it makes sense to encode into the workflows themselves that they should be run/not run on specific repositories because a user then would need to change the workflows to enable them which doesn't make sense in my opinion.

.github/workflows/ubuntu.yml
PatrickFerber: About Silvans last point, do we need to test different versions. My thoughts are:
We have to test the last supported python version. Python might introduce a nice python feature, I get used to it. Then use this feature in the FD driver and we lose support for python3.6. Thus, we have to test our last supported python version. 
If the python developer do not change the behaviour of a function for different python 3.X releases, anything that works for 3.6 should also work for newer versions (unless there is a bug introduced or fixed in python).

Does the behaviour of functions change or is there another reason why code that works for 3.6 should not work anymore for >3.6

.github/workflows/ubuntu.yml
PatrickFerber: My 5cents:
- Everything that we run to verify that a functionality works (i.e. translator tests) should be run with all relevant python versions (see comments above about which python we need to test actually)
- Everything that ensures code quality is sufficient to run on a single python version (whether the script fails on some python version is not important, important is to know if the code quality is right). Ensuring that those scripts work on any python version is nice for the developers, because we can use it on our machines (although, once it breaks with a new python version, we would just push a fix)

.github/workflows/ubuntu.yml
jendrikseipp: Yes, we need to test the driver using different Python versions. What I meant was to move all steps into one job, add Python 3.6 and 3.8 to the matrix, and add `if python_version == 3.6` to all compilation steps and `if cc == gcc and cxx == g++` to all test steps (haven't checked the if-syntax).

.github/workflows/ubuntu.yml
jendrikseipp: There's some Python code that exists solely for testing the C++ code (search, clang-tidy, valgrind). We only need to run it using a single Python version. The remaining Python code (driver incl. exit codes, autodoc, translator) needs to be tested with multiple Python versions. (So yes, the translator tests should ideally be under the default testenv in tox.ini. This is an oversight that comes from the fact that theses tests used to compare different Python versions instead of the same Python version with different random seeds.)

.github/workflows/ubuntu.yml
jendrikseipp: If this goes against our rules, we need to rethink them. I believe testing the same Python version on different Ubuntu versions is wasting CPU cycles. I also wouldn't compile the planner twice for the same GCC version on different Ubuntu versions.

.github/workflows/ubuntu.yml
jendrikseipp: I still think having the tests be opt-in is nicer and less surprising than opt-out. If a user wants them they need adapting anyway because otherwise the LP tests will fail.

.github/workflows/ubuntu.yml
jendrikseipp: Ideally, the style and clang-tidy tests would only run once per push (not once per OS or once per matrix config) and all other jobs would only execute if they're successful. That would require, however, that all jobs are in the same workflow (which makes for a long file, but I think that's negligible).

maltehelmert: Agreed on the "only run once per push". But is there really a good motivation for not trying the other tests if style or clang-tidy fail?

.github/workflows/ubuntu.yml
silvansievers: That's not right: the LP compilation will be skipped. I think the whole thing warrants another discussion.

jendrikseipp: In this PR, yes, it will be skipped. But is this desirable? In issue1005 we're planning to make this an error. Otherwise we might miss a failed LP compilation.

silvansievers: We already discussed this when Florian got the LP errors because he hadn't set up the secrets. I think this is a good solution, yes. Note that when we merge the branch, the LP solvers will be tested even if the person creating the PR does not test them because she doesn't have the secrets.

.github/workflows/ubuntu.yml
jendrikseipp: Usually, the failing tests will need to be run again anyway and I think it makes sense to save those CPU cycles.

maltehelmert: If I pushed something in order to get it tested, then came back a few hours later to see that the tests didn't run because I inserted an extra newline, I'd probably be annoyed. But if the people that work on the code more than me prefer this, I have no objections.

.github/workflows/ubuntu.yml
```diff
@@ -1,76 +1,67 @@
...
+            compiler-version: {cc: clang-11, cxx: clang++-11}
+    env:
+      CC: ${{ matrix.compiler-version.cc }}
+      CXX: ${{ matrix.compiler-version.cxx }}
+      CPLEX_URL: ${{ secrets.CPLEX_URL }}
+      SOPLEX_URL: ${{ secrets.SOPLEX_URL }}
+      DOWNWARD_CPLEX_ROOT: /home/runner/work/downward/lib/ibm/ILOG/CPLEX_Studio129/cplex
+      DOWNWARD_SOPLEX_ROOT: /home/runner/work/downward/lib/soplex-3.1.1
+      DOWNWARD_COIN_ROOT: /home/runner/work/downward/lib/coin
+    steps: # Eeach "-" is a new sequentially run step
```
jendrikseipp, line 35: Typo.

.github/workflows/ubuntu.yml
```diff
@@ -109,125 +94,102 @@ jobs:
...
 
-      - name: extract-files
-        # We need to make sure that paths are the same as in the first job,
-        # otherwise cmake exits with an error when called during tests.
-        # Alternatively, we could change the tests so that they don't build.
+      - name: Extract archive
+        # We need to make sure that paths are the same as in the first
+        # job, otherwise cmake exits with an error when called during
+        # tests. Alternatively, we could change the tests so that they
+        # don't build.
```
jendrikseipp, line 169: This should be resolved now, right?

.github/workflows/ubuntu.yml
```diff
@@ -109,125 +94,102 @@ jobs:
...
+        run: |
+          git clone https://github.com/KCL-Planning/VAL.git
+          cd VAL
+          git checkout a5565396007eee73ac36527fbf904142b3077c74
+          make clean  # Remove old build artifacts and binaries.
+          sed -i 's/-Werror //g' Makefile  # Ignore warnings.
+          make -j2
+          mv validate ../
+          cd ../
+          rm -rf VAL
```
jendrikseipp, line 158: You can make VAL available on PATH with a single call here: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#adding-a-system-path

.github/workflows/ubuntu.yml
```diff
@@ -109,125 +94,102 @@ jobs:
...
-      - name: upload-files
-        if: ${{ matrix.compiler-version }} == gcc
-        uses: actions/upload-artifact@v1
+      - name: Upload archive
+        if: ${{ matrix.compiler-version.cc == 'gcc' }}
+        uses: actions/upload-artifact@master
         with:
           name: compiled-planner-${{ matrix.ubuntu-version }}
-          path: /home/runner/work/downward/${{ matrix.compiler-version }}.tar.gz
+          path: /home/runner/work/downward/archive.tar.gz
```
jendrikseipp, line 122: Let's add a retention time limit of 1 day and/or delete the artifact at the end of the workflow.

.github/workflows/windows.yml
```diff
@@ -1,71 +1,103 @@
...
+        python-version: [3.6]
+        build-version: [release, debug]
     steps: # each - is a new sequentially run step
-      - uses: actions/checkout@master
+      - name: clone-repo
+        uses: actions/checkout@master
 
-      - name: Setup python
-        uses: actions/setup-python@v1
+      - name: setup-python
```
jendrikseipp, line 25: Set up Python

driver/tests.py
```diff
@@ -18,13 +19,13 @@
 
 def translate():
     """Create translated task."""
-    cmd = ["./fast-downward.py", "--translate",
+    cmd = [sys.executable, "fast-downward.py", "--translate",
```
jendrikseipp, line 22: Can we factor this (sys.executable and "fast-downward.py" ) out into run_driver()?

driver/tests.py
```diff
@@ -42,32 +43,37 @@ def test_commandline_args():
...
+        cmd = [sys.executable, "fast-downward.py", "--portfolio", portfolio,
                "--search-time-limit", "30m", "output.sas"]
         run_driver(cmd)
 
 
 def test_hard_time_limit():
+    # We cannot test this on systems like Windows where we cannot
+    # enforce time limits.
+    if not limits.can_set_time_limit():
+        return
```
jendrikseipp, line 65: We could use the decorator `@pytest.skipif(not limits.can_set_time_limit())` for this, I think. That gives us nicer test output and is "more standard".

jendrikseipp (changes requested):

.github/workflows/ubuntu.yml
silvansievers: We need to get the library paths right. I updated the comment.

driver/tests.py
silvansievers: I did, but needed another special case for `test_hard_time_limit`.

.github/workflows/ubuntu.yml
```diff
@@ -1,76 +1,67 @@
...
-## globally. Github actions can set environment variables globally,
-## but they can only be single values, not lists, and ideally we would
-## like to have something like COMPILER_VERSIONS = [gcc, ..].
-## Now, whenever we change versions, we have to remember to do this
-## *everywhere* in this file.
-
 jobs:
-  compile: # identifier of the job. Jobs run in parallel unless specified otherwise.
-    name: compile
+  compile: # Identifier of the job. Jobs run in parallel unless specified otherwise.
```
PatrickFerber, line 7: Can we just remove this comment '# Identifier of ...."

.github/workflows/ubuntu.yml
```diff
@@ -1,76 +1,67 @@
...
+            compiler-version: {cc: clang-11, cxx: clang++-11}
+    env:
+      CC: ${{ matrix.compiler-version.cc }}
+      CXX: ${{ matrix.compiler-version.cxx }}
+      CPLEX_URL: ${{ secrets.CPLEX_URL }}
+      SOPLEX_URL: ${{ secrets.SOPLEX_URL }}
+      DOWNWARD_CPLEX_ROOT: /home/runner/work/downward/lib/ibm/ILOG/CPLEX_Studio129/cplex
+      DOWNWARD_SOPLEX_ROOT: /home/runner/work/downward/lib/soplex-3.1.1
+      DOWNWARD_COIN_ROOT: /home/runner/work/downward/lib/coin
+    steps: # Each "-" is a new sequentially run step
```
PatrickFerber, line 35: Same for this comment "# Each ..."

.github/workflows/ubuntu.yml
```diff
@@ -1,76 +1,67 @@
...
         if: ${{ env.CPLEX_URL != 0 && env.SOPLEX_URL != 0 }}
+        env:
+          CPLEX_STUDIO: /home/runner/work/downward/lib/ibm/ILOG/CPLEX_Studio129
         run: |
           # We redirect output of wget to hide the secret URLs.
           wget $CPLEX_URL &> /dev/null
           export CPLEX_INSTALLER=cplex_studio129.linux-x86-64.bin
           chmod +x $CPLEX_INSTALLER
-          ./$CPLEX_INSTALLER -DLICENSE_ACCEPTED=TRUE -DUSER_INSTALL_DIR=${DOWNWARD_CPLEX_ROOT} -i silent
+          ./$CPLEX_INSTALLER -DLICENSE_ACCEPTED=TRUE -DUSER_INSTALL_DIR=${CPLEX_STUDIO} -i silent
```
PatrickFerber, line 61: Did you detect this due to the new LP solver tests?
Now we have two variables which are dependent on another. If we ever have to change one, we also have to change the other. I suggest:
1. remove CPLEX_STUDIO from the env
2. ${CPLEX_STUDIO} -> "$(dirname "${DOWNWARD_CPLEX_ROOT}")"
- if you think the line is too long or ugly, just store it in a variable, like CPLEX_STUDIO
(the amount of " symbols is correct :)

.github/workflows/ubuntu.yml
```diff
@@ -109,125 +94,116 @@ jobs:
...
+          tar xfz archive.tar.gz
           shopt -s dotglob
-          mv downward/* ../
-          mv lib/ ../../
-          cd ../
-          rm -r compiled-planner-${{ matrix.ubuntu-version }}
-
-      - name: setup-dependencies
+          mv downward/* .
+          mv lib/ ../
```
PatrickFerber, line 179: Just as a reminder to delete this at some point

.github/workflows/ubuntu.yml
```diff
@@ -109,125 +94,116 @@ jobs:
...
-      - name: test
-        env:
-          DOWNWARD_CPLEX_ROOT: /home/runner/work/downward/lib/ibm/ILOG/CPLEX_Studio129/cplex
-          DOWNWARD_SOPLEX_ROOT: /home/runner/work/downward/lib/soplex-3.1.1
-          DOWNWARD_COIN_ROOT: /home/runner/work/downward/lib/coin
+      - name: Run CPLEX tests
+        if: ${{ env.CPLEX_URL != 0 && env.SOPLEX_URL != 0 }}
+        run: |
+          cd misc/
+          tox -e cplex
```
PatrickFerber, line 202: add empty line

README.md
```diff
@@ -11,6 +11,24 @@ For further information:
 - Fast Downward main repository: <https://github.com/aibasel/downward>
 
 
+## Tested software versions
+
+This version of Fast Downward has been tested with the following systems,
+compilers and tools:
+
```
PatrickFerber, line 18: The comment below is not telling me which combinations are tested, especially, this looks as if we would already test Windows and MacOS with both LP solvers. 
I would at least at a comment about the supported LP solvers, e.g. add to Ubuntu 'with CPLEX and SoPlex' (I will do that for Windows.

driver/arguments.py
```diff
@@ -84,7 +84,7 @@
 Examples:
 
 %s
-""" % "\n\n".join("%s\n%s" % (desc, " ".join(cmd)) for desc, cmd in EXAMPLES)
+""" % "\n\n".join("%s\n%s" % (desc, " ".join(["./fast-downward.py"] + parameters)) for desc, parameters in EXAMPLES)
```
PatrickFerber, line 87: add sys.executable prior to ./fast-downward.py?

driver/tests.py
```diff
@@ -18,59 +19,63 @@
...
         limits.set_time_limit(10)
 
-    cmd = [
-        "./fast-downward.py", "--translate", "--translate-time-limit",
+    driver = [sys.executable, "fast-downward.py"]
+    parameters = [
+        "--translate", "--translate-time-limit",
         "10s", "misc/tests/benchmarks/gripper/prob01.pddl"]
-    subprocess.check_call(cmd, preexec_fn=preexec_fn, cwd=REPO_ROOT_DIR)
+    subprocess.check_call(driver+parameters, preexec_fn=preexec_fn, cwd=REPO_ROOT_DIR)
```
PatrickFerber, line 72: whitespaces around +

driver/tests.py
```diff
@@ -18,59 +19,63 @@
...
 
-    cmd = [
-        "./fast-downward.py", "--translate", "--translate-time-limit",
+    parameters = [
+        "--translate", "--translate-time-limit",
         "20s", "misc/tests/benchmarks/gripper/prob01.pddl"]
     with pytest.raises(subprocess.CalledProcessError) as exception_info:
-        subprocess.check_call(cmd, preexec_fn=preexec_fn, cwd=REPO_ROOT_DIR)
+        subprocess.check_call(driver+parameters, preexec_fn=preexec_fn, cwd=REPO_ROOT_DIR)
```
PatrickFerber, line 78: whitespaces around +

.github/workflows/windows.yml
```diff
@@ -1,71 +1,67 @@
...
+        # steps, we hope to be able to install VAL natively on Windows.
         run: |
           cd misc/
-          "${{ env.VC }}" ${{ env.ARCH }} & tox -e search,translator
+          IF "${{ matrix.windows-version }}" == "windows-2016" (
+            "${{ env.VC2017 }}" ${{ env.ARCH }} & tox -e translator,search
+          )
+          IF "${{ matrix.windows-version }}" == "windows-2019" (
+            "${{ env.VC2019 }}" ${{ env.ARCH }} & tox -e translator,search
+          )
```
PatrickFerber, line 65: do your driver tests now also support Windows? If yes, can you please add them here to see if they work correctly.

.github/workflows/ubuntu.yml
silvansievers: No, I realized this when archiving libraries. I'll do the suggested change.

driver/arguments.py
silvansievers: Nope, this is just print to command line when running `fast-downward.py --help`.

.github/workflows/windows.yml
silvansievers: No, unfortunately not. We still need VAL. See comment above.

.github/workflows/windows.yml
```diff
@@ -1,71 +1,67 @@
...
         with:
           python-version: ${{ matrix.python-version }}
-          architecture: ${{ env.ARCH }}
 
-      - name: compile
+      - name: Compile planner
         shell: cmd
+        env:
+          CC: cl
+          CXX: cl
```
jendrikseipp, line 33: Shall we set these env vars globally for the workflow?

jendrikseipp (changes requested):

.github/workflows/windows.yml
silvansievers: issue1005 will change this file completely; not worth for this issue.

driver/arguments.py
maltehelmert: Was there something broken about the old version? "./" makes the assumption that people are calling this from the current directory, which doesn't have to be true. One could use sys.args[0] to find out how they actually called it, but the actual unix convention is to:

1) compute the filename programmatically rather than hard-coding it (i.e., use sys.args[0]), but
2) strip the path

The advantage of 1) is that it adapts if people rename or symlink the file. The motivation for 2) is unclear to me, but nevertheless it's the expected behaviour.

So according to this convention, we should use `os.path.basename(sys.args[0])`.

misc/autodoc/autodoc.py
```diff
@@ -164,8 +161,6 @@ def add_page(title, text):
 
 if __name__ == '__main__':
     args = parse_args()
-    logging.info("building planner...")
-    build_planner(args.build)
```
maltehelmert, line 168: Has the buildbot been adapted for this change, or will this break our wiki?

driver/arguments.py
maltehelmert: PS: I see now that the old version was no better and the new version is better than the old one. (github didn't give enough context for this, but I've now looked at the full pull request.) Still, I suggest if we're improving this that we also make the change from my previous message.

silvansievers: Using `os.path.basename(sys.args[0])` makes `./fast-downward.py --help` called from the same directory print lines like `fast-downward.py --validate ....` Is this what you want? I thought the original purpose was to allow people to directly copy the example line, which is now not possible anymore, neither on linux (where it worked before) nor on windows (where you would use `python fast-downward.py` or maybe other syntax I don't know).

silvansievers: P.S.: the reason I had to change this is because the EXAMPLE variable is used from the driver tests, and there I needed to change all calls from `./fast-downward.py` to `sys.executable fast-downward.py`. Do you think we should also determine the script name automatically there?

misc/autodoc/autodoc.py
silvansievers: Good point. Looking at master.cfg in the infrastructure repo, it seems like the worker first compiles the planner and then runs autodoc: 
```
factory_dict["autodoc"] = util.BuildFactory([
    update_step(incremental=True),
    autodoc_step,
])
```
I'm not sure, though.

maltehelmert: Yes, that's the right file. The `update_step` is just to pull from the repository. If you look at the other build factories above, they all include `build_all_step`, which does the actual build.

driver/arguments.py
maltehelmert: I think it's not necessary to follow all such "best practices" in the tests. In general, please go with whatever you prefer in the tests. BTW, I did a little test with "grep" (as a typical Unix tool) to verify that I remembered the two things (use filename programmatically, strip directory name) correctly, and indeed grep seems to follow this:

```
$ grep -u
Usage: grep [OPTION]... PATTERN [FILE]...
Try 'grep --help' for more information.
$ cp /bin/grep foo/mygrep
$ ln -s /bin/grep myothergrep
$ foo/mygrep -u
Usage: mygrep [OPTION]... PATTERN [FILE]...
Try 'mygrep --help' for more information.
$ ./myothergrep -u
Usage: myothergrep [OPTION]... PATTERN [FILE]...
Try 'myothergrep --help' for more information.
```

PR #26, opened 2021-02-10
Issue995

salome-eriksson:


src/search/landmarks/landmark_factory_relaxation.cc
```diff
@@ -9,32 +11,20 @@ void LandmarkFactoryRelaxation::generate_landmarks(const shared_ptr<AbstractTask
...
-        approximate_reasonable_orders(task_proxy, false);
-        utils::g_log << "approx. obedient reasonable orders" << endl;
-        approximate_reasonable_orders(task_proxy, true);
-    }
     mk_acyclic_graph();
     calc_achievers(task_proxy, exploration);
 }
 
 void LandmarkFactoryRelaxation::discard_noncausal_landmarks(
     const TaskProxy &task_proxy, Exploration &exploration) {
```
salome-eriksson, line 24: I'm not sure if axioms mess up the causal landmarks computation.
I'm also not sure if we should include a guard for conjunctive landmarks. We know that they are not handled properly since lm_hm(m=2,only_causal_landmarks=true) pruned landmarks while it should only produce causal landmarks in the first place. Not having a guard is currently not a problem since we removed only_causal_landmarks as an option from the only factory that created conjunctive landmarks; but if someone comes up with a new way of computing conjunctive landmarks and wants to use this option there's no indication that it won't work correctly.

src/search/landmarks/landmark_count_heuristic.cc
```diff
@@ -54,11 +54,10 @@ LandmarkCountHeuristic::LandmarkCountHeuristic(const options::Options &opts)
 
     shared_ptr<LandmarkFactory> lm_graph_factory = opts.get<shared_ptr<LandmarkFactory>>("lm_factory");
     lgraph = lm_graph_factory->compute_lm_graph(task);
-    bool reasonable_orders = lm_graph_factory->use_reasonable_orders();
     lm_status_manager = utils::make_unique_ptr<LandmarkStatusManager>(*lgraph);
 
     if (admissible) {
```
salome-eriksson, line 59: I decided to move the information whether reasonable orders are used or not to the Landmark Graph instead of the Factories, because it avoid propagating the information through all factories (analogous to supports_conditional_effects) and is a property of the graph anyway.

src/search/landmarks/landmark_graph.h
```diff
@@ -112,7 +113,8 @@ class LandmarkGraph {
        when moving landmark graph creation there. */
     int get_num_edges() const;
```
salome-eriksson, line 115: We could also do something more sophisticated like counting how many reasonable / obedient reasonable orders we have if we are interested in this output.

src/search/landmarks/landmark_factory_h_m.cc
```diff
@@ -588,78 +589,25 @@ void LandmarkFactoryHM::initialize(const TaskProxy &task_proxy) {
...
 }
 
-void LandmarkFactoryHM::generate(const TaskProxy &task_proxy) {
-    if (only_causal_landmarks) {
-        Exploration exploration(task_proxy);
-        discard_noncausal_landmarks(task_proxy, exploration);
-    }
-    if (!disjunctive_landmarks)
-        discard_disjunctive_landmarks();
+void LandmarkFactoryHM::postprocess(const TaskProxy &task_proxy) {
```
ClemensBuechner, line 592: Why is the `only_causal_landmarks` of no interest anymore? According to the reports, this option affects the number of landmarks, doesn't it?

src/search/landmarks/landmark_factory_h_m.cc
```diff
@@ -588,78 +589,25 @@ void LandmarkFactoryHM::initialize(const TaskProxy &task_proxy) {
...
     if (!conjunctive_landmarks)
         discard_conjunctive_landmarks();
     lm_graph->set_landmark_ids();
 
-    if (no_orders)
+    if (!use_orders)
```
ClemensBuechner, line 597: This change here confirms my thoughts about `use_orders` and `reasonable_orders` being on a similar level. They both denote changes on the landmark graph after its generation. More so, this is also the case for the noncausal, disjunctive, and conjunctive landmarks options. To treat these things similarly, it would make sense to keep reasonable orders as an option I think.

src/search/landmarks/landmark_factory_h_m.h
```diff
@@ -101,7 +93,10 @@ class LandmarkFactoryHM : public LandmarkFactory {
     void print_fluentset(const VariablesProxy &variables, const FluentSet &fs);
     void print_pm_op(const VariablesProxy &variables, const PMOp &op);
```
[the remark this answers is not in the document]
ClemensBuechner: I see no reason why not.

src/search/landmarks/landmark_factory_hps_orders.cc
```diff
@@ -0,0 +1,398 @@
...
+
+#include "../utils/logging.h"
+#include "../utils/markup.h"
+
+using namespace std;
+namespace landmarks {
+LandmarkFactoryHPSOrders::LandmarkFactoryHPSOrders(const Options &opts)
+    : lm_factory(opts.get<shared_ptr<LandmarkFactory>>("lm_factory")) {
+}
+
```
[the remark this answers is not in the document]
ClemensBuechner: I think this is one more indicator for the need of a discussion. Currently, we're doing two different things which we both call "landmark factories". On the one hand, it's generating landmarks. On the other hand, it's changing a previous (collection of) landmark graph(s). 
In my opinion, this calls for a redesign on another level. Based on this, I'd rather keep the reasonable orders as an option instead of making this its own landmark factory. (Please don't take this personally, Salomé. I appreciate the work you've put into this so far!)

src/search/landmarks/landmark_factory_merged.cc
```diff
@@ -142,13 +139,9 @@ void LandmarkFactoryMerged::generate(const TaskProxy &task_proxy) {
        necessary to do so again here, so these steps are omitted. For
        reasonable orders, acyclicity of the landmark graph and the costs of
        landmarks we should also determine this.
+
+       Do we still need this comment?
```
ClemensBuechner, line 143: It is not true what the comment says, right? I mean, it is only true if all the individual landmark factories are set up accordingly. However, it is perfectly fine to do something like `lm_merged([lm_rhw(disjunctive_landmarks=true), lm_hm(disjunctive_landmarks=false)], disjunctive_landmarks=false)`. I know the example is outdated because `lm_hm` does not produce disjunctive landmarks and doesn't have the orderings anymore. What I wanted to say is that without doing the postprocessing for disjunctive landmarks in `lm_merged`, the outcome will have disjunctive landmarks even though I've specifically set the option that it shouldn't have.

src/search/landmarks/landmark_factory_merged.cc
```diff
@@ -170,14 +163,10 @@ static shared_ptr<LandmarkFactory> _parse(OptionParser &parser) {
         "Fact landmarks take precedence over disjunctive landmarks, "
         "orderings take precedence in the usual manner "
         "(gn > nat > reas > o_reas). ");
-    parser.document_note(
-        "Relevant options",
-        "Depends on landmarks");
```
ClemensBuechner, line 175: What does this even mean? ;)

src/search/landmarks/landmark_factory_relaxation.cc
ClemensBuechner: Without this comment from the pull request, I wouldn't understand the TODO comment.

src/search/landmarks/landmark_graph.h
```diff
@@ -112,7 +113,8 @@ class LandmarkGraph {
...
 
-
+    // only needed only by non-landmarkgraph-factories
```
ClemensBuechner, line 116: The word "only" appears twice in this comment.

ClemensBuechner (commented):
I think we should start a discussion at some point. In my opinion, moving reasonable orders to their own landmark factory should not be part of this issue, because it is quite similar to many of the other options that also don't become a factory.
I wrote this in one comment, but want to mention it here again: currently, we're doing two different things under the name "landmark factory": (a) generating landmarks and (b) postprocess them (e.g., add reasonable orders, remove disjunctive landmarks, etc.). I suggest to redesign this.

Most of my comments are concerned with the problem mentioned above. (At least for me it is a problem.) Concerning the code, I didn't see much need for change (if any).

src/search/landmarks/landmark_factory_h_m.cc
salome-eriksson: Yes, but only because the implementation is buggy :). For m=1 setting the option to true doesn't change the number of landmarks. While I don't know exactly what happens in `is_causal_landmark` with conjunctive landmarks I doubt it's correct and in theory m=2 should also only produce causal landmarks.

src/search/landmarks/landmark_factory_h_m.h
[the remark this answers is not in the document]
salome-eriksson: I was unsure because different parts of the code seem to do this differently, but I agree that we should declare things as constant if they are constant.

src/search/cegar/utils_landmarks.cc
```diff
@@ -23,12 +23,9 @@ static FactPair get_fact(const LandmarkNode &node) {
 shared_ptr<LandmarkGraph> get_landmark_graph(const shared_ptr<AbstractTask> &task) {
     Options hm_opts;
     hm_opts.set<int>("m", 1);
```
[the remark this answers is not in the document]
silvansievers: I'm not sure I fully understand the discussion here. This function just computes some kind of  "default" h^m landmarks, right? The previous code didn't use reasonable order and so doesn't the new code, so this is fine in my view.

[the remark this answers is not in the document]
silvansievers: Btw., I don't know if this is Github's fault, but I see all comments regarding specific lines above the line instead of below the line. If I click on the plus next to a line I want to comment on, my comment will end up below the line. Did you do the same or did the comment end up above the line for another reason? I think it makes a lot of sense to have comments below the lines they refer to due to scrolling through the code from top to bottom...

src/search/landmarks/landmark_count_heuristic.cc
silvansievers: I see Clemens' point, I think `has_reasonable_orders` feels more natural for the landmark graph than `uses`. But if you think that we will anyway move away from having this member at all by counting instead, then I'm fine with the current solution. Otherwise, I would prefer having a `has` method and ensuring that the user cannot use a factory which produces reasonable orders where this is not "safe" at another place.

src/search/landmarks/landmark_factory.h
```diff
@@ -30,36 +30,17 @@ class LandmarkFactory {
 
     std::shared_ptr<LandmarkGraph> compute_lm_graph(const std::shared_ptr<AbstractTask> &task);
 
-    bool use_disjunctive_landmarks() const {return disjunctive_landmarks;}
-    bool use_reasonable_orders() const {return reasonable_orders;}
     virtual bool supports_conditional_effects() const = 0;
 
 protected:
```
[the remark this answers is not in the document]
silvansievers: The default constructor initializes all members to their default values. I would assume that nullptr is the default of a pointer, but better double-check. If there is any member which cannot be default-constructed, the compiler will complain that it cannot create a default constructor.

src/search/landmarks/landmark_factory.h
```diff
@@ -30,36 +30,17 @@ class LandmarkFactory {
...
-    bool use_orders() const {return !no_orders;}   // only needed by HMLandmark
 
     void edge_add(LandmarkNode &from, LandmarkNode &to, EdgeType type);
 
-    void discard_disjunctive_landmarks();
-    void discard_conjunctive_landmarks();
     void discard_all_orderings();
-    void approximate_reasonable_orders(
-        const TaskProxy &task_proxy, bool obedient_orders);
     void mk_acyclic_graph();
```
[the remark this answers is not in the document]
silvansievers: I guess the topic of making landmark graphs acyclic has found its own issue, right?

src/search/landmarks/landmark_factory_h_m.cc
silvansievers: Hmm. Reading your two comments regrading this change, I thought you were going to suggest to computing orders in a separate factory, too. 

But maybe I don't understand enough anyway: this part of the code looks like orders are generated anyway and only discarded if the option is not set. So using a order-factory after (or rather "wrapping") the landmark factory wouldn't work in the same way, would it? But I guess it could be fine to just drop landmarks from a previously computed graph...

src/search/landmarks/landmark_factory_h_m.cc
```diff
@@ -1054,12 +1002,13 @@ static shared_ptr<LandmarkFactory> _parse(OptionParser &parser) {
...
         "Keyder, Richter & Helmert (ECAI 2010).");
-    parser.document_note(
-        "Relevant options",
-        "m, reasonable_orders, conjunctive_landmarks, no_orders");
     parser.add_option<int>(
         "m", "subset size (if unsure, use the default of 2)", "2");
-    _add_options_to_parser(parser);
+    parser.add_option<bool>(
+        "conjunctive_landmarks",
+        "keep conjunctive landmarks",
```
silvansievers, line 1009: "keeping" means they are always computed and discarded on demand, right? Then I would suggest to call this option `keep_conjunctive_landmarks` or even better, `discard_conjunctive_landmarks`, to match the other discarding option and the name of the method in the code.

src/search/landmarks/landmark_factory_h_m.h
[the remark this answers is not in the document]
silvansievers: I cannot see which code this discussion refers to, but it seems to be resolved anyway.

src/search/landmarks/landmark_factory_merged.cc
```diff
@@ -16,8 +16,7 @@ namespace landmarks {
 class LandmarkNode;
 
 LandmarkFactoryMerged::LandmarkFactoryMerged(const Options &opts)
```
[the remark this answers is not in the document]
silvansievers: I'm not sure what you ask. I think you have to call the constructor of the super class. Or maybe you don't, but I would always do it for readability. Calling or not calling it is independent of it being a default or a custom constructor.

Ah, I guess you might be asking if changing the constructor of this merged factory to a default one would correctly call the constructor of the super class... I don't know. But can we use a default constructor here? The factory needs to retrieve the `lm_factories` from the Options object.

src/search/landmarks/landmark_factory_merged.cc
silvansievers: I think we discussed this topic when discussing the design of having factories to post process landmark graphs, like the one for making it acyclic, right? Anyway, the TODO comment indeed seems to be wrong and should go away.

src/search/landmarks/landmark_factory_h_m.cc
silvansievers: Ok, no that I have seen how large the new factory for computing reasonable orders is, I think it makes sense that it is a separate class. Also, since we are working on making landmark graphs acyclic in a separate class, too, I would stick with this being its own factory. The question then is if use_orders should be its own factory, too. As said above, I'm probably not familiar enough with the details, but if we tend towards having "postprocessing" factories in the future, then use_orders (or better: discard orders) could indeed be one of them.

Speaking of `use_orders`: I would consider to change it "back", i.e., to `discard_orders` because this better describes what the option actually does and it goes along with `discard_conjunctive_landmarks` and `make_acyclic` because all of these remove landmarks rather then computing new ones or forbidding certain types during construction. (Unless I understood this wrong, of course.)

src/search/landmarks/landmark_factory_reasonable_orders_hps.h
```diff
@@ -0,0 +1,30 @@
+#ifndef LANDMARKS_LANDMARK_FACTORY_REASONABLE_ORDERS_HPS_H
+#define LANDMARKS_LANDMARK_FACTORY_REASONABLE_ORDERS_HPS_H
+
+#include "landmark_factory.h"
+
+namespace landmarks {
+class LandmarkFactoryReasonableOrdersHPS : public LandmarkFactory {
```
silvansievers, line 7: I assume that all of the implementation in the cc-file is mostly moved from the previous factory class, isn't it? Given that I'm lazy I would prefer to not look at the cc-file in all details.

src/search/landmarks/landmark_factory_relaxation.cc
silvansievers: I suggest adding some explanation to the TODO comment in the code.

src/search/landmarks/landmark_factory_rpg_exhaust.cc
```diff
@@ -17,7 +17,7 @@ namespace landmarks {
    method with others, don't use it by itself. */
 
 LandmarkFactoryRpgExhaust::LandmarkFactoryRpgExhaust(const Options &opts)
-    : LandmarkFactoryRelaxation(opts) {
+    : only_causal_landmarks(opts.get<bool>("only_causal_landmarks")) {
```
silvansievers, line 20: As stated earlier, I personally would always call the super constructor. But I don't know if we have any conventions for this.

src/search/landmarks/landmark_factory_h_m.cc
silvansievers: A final thought on the factory vs option thing: for the new/planned factories which drop landmarks, we could consider not making them landmark graph factories but "postprocessing factories" or similar to distinguish them from the "generating" factories.

silvansievers (commented):
I left quite a few comments, many times adding comments to Clemens' comments. I would also suggest to use the Github feature of resolving a conversation after you addressed a comment or if there isn't to do anything but you agree/don't see the need to further discuss. It makes it much easier to keep an overview of what is done and what isn't. 

Also, if you didn't so far, please try to comment **below** the line if you want to comment. Many of the current comments/discussions were really off-place when shown to me.

src/search/landmarks/landmark_factory_reasonable_orders_hps.h
salome-eriksson: Yes, the implementation is copy pasted. I don't want to exclude that I might have done some minor cosmetic changes, but I doubt it.

src/search/cegar/utils_landmarks.cc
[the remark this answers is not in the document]
silvansievers: Then I think we should update the comment.

src/search/landmarks/landmark_factory.h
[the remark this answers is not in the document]
silvansievers: Ok. Also the computation of landmark costs was moved somewhere else, wasn't it?

src/search/landmarks/landmark_factory_h_m.h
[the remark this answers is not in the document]
silvansievers: I know that I once had a discussion about this with Jendrik, but I forgot the outcome/the general opinion, so this is fine with me.

src/search/landmarks/landmark_factory_merged.cc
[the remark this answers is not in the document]
silvansievers: This is what I would do, yes. But I don't think we have any rules for this.

src/search/landmarks/landmark_count_heuristic.cc
silvansievers: Yes. If this is the purpose, then this is clearly a property of the factory and not of the graph.

src/search/landmarks/landmark_factory_h_m.cc
silvansievers: I think that the option `conjunctive_landmarks`is also not ideal. Does it mean we compute these? Keep these? I agree that either `keep/compute_conjunctive_landmarks` or `discard_conjunctive_landmarks` is long, but I would still go with the longer variants. Or abbreviate them to `keep/compute_conj_lms` as the code sometimes does.

I'm in favor of that variant where we don't have to negate the option in the code. So if we only ask `if (!keep_...)`, then discard, I think the option should be named the opposite way.

maltehelmert: > I'm in favor of that variant where we don't have to negate the option in the code. So if we only ask if (!keep_...), then
> discard, I think the option should be named the opposite way.

Trying to use positive rather than negative conditions is a good principle, but I think we should care about the user interface (= the command-line options) more than we should care about the implementation details.

Generating/keeping conjunctive landmarks is expressed more positively than discarding them, so I think that's the more obvious of the two choices. Also, keeping the "direction" of the flag in the user interface the way it always has been is preferable unless there is a strong reason to change it.

src/search/landmarks/landmark_factory_merged.cc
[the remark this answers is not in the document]
salome-eriksson: I added the call to the constructor of the super class, but clang complained about redundancy which is why I removed it again (also in the other place we discussed).

PR #31, opened 2021-02-19
[issue1012] Add workflow that runs daily tests.

jendrikseipp:


jendrikseipp:
We decided to drop the nightly translator tests.

PR #32, opened 2021-02-25
Issue1013

PatrickFerber:
First, I conditioned it on the existence of the DOWNWARD_AUTODOC_PASSWORD variable, but this condition can only be checked on the step level and not on the job level, that means, every fork will execute this workflow and set up a virtual machine that does nothing.
Now the condition is on the repository name. I think this is strictly better. Only our repository should push to the wiki. Furthermore, if we are now missing the secret for some reason, the workflow is not skipped, but fails (and tells us that we have to do something)

### DIFF
--- a/.github/workflows/autodoc.yml
+++ b/.github/workflows/autodoc.yml
@@ -6,7 +6,7 @@ on: [push]
 jobs:
   autodoc:
     name: Autodoc
-    if: github.repository == 'PatrickFerber/downward'
+    if: github.repository == 'aibasel/downward'
     timeout-minutes: 60
     runs-on: ubuntu-20.04
     env:
--- a/.github/workflows/autodoc.yml
+++ b/.github/workflows/autodoc.yml
@@ -1,0 +1,30 @@
+---
+name: Autodoc
+
+on: [push]
+
+jobs:
+  autodoc:
+    name: Autodoc
+    timeout-minutes: 60
+    runs-on: ubuntu-20.04
+    env:
+      DOWNWARD_AUTODOC_PASSWORD: ${{ secrets.DOWNWARD_AUTODOC_PASSWORD }}
+    steps:
+      - name: Clone repository
+        if: ${{ env.DOWNWARD_AUTODOC_PASSWORD != 0}}
+        uses: actions/checkout@master
+
+      - name: Install Python
+        if: ${{ env.DOWNWARD_AUTODOC_PASSWORD != 0}}
+        uses: actions/setup-python@master
+        with:
+          python-version: 3.8
+
+      - name: Run Autodoc
+        if: ${{ env.DOWNWARD_AUTODOC_PASSWORD != 0}}
+        run: |
+          cd misc/autodoc
+          python3 autodoc.py
+
+...
--- a/.github/workflows/autodoc.yml
+++ b/.github/workflows/autodoc.yml
@@ -6,23 +6,21 @@ on: [push]
 jobs:
   autodoc:
     name: Autodoc
+    if: github.repository == 'PatrickFerber/downward'
     timeout-minutes: 60
     runs-on: ubuntu-20.04
     env:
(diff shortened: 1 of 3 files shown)


silvansievers:
Could you please update the pull request against the right main branch? The current PR displays a diff due to issue1004, which is probably because Thomas force-pushed the merge commit.

.github/workflows/autodoc.yml
```diff
@@ -0,0 +1,28 @@
+---
+name: Autodoc
+
+on: [push]
+
+jobs:
+  autodoc:
+    name: Autodoc
+    if: github.repository == 'aibasel/downward'
```
silvansievers, line 9: Alternatively, we could use if: github.ref == 'refs/heads/main' because we only want to run this job after pushes to the main branch. Or actually, we should use both conditions, to cover the case that someone pushes an issue branch to the aibasel repo, for which we wouldn't want to run autodoc, and to still exclude the whole workflow from being run on forks.

PatrickFerber:
Indeed. My main branch was prior to Thomas force push, thus, something was wrong. Now I updated it.

jendrikseipp (approved):

.github/workflows/autodoc.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@master
+
+      - name: Install Python
+        uses: actions/setup-python@master
+        with:
+          python-version: 3.8
+
+      - name: Run Autodoc
```
silvansievers, line 23: `autodoc` (lower case)

.github/workflows/autodoc.yml
jendrikseipp, line 9: Should we also only run this on the "main" branch? I think this is best configured above in the "on" part, i.e.,

```
on: 
  push:
    [main]
```
or similar.

misc/autodoc/autodoc.py
```diff
@@ -162,8 +160,11 @@ def add_page(title, text):
     add_page(overview_title, overview_text)
     return changed_pages
 
+
 if __name__ == '__main__':
     args = parse_args()
+    if not args.dry_run and PASSWORD is None:
+        sys.exit(f"{ENV_VAR_PASSWORD} not set.")
```
maltehelmert, line 167: This should follow the conventions (using `logging`, then `sys.exit(1)`) used in the rest of the file.

maltehelmert: Specifically, I would go with `logging.critical`, as in the old code that read the password from the file.

.github/workflows/autodoc.yml
maltehelmert: Would this not have to be something like:
```
on:
  push:
    branches:
      - main
```
?

It is important is that we only run this on a revision that is on main. For example, if we ever push to an old release branch (for a bugfix), we don't want this to overwrite the documentation.

PatrickFerber: I think too that we need both checks (pushing an issue branch can easily happen, and pushing a release branch is intended).

I prefer the branch check to use the ``` on: ...``` syntax, because I think it is easier to read and the more frequently used approach (the on thing is described everywhere, the if: github.X rarely).

Would this resolve your comment or did I miss something?

PatrickFerber: Done. I took Maltes code, because that is also how it is written in the documentation.

misc/autodoc/autodoc.py
PatrickFerber: I dislike how some comments simply disappear in the diff-GUI, because the code section was changed...

I have incorporated your comment.

.github/workflows/autodoc.yml
silvansievers: Fine.

jendrikseipp: Thanks! Malte's code is the correct solution. I hadn't looked up the real syntax.

PR #33, opened 2021-03-01
Issue1007 - Integration of the CEGAR pattern collection generators

silvansievers:


src/search/pdbs/cegar.h
```diff
@@ -0,0 +1,170 @@
...
+          variable(variable) {
+    }
+};
+
+using FlawList = std::vector<Flaw>;
+
+/*
+  This class implements the CEGAR algorithm for computing pattern collections.
+
+  In a nutshell, it receives a concrete task plus a (sub)set of its goal in a
```
thomaskeller79, line 88: goals

thomaskeller79 (commented):
I looked at the pattern collection single cegar and cegar code and left a few comments that I'd like to discuss before proceeding with the code review. The pattern collection multiple cegar code can be reviewed independently of the discussion, this only affects the other new classes.

src/search/pdbs/cegar.h
```diff
@@ -0,0 +1,170 @@
...
+  Further parameters allow blacklisting a (sub)set of the non-goal variables
+  which are then never added to the collection, limiting PDB and collection
+  size, setting a time limit and switching between computing regular or
+  wildcard plans, where the latter are sequences of parallel operators
+  inducing the same abstract transition.
+*/
+class CEGAR {
+    const int max_pdb_size;
+    const int max_collection_size;
+    const bool wildcard_plans;
```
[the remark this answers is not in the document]
silvansievers: If anything, it makes the application of the plan a bit slower because we need to test all equivalent operators in each step. But this certainly wouldn't be a bottle neck. I would prefer to keep the option, though, because we evaluate it in the paper.

src/search/pdbs/cegar.cc
```diff
@@ -0,0 +1,589 @@
...
+        "true");
+    parser.add_option<double>(
+        "max_time",
+        "maximum time in seconds for the CEGAR algorithm (ignored for"
+        "computing initial collection)",
+        "infinity",
+        Bounds("0.0", "infinity"));
+}
+
+vector<FactPair> get_goals_in_random_order(
```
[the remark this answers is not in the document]
silvansievers: You are right. I added it there because it is used only by those generators based on CEGAR. I could move it to `utils.h`, even though the scope of that file currently is much more general. But it probably doesn't hurt to have a few more specific functions there.

src/search/pdbs/cegar.cc
```diff
@@ -0,0 +1,589 @@
...
+
+#include "../task_utils/task_properties.h"
+
+#include "../utils/countdown_timer.h"
+#include "../utils/rng.h"
+
+using namespace std;
+
+namespace pdbs {
+static unique_ptr<Projection> compute_projection(
```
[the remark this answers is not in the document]
silvansievers: I think this happened when I started splitting the computation of projection/abstract plans from CEGAR. In the end, I never reconsidered whether this function belongs more to the computation of the plan (i.e., `steepest_ascent_hill_climbing`) or to the CEGAR algorithm. I now made it a member of the CEGAR algorithm. This also removes all parameters except of the pattern, which is a good sign this function should be a member indeed.

src/search/pdbs/cegar.cc
```diff
@@ -0,0 +1,589 @@
...
+FlawList CEGAR::get_flaws() {
+    FlawList flaws;
+    State concrete_init = task_proxy.get_initial_state();
+    for (size_t collection_index = 0;
+         collection_index < projection_collection.size(); ++collection_index) {
+        if (projection_collection[collection_index] &&
+            !projection_collection[collection_index]->is_solved()) {
+            FlawList new_flaws =
+                get_flaws_for_projection(collection_index, concrete_init);
+            if (concrete_solution_index != -1) {
```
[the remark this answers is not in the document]
silvansievers: We need to pass the index to `get_flaws_for_projection` because that in turn passes it to `apply_plan`. Note that indices are part of what a flaw constitutes: a variable plus the index of the projection in which it is a flaw. So we cannot simplify the loop unfortunately. I also dislike pairs as return values. 

If anything, I could imagine dropping `new_flaws` and instead passing `flaws` to `get_flaws_for_projection` which then modifies it and returns a single boolean. We could even go a step further and also pass `flaws` as an argument to `get_flaws`, which could then return a boolean if the problem is solved. This would allow us removing the `concrete_solution_index`. Instead, we could exit the main method `compute_pattern_collection` if the returned value is true. Otherwise, assert that flaws is non-empty and proceed. 

I know that passing non-const empty parameters for modifying them isn't that beautiful either, but I think it is a more common pattern and better than lists (or pairs) of return values. What do you think?

src/search/pdbs/cegar.cc
```diff
@@ -0,0 +1,589 @@
...
+                assert(concrete_solution_index == static_cast<int>(collection_index));
+                assert(new_flaws.empty());
+                // Return empty flaws to terminate the algorithm.
+                flaws.clear();
+                return flaws;
+            }
+            for (Flaw &flaw : new_flaws) {
+                flaws.push_back(move(flaw));
```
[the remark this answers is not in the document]
silvansievers: Good catch! I even knew about this but simply forgot to apply it here.

src/search/pdbs/cegar.cc
```diff
@@ -0,0 +1,589 @@
...
+
+FlawList CEGAR::apply_plan(int collection_index, State &current) const {
+    FlawList flaws;
+    Projection &projection = *projection_collection[collection_index];
+    const vector<vector<OperatorID>> &plan = projection.get_plan();
+    if (verbosity >= utils::Verbosity::VERBOSE) {
+        utils::g_log << "executing plan for pattern "
+                     << projection.get_pattern() << ": ";
+    }
+    for (const vector<OperatorID> &equivalent_ops : plan) {
```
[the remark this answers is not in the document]
silvansievers: You are absolutely right. I am very surprised that this doesn't happen somewhere, but I double-checked, and the operators are not shuffled elsewhere (e.g., directly in the computation of the plan). I'll change this separately to evaluate the impact of this change in an isolated manner.

src/search/pdbs/cegar.cc
```diff
@@ -0,0 +1,589 @@
...
+        bool step_failed = true;
+        for (OperatorID op_id : equivalent_ops) {
+            OperatorProxy op = task_proxy.get_operators()[op_id];
+
+            /*
+              Check if the operator is applicable. If not, add its violated
+              preconditions to the list of flaws.
+            */
+            bool flaw_detected = false;
```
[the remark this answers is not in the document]
silvansievers: It is again not that simple because `operator_is_applicable` would actually and primarily collect violated preconditions as flaws. It would look something like, using the same approach as suggested above, using a single bool return value and filling flaws passed in as non-const parameter:
```
bool CEGAR::get_violated_preconditions(
    const State &current,
    const OperatorProxy &op,
    int collection_index,
    FlawList &flaws) const {
    bool violated_preconditions = false;
    for (FactProxy precondition : op.get_preconditions()) {
        int var = precondition.get_variable().get_id();

        // Ignore blacklisted variables.
        if (blacklisted_variables.count(var)) {
            continue;
        }

        if (current[precondition.get_variable()] != precondition) {
            violated_preconditions = true;
            flaws.emplace_back(collection_index, var);
        }
    }
    return violated_preconditions;
}

if (!get_violated_preconditions(current, op, collection_index, flaws)) {
    step_failed = false;
    flaws.clear();
    current = get_unregistered_successor(task, current, op);
    break;
}
```
I'm not sure this is better than before, but i'm also not really opposed to it. What do you think? Or do you have better ideas?

src/search/pdbs/cegar.cc
```diff
@@ -0,0 +1,589 @@
...
+        if (verbosity >= utils::Verbosity::NORMAL) {
+            utils::g_log << "time limit reached." << endl;
+        }
+        return true;
+    }
+    return false;
+}
+
+/*
+  TODO: this is a duplicate of State::get_unregistered_successor.
```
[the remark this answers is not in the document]
silvansievers: I think it is a very important assertion, though, which I would like to have in my regular code. Therefore I'm not sure what to do about it and would wait/ask for Malte's opinion.

src/search/pdbs/cegar.cc
```diff
@@ -0,0 +1,589 @@
...
+            added_var = true;
+        }
+    }
+
+    if (!added_var) {
+        if (verbosity >= utils::Verbosity::VERBOSE) {
+            utils::g_log << "could not add var/merge patterns due to size "
+                "limits. Blacklisting." << endl;
+        }
+        blacklisted_variables.insert(var);
```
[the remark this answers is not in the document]
silvansievers: We initially also had local blacklists, but this was problematic for the guaranteed linear runtime of the algorithm: the global blacklist guarantees that we "handle" one variable in every iteration - either by including it into the collection or blacklisting it. Note that we think of the collection as a whole here and not of individual patterns, i.e., having variables in or not in the collection. This makes sense because the collection is disjoint. 

If we excluded a variable for a pattern only, we would lose that runtime guarantee. Additionally, the global blacklist is used to forbid adding variables to the entire collection up-front when using the multiple CEGAR algorithm.

src/search/pdbs/cegar.cc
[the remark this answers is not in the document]
maltehelmert: Which one is the added assertion? Line 115 seems redundant because it is implied by line 125, so I assume line 125 is a relevant one. Is there a reason why we want to assert this for every successors rather than -- like we do for other heuristics -- once, in a larger context? If we absolutely need this assertion here and it is the only change compared to the original `get_unregistered_successor`, why do we need to copy the code rather than have a wrapper function that consists just of a call to the original `get_unregistered_successor` and the assertion?

src/search/pdbs/cegar.h
```diff
@@ -0,0 +1,170 @@
...
+    const double max_time;
+    const utils::Verbosity verbosity;
+    std::shared_ptr<utils::RandomNumberGenerator> rng;
+    const std::shared_ptr<AbstractTask> &task;
+    const TaskProxy task_proxy;
+    const std::vector<FactPair> goals;
+    std::unordered_set<int> blacklisted_variables;
```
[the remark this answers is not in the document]
silvansievers: Interesting suggestion. I would also like to measure the impact of such a change, thereby adding it to my list of things to try.

src/search/pdbs/cegar.cc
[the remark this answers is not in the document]
silvansievers: Sorry for the confusion, but it is about neither of these assertions, but about one the original method has, which I copy here:
```
State State::get_unregistered_successor(const OperatorProxy &op) const {
    assert(!op.is_axiom());
    assert(task_properties::is_applicable(op, *this));
    assert(values);
    vector<int> new_values = get_unpacked_values();

    for (EffectProxy effect : op.get_effects()) {
        if (does_fire(effect, *this)) {
            FactPair effect_fact = effect.get_fact().get_pair();
            new_values[effect_fact.var] = effect_fact.value;
        }
    }

    if (task->get_num_axioms() > 0) {
        AxiomEvaluator &axiom_evaluator = g_axiom_evaluators[TaskProxy(*task)];
        axiom_evaluator.evaluate(new_values);
    }
    return State(*task, move(new_values));
}
```
The is_applicable assertion doesn't hold in our case. I further replaced the code that does axiom evaluation by the assertion on line 125. The assertion on line 115 is also present in the original code. 

The only difference we would really need is to not assert that the operator is applicable.

src/search/pdbs/cegar.cc
```diff
@@ -0,0 +1,589 @@
...
+            task, move(new_pattern), rng, wildcard_plans, verbosity);
+
+    // Update collection size.
+    collection_size -= pdb_size1;
+    collection_size -= pdb_size2;
+    collection_size += merged->get_pdb()->get_size();
+
+    // Clean up.
+    projection_collection[index1] = move(merged);
+    projection_collection[index2] = nullptr;
```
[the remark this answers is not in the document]
silvansievers: That's right, but I don't think gabs are a problem here. It makes iterating over the collection a bit more tedious, but we could actually write an iterator for that. I don't like using shared_ptr for something which is not shared only to avoid gaps. I'd rather try to hide the gaps behind an iterator.

src/search/pdbs/cegar.cc
```diff
@@ -0,0 +1,589 @@
...
+         collection_index < projection_collection.size(); ++collection_index) {
+        if (projection_collection[collection_index] &&
+            !projection_collection[collection_index]->is_solved()) {
+            FlawList new_flaws =
+                get_flaws_for_projection(collection_index, concrete_init);
+            if (concrete_solution_index != -1) {
+                assert(concrete_solution_index == static_cast<int>(collection_index));
+                assert(new_flaws.empty());
+                // Return empty flaws to terminate the algorithm.
+                flaws.clear();
```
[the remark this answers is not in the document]
silvansievers: I'm leaving this comment open until we agreed how to fix the above comment. If we go with my suggestion, then we would return a bool here.

src/search/pdbs/cegar.cc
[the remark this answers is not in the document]
maltehelmert: I see. Yes, I agree this is a valuable assertion. This perhaps needs to be considered in a wider context then. If it doesn't already, I think it would be good for the version in `cegar.cc` to explain how it differs from the original function and why this is desirable. (We can discuss this later.)

src/search/pdbs/cegar.cc
```diff
@@ -0,0 +1,589 @@
...
+                }
+            }
+        }
+    }
+    return flaws;
+}
+
+FlawList CEGAR::get_flaws() {
+    FlawList flaws;
+    State concrete_init = task_proxy.get_initial_state();
```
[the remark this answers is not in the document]
silvansievers: Good idea, even though I don't think it matters performance-wise. I just wonder how the copy actually works: if I copy an unpacked state, will the copy be unpacked, too? I couldn't find a copy constructor in `task_proxy.h`.

silvansievers:
Thanks a lot for the comments so far. I addressed most of them and replied to some others where I think we need to discuss a little bit more. Plus some that I need to experimentally evaluate individually.

I marked many conversations as resolved as to keep track of what I did already and what I did not. Feel free to still reply to those (I hope this is possible; if not, just unmark as resolved). Feel free to mark as resolved those comments where I said I don't want to do the change if you are fine with this.

src/search/pdbs/cegar.cc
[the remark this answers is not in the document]
silvansievers: I went ahead and implemented the suggestion. Please let me know if you don't like it.

src/search/pdbs/cegar.cc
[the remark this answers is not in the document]
thomaskeller79: As the flaws are empty iff violated_preconditions is false, I'd suggest to change this to:

```
FlawList CEGAR::get_violated_preconditions(
    const State &current,
    const OperatorProxy &op,
    int collection_index) {
    FlawList flaws;
    ...
    return flaws;
}

flaws = get_violated_preconditions(current, op, collection_index);
if (!flaws.empty()) {
    ...
}
```

src/search/pdbs/cegar.cc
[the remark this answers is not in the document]
thomaskeller79: I assume you have reasons not to change this to something along the lines of
``` 
State State::get_unregistered_successor(const OperatorProxy &op) const {
    assert(task_properties::is_applicable(op, *this));
    return get_unregistered_successor_ignoring_preconditions(op);
}

State State::get_unregistered_successor_ignoring_preconditions(const OperatorProxy &op) const {
    // original function without the assertion
}
```

src/search/pdbs/cegar.cc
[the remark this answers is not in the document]
thomaskeller79: > Good idea, even though I don't think it matters performance-wise.

You're probably right that it doesn't matter, it's just a few pointers and ints that are copied. 

>  I just wonder how the copy actually works: if I copy an unpacked state, will the copy be unpacked, too? I couldn't find a copy constructor in `task_proxy.h`.

I think the copy is unpacked as the implicit copy constructor of the State class calls all copy constructors of its member variables, and the copy consrtuctor of a shared_ptr should create a copy of the shared_ptr. I don't know for sure though.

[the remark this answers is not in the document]
silvansievers: This is my understanding, too.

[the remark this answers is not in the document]
maltehelmert: Yes, if there are no user-defined copy constructors, copies are performed elementwise, which in case of a shared pointer means just copying the pointer (and incrementing its reference count), so copying a state is a fairly cheap operation.

src/search/pdbs/pattern_collection_generator_multiple_cegar.cc
```diff
@@ -0,0 +1,300 @@
...
+            }
+        }
+        if (!is_goal_var) {
+            non_goal_variables.push_back(var_id);
+        }
+    }
+
+    if (verbosity >= utils::Verbosity::DEBUG) {
+        utils::g_log << "goal variables: ";
+        for (auto goal : goals) {
```
thomaskeller79, line 66: I try to avoid using auto unless the class type is clear and the code gets much more readable with auto (e.g., in the case of iterators). In a case like this one, I'd go with "FactProxy".

src/search/pdbs/pattern_collection_generator_multiple_cegar.cc
```diff
@@ -0,0 +1,300 @@
...
+    // Collect all unique patterns and their PDBs.
+    utils::HashSet<Pattern> generated_patterns;
+    shared_ptr<PDBCollection> generated_pdbs = make_shared<PDBCollection>();
+
+    bool blacklisting = false;
+    double stagnation_start_time = -1;
+    int num_iterations = 1;
+    int goal_index = 0;
+    int remaining_collection_size = max_collection_size;
+    const utils::Verbosity cegar_verbosity(utils::Verbosity::SILENT);
```
thomaskeller79, line 82: Is there a reason why a different verbosity is passed to CEGAR?

src/search/pdbs/pattern_collection_generator_multiple_cegar.cc
```diff
@@ -0,0 +1,300 @@
...
+            blacklisting = true;
+            // Also reset stagnation timer in case it was already set.
+            stagnation_start_time = -1;
+            if (verbosity >= utils::Verbosity::NORMAL) {
+                utils::g_log << "given percentage of total time limit "
+                             << "exhausted; enabling blacklisting." << endl;
+            }
+        }
+
+        unordered_set<int> blacklisted_variables;
```
thomaskeller79, line 100: replace with `vector<bool> blacklisted_variables(num_vars, false)` (or bitset)

(I am aware you'd have to change this in the CEGAR implementation as well, which should easily be possible. However, I understand if you don't think it's worth the trouble.)

src/search/pdbs/pattern_collection_generator_multiple_cegar.cc
```diff
@@ -0,0 +1,300 @@
...
+        if (blacklisting && !non_goal_variables.empty()) {
+            /*
+              Randomize the number of non-goal variables for blacklisting.
+              We want to choose at least 1 non-goal variable and up to the
+              entire set of non-goal variables.
+            */
+            int blacklist_size = (*rng)(non_goal_variables.size());
+            ++blacklist_size; // [1, |non-goal variables|]
+            rng->shuffle(non_goal_variables);
+            for (int i = 0; i < blacklist_size; ++i) {
```
thomaskeller79, line 110: If blacklisted_variables won't be replaced with a bool vector, this can be done without the loop:
```
blacklisted_variables.insert(
    non_goal_variables.begin(), non_goal_variables.begin() + blacklist_size);
```

src/search/pdbs/pattern_collection_generator_multiple_cegar.cc
```diff
@@ -0,0 +1,300 @@
...
+                             << " non-goal variables: ";
+                for (int var : blacklisted_variables) {
+                    utils::g_log << var << ", ";
+                }
+                utils::g_log << endl;
+            }
+        }
+
+        int remaining_pdb_size_for_cegar = min(remaining_collection_size, max_pdb_size);
+        double remaining_time_for_cegar = min(timer.get_remaining_time().operator double(), cegar_max_time);
```
thomaskeller79, line 126: Fancy, I've never seen something like this before ;)

src/search/pdbs/pattern_collection_generator_multiple_cegar.cc
```diff
@@ -0,0 +1,300 @@
...
+                break;
+            }
+        } else {
+            // Pattern is not new. Set stagnation start time if not already set.
+            if (stagnation_start_time == -1) {
+                stagnation_start_time = timer.get_elapsed_time();
+            }
+        }
+
+        if (timer.is_expired()) {
```
thomaskeller79, line 181: I'd move the code from here to line 213 to a function `bool more_iterations(...)` and replace the `while(true)` in line 88 with `while(more_iterations(...))`

src/search/pdbs/pattern_collection_generator_multiple_cegar.h
```diff
@@ -0,0 +1,41 @@
...
+    const int max_pdb_size;
+    const int max_collection_size;
+    const bool wildcard_plans;
+    const double cegar_max_time;
+
+    // Options for this generator.
+    const utils::Verbosity verbosity;
+    std::shared_ptr<utils::RandomNumberGenerator> rng;
+    // We store the random seed for creating different RNG objects for CEGAR.
+    const int random_seed;
```
thomaskeller79, line 27: You could instead store a second RandomNumberGenerator here (e.g.,  `std::shared_ptr<utils::RandomNumberGenerator> rng_for_cegar`) that you reuse in all calls to CEGAR.

src/search/pdbs/pattern_database.cc
```diff
@@ -280,16 +280,28 @@ bool PatternDatabase::is_goal_state(
...
-size_t PatternDatabase::hash_index(const vector<int> &state) const {
+size_t PatternDatabase::hash_index_of_concrete_state(const vector<int> &state) const {
     size_t index = 0;
     for (size_t i = 0; i < pattern.size(); ++i) {
         index += hash_multipliers[i] * state[pattern[i]];
     }
     return index;
 }
 
+size_t PatternDatabase::hash_index_of_projected_state(const State &projected_state) const {
```
thomaskeller79, line 291: shouldn't this (for consistency with `hash_index_of_concrete_state`) also just take a `const vector<int>` as parameter?

src/search/abstract_task.cc
```diff
@@ -14,6 +14,21 @@ ostream &operator<<(ostream &os, const FactPair &fact_pair) {
     return os;
 }
```
thomaskeller79, line 16: it might be good to add a comment that this shouldn't be called repeatedly in efficiency critical parts of the code since it is a static property that can just be computed once

src/search/pdbs/utils.cc
```diff
@@ -5,6 +5,7 @@
 #include "pattern_information.h"
 
 #include "../utils/logging.h"
+#include "../utils/rng.h"
 
 #include "../task_proxy.h"
```
thomaskeller79, line 11: For some reason, I cannot comment in the right line here, this comment belongs to line 14.

I'd use an overflow safe multiplication here, e.g. by adding an `is_product_within_limit` check (as defined in `src/search/utlis/math.h`).

thomaskeller79 (commented):
I looked through the remaining files, except those that we discussed that might still be changed quite a bit (steepest ascent hc and the tasks)

src/search/pdbs/pattern_collection_generator_multiple_cegar.cc
maltehelmert: It's probably more idiomatic to write `static_cast<double>(timer.get_remaining_time())`

silvansievers: I agree. I actually don't recall writing this myself; I assume my IDE did when I was refactoring something.

src/search/pdbs/cegar.cc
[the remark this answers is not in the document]
silvansievers: I don't know. Malte, what wider context did you mean?

src/search/pdbs/cegar.cc
```diff
@@ -0,0 +1,589 @@
...
+    if (verbosity >= utils::Verbosity::VERBOSE) {
+        utils::g_log << endl;
+    }
+
+    shared_ptr<PatternCollection> patterns = make_shared<PatternCollection>();
+    shared_ptr<PDBCollection> pdbs = make_shared<PDBCollection>();
+    if (concrete_solution_index != -1) {
+        const shared_ptr<PatternDatabase> &pdb =
+            projection_collection[concrete_solution_index]->get_pdb();
+        pdbs->push_back(pdb);
```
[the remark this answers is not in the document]
silvansievers: Actually, this is not possible because `pdb` is a const reference and cannot be moved.

src/search/pdbs/cegar.cc
```diff
@@ -0,0 +1,589 @@
...
+        patterns->push_back(pdb->get_pattern());
+    } else {
+        for (const unique_ptr<Projection> &projection : projection_collection) {
+            if (projection) {
+                const shared_ptr<PatternDatabase> &pdb = projection->get_pdb();
+                pdbs->push_back(pdb);
```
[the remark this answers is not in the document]
silvansievers: Same as above.

src/search/pdbs/cegar.cc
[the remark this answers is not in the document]
maltehelmert: If we ignore preconditions, it is not really a successor, so it is a somewhat weird concept to be promoted to the (very prominent) public interface of the state class. If we don't have such a function, we have potentially useful code invariants that get lost if we add it (only reachable states can be generated).

By wider context I meant other solutions to the same problem, or ways to avoid needing this in the first place. To judge this I would need to know more about where this is used and why.

An example of another solution to the same problem could be a task adapter that ignores preconditions. But I also wonder why we want this function.

src/search/pdbs/cegar.cc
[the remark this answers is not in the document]
silvansievers: Good point. I think I simplified the code further by exiting directly from the loop if a plan step fails. Please let me know if you think this should be amended even further.

src/search/pdbs/cegar.cc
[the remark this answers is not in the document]
silvansievers: Ok. I don't think we need anything of this.

Instead, I'd suggest to not use `State` at all here but unpacked values, i.e., `vector<int>`. The function in question would just be called `apply_operator`, not check applicability and update the vector in-place, saving the copy. 

At the end, we would only once need to construct a `State` object from the vector because we want to check if the resulting state is a goal state.

src/search/pdbs/pattern_collection_generator_multiple_cegar.cc
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    TaskProxy task_proxy(*task);
+    utils::CountdownTimer timer(total_max_time);
+
+    // Store the set of goals in random order.
+    vector<FactPair> goals = get_goals_in_random_order(task_proxy, rng);
+
+    // Store the non-goal variables for potential blacklisting.
+    int num_vars = task_proxy.get_variables().size();
```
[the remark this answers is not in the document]
silvansievers: Writing it like this is explicit enough; we usually don't use casts in such cases. I think we only use casts when doing comparisons and the compiler would warn about a comparison of unsigned and signed data types.

src/search/pdbs/pattern_collection_generator_multiple_cegar.cc
```diff
@@ -0,0 +1,300 @@
...
+        utils::g_log << "Generating patterns using the multiple CEGAR algorithm" << endl;
+    }
+
+    TaskProxy task_proxy(*task);
+    utils::CountdownTimer timer(total_max_time);
+
+    // Store the set of goals in random order.
+    vector<FactPair> goals = get_goals_in_random_order(task_proxy, rng);
+
+    // Store the non-goal variables for potential blacklisting.
```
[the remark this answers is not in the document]
silvansievers: But the comment refers to the whole block below it. How would you write such comments?

src/search/pdbs/pattern_collection_generator_multiple_cegar.cc
silvansievers: It is `FactPair` but yes, you are absolutely right. This is a remainder of what Alex wrote.

src/search/pdbs/pattern_collection_generator_multiple_cegar.cc
silvansievers: Yes. Even with normal verbosity, there would be way too much output. Also, I don't think that the multiple CEGAR algorithm should care about how the pattern it gets was generated. It produces output itself whenever it think this is useful, such as when turning blacklisting on. Maybe it could produce even more output on the way, as to have some visual kind of progress. But dumping something every iteration is just too verbose.

src/search/pdbs/pattern_collection_generator_multiple_cegar.cc
silvansievers: As commented there, I like the idea and will evaluate changing this once I'm done with the other refactoring from your review.

src/search/pdbs/pattern_collection_generator_multiple_cegar.cc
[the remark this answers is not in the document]
thomaskeller79: I wanted to suggest to move lines 47 - 62 to a separate method `get_non_goal_variables`. Sorry for the confusing comment.

src/search/pdbs/pattern_collection_generator_multiple_cegar.cc
thomaskeller79: I would turn ```bool blacklisting``` and ```double stagnation_start_time``` into member variables of the class, but I'm also available on zoom if you wish to discuss this in person.

src/search/pdbs/utils.cc
silvansievers: It is a bit redundant to add this because the assumption is that you only want to compute the PDB size of a pattern if you also have a PDB for it. The PDB itself, however, can only be constructed if there is no overflow when computing the hash multipliers. I added the suggested safe-guard anyway.

src/search/abstract_task.cc
silvansievers: Does this refer to `has_conditional_effects`? I could add a comment, but this has nothing to do with this issue.

src/search/pdbs/pattern_database.cc
silvansievers: I'll leave that open for now. If we actually get rid of `ProjectedTask`, then this needs to be changed anyway.

src/search/abstract_task.cc
thomaskeller79: Yes it does. The diff view shows me that this function has been added.

silvansievers: You are right, my bad. I thought I got this from somewhere else and only exposed it, but I can't find it anymore. I updated the comment in the header file.

thomaskeller79: After reading my previous comment I realize that it sounds very harsh, but it wasn't intended that way. My "yes it does" was supposed to answer your question, I didn't mean to insist it has something to do with the issue!

src/search/tasks/projected_task.cc
```diff
@@ -0,0 +1,156 @@
...
+vector<int> ProjectedTask::get_initial_state_values() const {
+    vector<int> values = parent->get_initial_state_values();
+    convert_state_values_from_parent(values);
+    return values;
+}
+
+void ProjectedTask::convert_state_values_from_parent(
+    vector<int> &values) const {
+    vector<int> converted;
+    converted.reserve(values.size());
```
jendrikseipp, line 150: Shouldn't this be `variables.size()`?

src/search/tasks/projected_task.cc
```diff
@@ -0,0 +1,156 @@
...
+    return goals[index];
+}
+
+vector<int> ProjectedTask::get_initial_state_values() const {
+    vector<int> values = parent->get_initial_state_values();
+    convert_state_values_from_parent(values);
+    return values;
+}
+
+void ProjectedTask::convert_state_values_from_parent(
```
jendrikseipp, line 147: If this method is executed for each evaluated state and each PDB, it will be a major bottleneck, I think. At least this is the case for the DomainAbstractedTask class used by the landmark decomposition in the Cartesian CEGAR code.

src/search/pdbs/pattern_database.h
```diff
@@ -82,6 +92,10 @@ class PatternDatabase {
     */
     std::vector<int> distances;
 
+    std::vector<int> generating_op_ids;
+    std::vector<int> number_of_zero_cost_ops;
+    std::vector<std::vector<OperatorID>> wildcard_plan;
```
jendrikseipp, line 97: Wouldn't it make sense to decouple PDB creation from PDB lookup first (as proposed in another issue)? Otherwise, all other pattern generators will need to store these extra members for each PDB even though they don't need them.

src/search/pdbs/pattern_database.cc
```diff
@@ -239,6 +248,21 @@ void PatternDatabase::create_pdb(
...
+          an operator leading from that state to another state on a
+          strongly optimal plan of the PDB. We store the first operator
+          encountered during Dijstra and only update it if the goal distance
+          of the state was updated or, in the presence of zero-cost operators,
+          the state could be reached with the same cost but with fewer
+          zero-cost operators. For the latter, we need to count the number of
+          zero-cost operators leading to each state.
+         */
+        generating_op_ids.resize(num_states);
+        number_of_zero_cost_ops.resize(num_states);
```
jendrikseipp, line 263: I'd store the info in a single vector of two-int structs: better cache-locality, less memory, maybe simpler code.

src/search/pdbs/pattern_collection_generator_multiple_cegar.cc
```diff
@@ -0,0 +1,336 @@
...
+        "20.0",
+        Bounds("1.0", "infinity"));
+    parser.add_option<double>(
+        "blacklist_trigger_percentage",
+        "percentage of total_max_time after which the multiple CEGAR "
+        "algorithm enables blacklisting for diversification",
+        "0.75",
+        Bounds("0.0", "1.0"));
+    parser.add_option<bool>(
+        "blacklist_on_stagnation",
```
jendrikseipp, line 317: `enable_blacklist_on_stagnation`?

src/search/pdbs/utils.h
```diff
@@ -18,6 +22,10 @@ extern int compute_pdb_size(const TaskProxy &task_proxy, const Pattern &pattern)
 extern int compute_total_pdb_size(
     const TaskProxy &task_proxy, const PatternCollection &pattern_collection);
 
+extern std::vector<FactPair> get_goals_in_random_order(
+    const TaskProxy &task_proxy, const std::shared_ptr<utils::RandomNumberGenerator> &rng);
```
jendrikseipp, line 26: Use plain reference for RNG instead?

src/search/pdbs/utils.cc
```diff
@@ -28,6 +38,36 @@ int compute_total_pdb_size(
     return size;
 }
 
+vector<FactPair> get_goals_in_random_order(
+    const TaskProxy &task_proxy, const shared_ptr<utils::RandomNumberGenerator> &rng) {
+    vector<FactPair> goals;
+    for (FactProxy goal : task_proxy.get_goals()) {
+        goals.push_back(goal.get_pair());
+    }
```
jendrikseipp, line 46: This could use `get_fact_pairs(task_proxy.get_goals())` from task_properties.h, which also makes sure the vector has the minimum reserved size.

src/search/pdbs/utils.cc
```diff
@@ -28,6 +38,36 @@ int compute_total_pdb_size(
...
+                is_goal_var = true;
+                break;
+            }
+        }
+        if (!is_goal_var) {
+            non_goal_variables.push_back(var_id);
+        }
+    }
+    return non_goal_variables;
+}
```
jendrikseipp, line 69: I think it's simplest to create a `vector<bool>(num_vars, false)`, set the goal bits, then extract the info into a `vector<int>`.

src/search/pdbs/utils.cc
```diff
@@ -44,16 +84,21 @@ void dump_pattern_collection_generation_statistics(
...
                  << endl;
-    utils::g_log << identifier << " total PDB size: "
-                 << compute_total_pdb_size(pci.get_task_proxy(), pattern_collection)
-                 << endl;
+    utils::g_log << identifier << " total PDB size: ";
+    if (collection_size != -1) {
+        utils::g_log << collection_size << endl;
+    } else {
+        utils::g_log << compute_total_pdb_size(
+            pci.get_task_proxy(), pattern_collection) << endl;
```
jendrikseipp, line 100: I'd set `collection_size` to the correct value and then just output that variable.

jendrikseipp (changes requested):
I focused on the glue code, not the implementation details.

src/search/abstract_task.h
```diff
@@ -103,7 +102,7 @@ class AbstractTask : public subscriber::SubscriberService<AbstractTask> {
       it should create the new vector in a local variable and then swap it with
       the parameter.
     */
-    virtual void convert_state_values(
+    virtual void convert_ancestor_state_values(
```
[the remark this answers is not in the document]
maltehelmert: I also stumbled over this name at some point, but don't remember much about the details.

src/search/pdbs/pattern_collection_generator_single_cegar.cc
```diff
@@ -0,0 +1,70 @@
...
+#include "../utils/rng_options.h"
+
+using namespace std;
+
+namespace pdbs {
+PatternCollectionGeneratorSingleCegar::PatternCollectionGeneratorSingleCegar(
+    const options::Options &opts)
+    : max_pdb_size(opts.get<int>("max_pdb_size")),
+      max_collection_size(opts.get<int>("max_collection_size")),
+      wildcard_plans(opts.get<bool>("wildcard_plans")),
```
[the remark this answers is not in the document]
maltehelmert: Agreed.

src/search/pdbs/pattern_database.cc
```diff
@@ -8,6 +8,7 @@
 #include "../utils/logging.h"
 #include "../utils/math.h"
 #include "../utils/timer.h"
+#include "../utils/rng.h"
```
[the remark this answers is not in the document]
maltehelmert: Or-derrrrrr! Or-derrrrrrrrrrrr!

src/search/pdbs/pattern_database.cc
```diff
@@ -70,7 +73,10 @@ PatternDatabase::PatternDatabase(
     const TaskProxy &task_proxy,
     const Pattern &pattern,
     bool dump,
-    const vector<int> &operator_costs)
+    const vector<int> &operator_costs,
+    bool compute_plan,
+    const shared_ptr<utils::RandomNumberGenerator> &rng,
```
maltehelmert, line 78: Is the rng still needed?

maltehelmert: (Applies to many places.)

src/search/abstract_task.h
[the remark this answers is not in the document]
jendrikseipp: Ah, I forgot to remove this comment. At the bottom of the pull request I found the corresponding name changes, so I think this name change is fine.

src/search/pdbs/pattern_database.cc
```diff
@@ -239,6 +248,21 @@ void PatternDatabase::create_pdb(
...
+        /*
+          If computing a plan during Dijkstra, we store, for each state,
+          an operator leading from that state to another state on a
+          strongly optimal plan of the PDB. We store the first operator
+          encountered during Dijstra and only update it if the goal distance
+          of the state was updated or, in the presence of zero-cost operators,
+          the state could be reached with the same cost but with fewer
+          zero-cost operators. For the latter, we need to count the number of
+          zero-cost operators leading to each state.
+         */
```
maltehelmert, line 261: Is this comment still correct? The "strongly optimal"/"fewer zero-cost operator" parts are no longer strictly needed now that we don't break ties uniformly any more. Are they still there?

src/search/pdbs/pattern_database.cc
```diff
@@ -258,8 +282,75 @@ void PatternDatabase::create_pdb(
             if (alternative_cost < distances[predecessor]) {
                 distances[predecessor] = alternative_cost;
                 pq.push(alternative_cost, predecessor);
+                if (compute_plan) {
+                    generating_op_ids[predecessor] = op_id;
+                    if (op.get_cost() == 0) {
+                        number_of_zero_cost_ops[predecessor] = number_of_zero_cost_ops[state_index] + 1;
+                    } else {
+                        number_of_zero_cost_ops[predecessor] = 0;
```
maltehelmert, line 290: Looks like they are. We might consider not doing that. This would make computing the PDB cheaper, but the found abstract plans would no longer be guaranteed to be strongly optimal, so it's a tradeoff. It's also worth considering disabling this computation in domains without zero-cost operators, since most of them don't have any.

src/search/pdbs/pattern_database.cc
```diff
@@ -258,8 +282,75 @@ void PatternDatabase::create_pdb(
...
+                if (num_zero_cost_ops_on_current_path < pred_num_zero_cost_ops) {
+                    generating_op_ids[predecessor] = op_id;
+                    number_of_zero_cost_ops[predecessor] = num_zero_cost_ops_on_current_path;
+                }
+            }
+        }
+    }
+
+    // Compute abstract plan
+    if (compute_plan) {
```
maltehelmert, line 305: This is a long function that does many things. There is a general guideline that a comment like the above "Compute abstract plan" in front of a larger block of code usually means that this block wants to be its own function. But this is not always easy, and I don't want to keep things up with little polishing suggestions like this. Just food for thought.

src/search/pdbs/pattern_database.h
```diff
@@ -8,6 +8,10 @@
 #include <utility>
 #include <vector>
 
+namespace utils {
+class RandomNumberGenerator;
+}
+
```
maltehelmert, line 14: Needed?

src/search/pdbs/pattern_database.h
maltehelmert: More generally, having data that is only relevant some of the time is a bit of a code smell. Not something we need to address right now, but something to keep in mind for future refactoring.

src/search/pdbs/utils.cc
maltehelmert: Yes, this would be much better than the current O(n^2) algorithm (which almost always becomes a problem eventually). Alternatively, something like set_difference can be used because we should be guaranteed that the variables occur in sorted order. (And even if that were not the case, sorting them and then using set_difference would be much more efficient.)

src/search/pdbs/utils.cc
maltehelmert: Is there a reason for the two cases in the first place? If the second one cannot always be used, then the reason is not obvious enough. If we're worried about the time it takes to output statistics, then we probably output too many statistics.

src/search/abstract_task.h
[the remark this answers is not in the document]
silvansievers: I just thought that the name was missing the direction of conversion. I'll mark this as resolved.

src/search/pdbs/pattern_collection_generator_single_cegar.cc
```diff
@@ -0,0 +1,70 @@
...
+    utils::add_rng_options(parser);
+
+    Options opts = parser.parse();
+    if (parser.dry_run())
+        return nullptr;
+
+    return make_shared<PatternCollectionGeneratorSingleCegar>(opts);
+}
+
+static Plugin<PatternCollectionGenerator> _plugin("single_cegar", _parse);
```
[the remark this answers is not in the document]
silvansievers: I would prefer to argue that the existing `cegar()` plugin could also be renamed to be more specific :) It does compute an additive Cartesian heuristic, doesn't it?.

Furthermore, the whole directory `cegar` is no longer the unique CEGAR in Fast Downward. I actually wondered if it wouldn't make more sense to name the directory `cartesian_abstractions` or alike because I already thought that the presence of `cegar.h/cc` in `pdbs` can now be confusing since there is also a directory `cegar`.

src/search/pdbs/utils.cc
silvansievers: I just thought that it is unnecessary to re-compute the size of all PDBs of a collection if that value is already known by the generating method. This is why I added an optional size parameter which is printed if given and computed (as before) otherwise. I know it is not really nice and we can change this back if you prefer. (To clarify: this is not about worrying about too much output, but about unnecessary computation time.)

src/search/pdbs/pattern_database.cc
silvansievers: Yes, when computing a regular plan, we randomly choose one operator from the set of equivalent ones. This could be done in a post processing step, though, but I thought it would be nicer to have the plan computation "done" when it's done.

src/search/pdbs/pattern_database.cc
silvansievers: We missed that detail in the discussion with Thomas, too: if we don't take care to consider an operator as the "best" only if it has minimal applications of zero-cost operators before, we are back to the first version where we tested this and where we ran into cycles of zero-cost operators when attempting to extract the plan. So I think this is indeed needed. If not, we can discuss in the sprint meeting how we can avoid doing this.

src/search/pdbs/pattern_database.cc
silvansievers: You are absolutely right. I added this within `create_pdb` to mimic the other steps, which are also not separate function, although this is a very stupid argument admittedly :) In the patch where I started separating the computation of PDBs from PatternDatabase, all of the blocks in this method here (`create_pdb`) have become separate steps. But this requires passing around a lot things. I'll leave this as it is for the moment because we might need to address the separation of the PDB computation anyway if Jendrik is right about a possible performance issue due to the added members in this class.

src/search/pdbs/pattern_database.h
silvansievers: We discussed yesterday that ideally, one would do the other issue first, but at the same time, we wanted to not keep too many issues open waiting for other issues to be integrated first. The plan is to directly continue working on the other issue.

jendrikseipp: Sounds good.

src/search/pdbs/pattern_collection_generator_single_cegar.cc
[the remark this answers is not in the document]
jendrikseipp: I'd only change existing user-facing options if it's really necessary. I agree that the "cegar" directory name is not ideal, but I think changing the name has more cons than pros. Renaming it will make merges more difficult and I think the two occurences of "cegar" are separated enough by the fact that one resides in the "pdbs" directory.

src/search/pdbs/pattern_database.cc
maltehelmert: The current version that doesn't update the parent pointers cannot run in cycles. The cycles can only occur if the parent of a node can be updated after it has been expanded.

src/search/pdbs/utils.cc
maltehelmert: I thought that was the rationale, but I'm not sure it's a good idea. I would view this as essentially debug output, and I don't think it would ever be time-critical except in cases where we produce excessive amounts of output. The drawbacks are a more convoluted interface for the function and implementation, the fact that it's fast sometimes but not always, and the fact that the output could be wrong if a wrong number is passed in. None of them critical, but I'd prefer to keep it simple, or alternatively track the total size for all pattern collections in a member variable.

thomaskeller79 (commented):
I just looked at the CEGAR documentation and left a few comments.

src/search/pdbs/cegar.cc
```diff
@@ -0,0 +1,607 @@
...
+        "Implementation Notes",
+        "The following describes differences of the implementation to "
+        "the original implementation used and described in the paper.\n\n"
+        "Conceptually, there is one larger difference which concern the "
+        "computation of (regular or wildcard) plans for PDBs. The original "
+        "implementation used an enforced hill-climbing search with the PDB "
+        "as the perfect heuristic. To handle domains with zero-cost operators, "
+        "it used a breadth-first search in each iteration to find the best "
+        "improving successor. This ensured to compute strongly optimal plans, "
+        "i.e., optimal plans with a minimum number of zero-cost operators. "
```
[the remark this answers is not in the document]
silvansievers: I think minimum is fine.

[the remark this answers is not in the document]
silvansievers: See also: https://english.stackexchange.com/questions/277236/what-is-the-difference-between-minimum-and-minimal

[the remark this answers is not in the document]
thomaskeller79: Interesting, I didn't know this difference. Thanks for the link!

src/search/pdbs/cegar.cc
```diff
@@ -532,28 +532,28 @@ void add_implementation_notes_to_parser(options::OptionParser &parser) {
...
+        "Conceptually, there is one larger difference which concerns the "
         "computation of (regular or wildcard) plans for PDBs. The original "
-        "implementation used an enforced hill-climbing search with the PDB "
-        "as the perfect heuristic. To handle domains with zero-cost operators, "
-        "it used a breadth-first search in each iteration to find the best "
-        "improving successor. This ensured to compute strongly optimal plans, "
-        "i.e., optimal plans with a minimum number of zero-cost operators. "
-        "The original implementation also ensured to use uniform random tie-"
-        "breaking when choosing the best successor.\n\n"
+        "implementation used an enforced hill-climbing (EHC) search with the "
```
thomaskeller79, line 537: the word "PDB" got lost here

src/search/pdbs/cegar.cc
```diff
@@ -532,28 +532,28 @@ void add_implementation_notes_to_parser(options::OptionParser &parser) {
...
+        "as the perfect heuristic, which ensured finding strongly optimal PDB "
+        "plans, i.e., optimal plans with a minimum number of zero-cost "
+        "operators, in domains with zero-cost operators. The original "
+        "implementation also slightly modified EHC to search for an best-"
```
thomaskeller79, line 541: *a* best-improving

src/search/pdbs/cegar.cc
```diff
@@ -532,28 +532,28 @@ void add_implementation_notes_to_parser(options::OptionParser &parser) {
...
         "computation of the PDB itself. A modification to Dijkstra's algorithm "
         "for computing the PDB values stores, for each state, the operator "
         "leading to that state (in a regression search). This generating "
-        "operator is updated only if the algorithm found a better value for the "
-        "state. After Dijkstra's finishes, the plan computation starts at the "
+        "operator is updated only if the algorithm found a cheaper path to "
+        "the state. After Dijkstra finishes, the plan computation starts at the "
         "initial state and iteratively follows the generating operator, computes "
         "all operators of the same cost inducing the same transition, until "
         "reaching a goal. This constitutes a wildcard plan. It it turned into a "
```
thomaskeller79, line 552: It it -> It is

thomaskeller79 (approved):
Just 3 typos

src/search/pdbs/cegar.cc
silvansievers: Actually, I forgot to move it (it was still there, but at the wrong position).

PR #38, opened 2021-04-28
[issue1017] Allow running Ubuntu workflow for repos with arbitrary names.

jendrikseipp:


jendrikseipp:
My patch doesn't work when LP solvers are present. I'm working on a fix now.

jendrikseipp:
Now it should work with and without the secrets.

.github/workflows/ubuntu.yml
```diff
@@ -29,10 +29,11 @@ jobs:
...
       CPLEX_URL: ${{ secrets.CPLEX129_LINUX_URL }}
       SOPLEX_URL: ${{ secrets.SOPLEX311_URL }}
-      DOWNWARD_CPLEX_ROOT: /home/runner/work/downward/lib/ibm/ILOG/CPLEX_Studio129/cplex
-      DOWNWARD_SOPLEX_ROOT: /home/runner/work/downward/lib/soplex-3.1.1
-      DOWNWARD_COIN_ROOT: /home/runner/work/downward/lib/coin
+      DOWNWARD_CPLEX_ROOT: /home/runner/lib/ibm/ILOG/CPLEX_Studio129/cplex
+      DOWNWARD_SOPLEX_ROOT: /home/runner/lib/soplex-3.1.1
+      DOWNWARD_COIN_ROOT: /home/runner/lib/coin
     steps:
+      # Check out repo at $GITHUB_WORKSPACE (= /home/runner/work/my-repo/my-repo).
```
silvansievers, line 36: This is interesting to know, but how does the compilation of the planner below work without switching to this directory? 

Wouldn't it make sense to have this path more explicitly, i.e., checkout to downward in home (= /home/runner) so that we have one directory for the planner and one for the libraries?

.github/workflows/ubuntu.yml
```diff
@@ -110,7 +111,6 @@ jobs:
         # archive the entire lib directory of Osi because we need all
         # 4 large library files and several file links to these.
         run: |
-          cd ../
```
silvansievers, line 113: Why is this not necessary anymore? Isn't the default pwd the home directory?

.github/workflows/ubuntu.yml
```diff
@@ -182,10 +183,8 @@ jobs:
         # during compilation.
         run: |
           tar xfz archive.tar.gz
-          shopt -s dotglob
-          mv downward/* .
           if [[ ! -z "${CPLEX_URL}" || ! -z "${SOPLEX_URL}" ]]; then
-            mv lib/ ../
+            mv lib/ /home/runner
```
silvansievers, line 187: Why don't we have to move downward (or however the repo is called) to the "right" place?

silvansievers (commented):
All of my comments basically relate to the fact that I find the paths used by Github actions somewhat obscure. If you understood how they work, I would appreciate if you either documented the paths implicitly used when checking out/compiling/moving directories or to explicitly provide the paths (e.g., use cd /home/runner/downward even if this is not necessary etc.).

.github/workflows/ubuntu.yml
jendrikseipp: The checkout path is configurable, but it must lie under $GITHUB_WORKSPACE.

.github/workflows/ubuntu.yml
jendrikseipp: I added some documentation about this. The pwd is $GITHUB_WORKSPACE.

.github/workflows/ubuntu.yml
jendrikseipp: Because the working dir is $GITHUB_WORKSPACE :-)

silvansievers (approved):

PR #43, opened 2021-05-25
Improve errors

victorpaleologue:
Slightly better error messages for some corner cases:
- when PDDL does not start with `(define `
- when a condition is empty (`()`)

### DIFF
--- a/src/translate/pddl_parser/parsing_functions.py
+++ b/src/translate/pddl_parser/parsing_functions.py
@@ -61,6 +61,8 @@ def parse_condition(alist, type_dict, predicate_dict):
 
 def parse_condition_aux(alist, negated, type_dict, predicate_dict):
     """Parse a PDDL condition. The condition is translated into NNF on the fly."""
+    if len(alist) == 0:
+        raise RuntimeError("Empty condition expression")
     tag = alist[0]
     if tag in ("and", "or", "not", "imply"):
         args = alist[1:]
--- a/src/translate/pddl_parser/parsing_functions.py
+++ b/src/translate/pddl_parser/parsing_functions.py
@@ -317,7 +317,7 @@ def parse_domain_pddl(domain_pddl):
     iterator = iter(domain_pddl)
 
     define_tag = next(iterator)
-    assert define_tag == "define"
+    assert define_tag == "define", "define_tag = \"%s\" instead of \"define\"" % define_tag
     domain_line = next(iterator)
     assert domain_line[0] == "domain" and len(domain_line) == 2
     yield domain_line[1]


src/translate/pddl_parser/parsing_functions.py
```diff
@@ -317,7 +319,7 @@ def parse_domain_pddl(domain_pddl):
     iterator = iter(domain_pddl)
 
     define_tag = next(iterator)
-    assert define_tag == "define"
+    assert define_tag == "define", "define_tag = \"%s\" instead of \"define\"" % define_tag
```
roeger, line 322: This is better than the current state but still not very obvious because it refers to the internals of the code. Maybe the text should rather be something along the lines of 'The domain specification must begin with "(define" but begins with "(%s".'

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -61,6 +61,8 @@ def parse_condition(alist, type_dict, predicate_dict):
 
 def parse_condition_aux(alist, negated, type_dict, predicate_dict):
     """Parse a PDDL condition. The condition is translated into NNF on the fly."""
+    if len(alist) == 0:
```
roeger, line 64: This should be more pythonic:

```
if not alist:
   ...

```
Where did this happen to you?

roeger (changes requested):

roeger:
Hi Victor,

I added a review with some comments. But even if you changed this, I could not accept the pull request in the current form.

In particular, you would have to open an issue in the issue tracker (http://issues.fast-downward.org) and follow the instructions of our git workflow (http://www.fast-downward.org/ForDevelopers/Git) 1.-7 (skip the change notes for something small like this).

In the long term, we would like to significantly improve the error reporting of the translator (issue220) but since we also want to output the context of the problem (e.g. in which action/line), this would require a larger revision of the parsing code.

If you want to identify problems in your PDDL files, I would for the moment recommend to use the parser component of VAL (https://github.com/KCL-Planning/VAL) for this purpose.

Best,

Gabi

src/translate/pddl_parser/parsing_functions.py
victorpaleologue: I am not sure I remember. It was either an action with no precondition, or an empty goal statement.

src/translate/pddl_parser/parsing_functions.py
victorpaleologue: To continue in this PR: https://github.com/aibasel/downward/pull/59

src/translate/pddl_parser/parsing_functions.py
victorpaleologue: To continue in PR: https://github.com/aibasel/downward/pull/60

victorpaleologue:
Split into two corrected PRs: https://github.com/aibasel/downward/pull/59 & https://github.com/aibasel/downward/pull/60

PR #44, opened 2021-05-25
Improve build

victorpaleologue:
- Let host decide which compiler to use.
- Exclude build files from Git index.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,10 @@
 __pycache__/
+*.pyc
+*.o
+CMakeFiles/*
+CMakeCache.txt
+Makefile
+cmake_install.cmake
 /output.sas
 /sas_plan
 /sas_plan.*
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -30,10 +30,6 @@
 # Version 2.8.3 introduces CMakeParseArguments.
 cmake_minimum_required(VERSION 2.8.3)
 
-# Respect the PATH environment variable when searching for compilers.
-find_program(CMAKE_C_COMPILER NAMES $ENV{CC} gcc PATHS ENV PATH NO_DEFAULT_PATH)
-find_program(CMAKE_CXX_COMPILER NAMES $ENV{CXX} g++ PATHS ENV PATH NO_DEFAULT_PATH)
-
 # Path containing custom CMake modules
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake_modules)
 include(FastDownwardMacros)


.gitignore
```diff
@@ -1,4 +1,10 @@
 __pycache__/
+*.pyc
+*.o
+CMakeFiles/*
+CMakeCache.txt
+Makefile
+cmake_install.cmake
```
FlorianPommerening, line 7: Our build system always builds in the directory `builds` which is already ignored (in line 5 of the old file). What is the reason to also ignore the build artifacts directly?

src/CMakeLists.txt
```diff
@@ -30,10 +30,6 @@
 # Version 2.8.3 introduces CMakeParseArguments.
 cmake_minimum_required(VERSION 2.8.3)
 
-# Respect the PATH environment variable when searching for compilers.
```
FlorianPommerening, line 33: We use these lines so CMake respects the default compiler set in the PATH (see https://stackoverflow.com/questions/29902862/how-to-get-cmake-to-use-the-default-compiler-on-system-path). This is important for our automated tests. What is the reason for removing them?

The PR comment says that this is done so the host can decide which compiler to use. This is already possible like this:
```CC=/path/to/gcc CXX=/path/to/g++ ./build.py```

**Review by FlorianPommerening — COMMENTED** [review]
I do not understand the intention of these changes. Could you comment on the questions I'm adding below?

File: .gitignore (hunk 1 above)
Comment (victorpaleologue) [reply .gitignore:7]: At the time I coded this, Fast Downward was not on Git yet. It was Mercurial, and I had Gitified it myself.
For sure now I trust your .gitignore more. I will remove the commit.

File: src/CMakeLists.txt (hunk 2 above)
Comment (victorpaleologue) [reply src/CMakeLists.txt:33]: The problem is that you force the look-up for g++, when some clients like me rather use clang.
According to https://stackoverflow.com/questions/29902862/how-to-get-cmake-to-use-the-default-compiler-on-system-path, the system already looks up for your default compiler under the names `cc` and `c++`. If that does not work for you, that means that the `PATH` you provide for your tests do not provide these binaries. Show me your automated tests and let's try to fix it.

**Comment by victorpaleologue** [comment]
Superseded by https://github.com/aibasel/downward/pull/61

File: src/CMakeLists.txt (hunk 2 above)
Comment (FlorianPommerening) [reply src/CMakeLists.txt:33]: The code here uses `gcc` and `g++` only as a fallback. The first choice is `$ENV{CC}` and `$ENC{CXX}`, so it already uses `CC` and `CXX`. This is the goal of those lines: if `CC` and `CXX` are set, they should be used instead of the system's default compiler. You can use clang with `CC=clang CXX=clang++ ./build.py`.

Our automated tests do this as well: https://github.com/aibasel/downward/blob/main/.github/workflows/ubuntu.yml
The relevant lines are 33 and 34:
```
      CC: ${{ matrix.compiler-version.cc }}
      CXX: ${{ matrix.compiler-version.cxx }}
```

Maybe a better fix would be to add `cc` and `c++` as a secondary choice (after `$ENV{CC}` but before `gcc`) in the `find_program` call?

FlorianPommerening: Actually, since this PR is now closed, lets continue the discussion in the issue tracker.

PR #45, opened 2021-05-25
Support compilation on Android

victorpaleologue:
Even if that does not make Fast Downward immediately usable on Android, this overcomes build issues with the Android NDK.


silvansievers:
We discussed this pull request within the core team and we see little benefit in compiling Fast Downward on Android. For this reason, we don't want to invest the maintenance overhead of officially supporting Android.

PR #50, opened 2021-06-21
Issue1008

silvansievers:


src/search/pdbs/rcg.cc
```diff
@@ -0,0 +1,87 @@
...
+#include "../utils/logging.h"
+#include "../utils/rng.h"
+
+#include <algorithm>
+#include <unordered_set>
+
+using namespace std;
+
+namespace pdbs{
+static bool time_limit_reached(
```
FlorianPommerening, line 17: This looks like it belongs to `utils::CountdownTimer`. Is this the right place to implement this?

src/search/pdbs/rcg.cc
```diff
@@ -0,0 +1,87 @@
...
+    if (timer.is_expired()) {
+        if (verbosity >= utils::Verbosity::NORMAL) {
+            utils::g_log << "time limit reached." << endl;
+        }
+        return true;
+    }
+    return false;
+}
+
+Pattern generate_pattern_rcg(
```
FlorianPommerening, line 28: What does RCG stand for? Is this something I should be familiar with? If not, is it small enough that we can use the non-abbreviated version in the file or class name?

src/search/pdbs/rcg.cc
```diff
@@ -0,0 +1,87 @@
...
+    int max_pdb_size,
+    double max_time,
+    int goal_variable,
+    utils::Verbosity verbosity,
+    const shared_ptr<utils::RandomNumberGenerator> &rng,
+    const TaskProxy &task_proxy,
+    vector<vector<int>> &cg_neighbors) {
```
FlorianPommerening, line 35: `const`?

src/search/pdbs/rcg.cc
```diff
@@ -0,0 +1,87 @@
...
+    utils::CountdownTimer timer(max_time);
+    int current_var = goal_variable;
+    unordered_set<int> visited_vars;
+    visited_vars.insert(current_var);
+    int pdb_size = task_proxy.get_variables()[current_var].get_domain_size();
```
FlorianPommerening, line 40: `task_proxy.get_variables()` is used three times and could be stored in a local variable.

src/search/pdbs/rcg.cc
```diff
@@ -0,0 +1,87 @@
...
+    parser.add_option<int>(
+        "max_pdb_size",
+        "maximum number of states per pattern database (ignored for the "
+        "initial collection consisting of singleton patterns for each goal "
+        "variable)",
+        "2000000",
+        Bounds("1", "infinity"));
+    parser.add_option<double>(
+        "max_time",
+        "maximum time in seconds for the RCG algorithm (ignored for"
```
FlorianPommerening, line 77: missing space at the line break.

src/search/pdbs/rcg.cc
```diff
@@ -0,0 +1,87 @@
...
+        "computing initial collection)",
+        "infinity",
+        Bounds("0.0", "infinity"));
+    parser.add_option<bool>(
+        "bidirectional",
+        "consider pre-eff edges of the causal graph in both directions",
```
FlorianPommerening, line 83: This could be clearer: you could describe the default behavior first and then describe what changes when this parameter is changed.

src/search/pdbs/pattern_generator_rcg.h
```diff
@@ -0,0 +1,31 @@
...
+namespace pdbs {
+class PatternGeneratorRCG : public PatternGenerator {
+    const int max_pdb_size;
+    const int max_time;
+    const bool bidirectional;
+    const utils::Verbosity verbosity;
+    std::shared_ptr<utils::RandomNumberGenerator> rng;
+public:
+    explicit PatternGeneratorRCG(options::Options &opts);
+    virtual ~PatternGeneratorRCG() = default;
```
FlorianPommerening, line 24: Isn't our policy to leave out the destructor if it has a default implementation?

src/search/pdbs/pattern_generator_rcg.cc
```diff
@@ -0,0 +1,76 @@
...
+#include "../utils/timer.h"
+
+#include <vector>
+
+using namespace std;
+
+namespace pdbs {
+
+PatternGeneratorRCG::PatternGeneratorRCG(
+    options::Options& opts)
```
FlorianPommerening, line 23: Does this fit in one line?

src/search/pdbs/pattern_generator_rcg.cc
```diff
@@ -0,0 +1,76 @@
...
+    : max_pdb_size(opts.get<int>("max_pdb_size")),
+      max_time(opts.get<double>("max_time")),
+      bidirectional(opts.get<bool>("bidirectional")),
+      verbosity(opts.get<utils::Verbosity>("verbosity")),
+      rng(utils::parse_rng_from_options(opts)) {
+}
+
+PatternInformation PatternGeneratorRCG::generate(
+    const std::shared_ptr<AbstractTask> &task) {
```
FlorianPommerening, line 32: one line?

src/search/pdbs/pattern_generator_rcg.cc
```diff
@@ -0,0 +1,76 @@
...
+    utils::Timer timer;
+    if (verbosity >= utils::Verbosity::NORMAL) {
+        utils::g_log << "Single RCG: generating patterns" << endl;
```
FlorianPommerening, line 35: Write "RCG" without an abbreviation here.

src/search/pdbs/pattern_generator_rcg.cc
```diff
@@ -0,0 +1,76 @@
...
+    }
+    vector<vector<int>> cg_neighbors = compute_cg_neighbors(
+        task, bidirectional);
```
FlorianPommerening, line 38: one line?

src/search/pdbs/pattern_generator_rcg.cc
```diff
@@ -0,0 +1,76 @@
...
+    TaskProxy task_proxy(*task);
+    vector<FactPair> goals = get_goals_in_random_order(task_proxy, *rng);
+
+    Pattern pattern = generate_pattern_rcg(
+        max_pdb_size,
```
FlorianPommerening, line 43: I actually like this formatting but I think it doesn't follow our conventions.

src/search/pdbs/pattern_generator_rcg.cc
```diff
@@ -0,0 +1,76 @@
...
+
+    Options opts = parser.parse();
+    if (parser.dry_run()) {
+        return nullptr;
+    }
+
+    return make_shared<PatternGeneratorRCG>(opts);
+}
+
+static Plugin<PatternGenerator> _plugin("rcg_pattern", _parse);
```
FlorianPommerening, line 75: The naming convention so far is a bit inconsistent. Most names do not mention "pattern" in the option name, some use "pattern"/"patterns" to distinguish between generating a pattern or a collection, some use "single"/multiple" for that reason. Can we make this more consistent?

FlorianPommerening: Also, I'd prefer a name without abbreviation as I still do not know what RCG stands for.

src/search/pdbs/pattern_collection_generator_rcg.cc
```diff
@@ -0,0 +1,309 @@
...
+        "total_max_time",
+        "maximum time in seconds for the multiple RCG algorithm. The "
+        "algorithm will always execute at least one iteration, i.e., call the "
+        "RCG algorithm once. This limit possibly overrides the limit "
+        "specified for the RCG algorithm.",
+        "100.0",
+        Bounds("0.0", "infinity"));
+    parser.add_option<double>(
+        "stagnation_limit",
+        "maximum time in seconds the multiple RCG algorithm allows without "
```
FlorianPommerening, line 277: Are the terms "RCG algorithm" and "multiple RCG algorithms" describing different things? This is really hard to understand. I think what is missing here is a synopsis entry that explains what the pattern generator actually does.

src/search/pdbs/pattern_collection_generator_rcg.h
```diff
@@ -0,0 +1,59 @@
...
+        const TaskProxy &task_proxy,
+        const Pattern &pattern,
+        std::set<Pattern> &generated_patterns,
+        const utils::CountdownTimer &timer);
+    bool collection_size_limit_reached() const;
+    bool time_limit_reached(const utils::CountdownTimer &timer) const;
+    bool check_for_stagnation(const utils::CountdownTimer &timer);
+public:
+    explicit PatternCollectionGeneratorRCG(options::Options &opts);
+    virtual ~PatternCollectionGeneratorRCG() = default;
```
FlorianPommerening, line 52: Remove destructor.

src/search/pdbs/pattern_collection_generator_rcg.cc
```diff
@@ -0,0 +1,309 @@
...
+        "maximum number of states in the pattern collection (ignored for the "
+        "initial collection consisting of singleton patterns for each goal "
+        "variable)",
+        "20000000",
+        Bounds("1", "infinity"));
+    parser.add_option<double>(
+        "total_max_time",
+        "maximum time in seconds for the multiple RCG algorithm. The "
+        "algorithm will always execute at least one iteration, i.e., call the "
+        "RCG algorithm once. This limit possibly overrides the limit "
```
FlorianPommerening, line 271: overriding an option that we pass on a different level reminds me of a long standing "option caveat" we had in the landmark heuristic that took some effort to get rid off. Are you sure this is a good idea here?

FlorianPommerening (commented):
I didn't quite finish with `pattern_collection_generator_rcg.cc` but I have to leave now, so I'm posting the comments I already have.

src/search/pdbs/rcg.cc
silvansievers: I don't think it should be a method of `utils::CountdownTimer`: there is already `is_expired` for this and I just want to have the added output. It could be a function in the `utils::countdown_timer`, if you prefer.

src/search/pdbs/rcg.cc
silvansievers: No, the neighbors of a variable are shuffled after each use.

src/search/pdbs/pattern_collection_generator_rcg.cc
silvansievers: I don't think it that this is problematic. It seems natural to have a time limit for each call to a function which generates a single pattern and to have an overall time budget. We do something similar when we limit the size of an individual PDB and of all PDBs.

src/search/pdbs/pattern_generator_rcg.cc
silvansievers: I'm also not sure, but I also don't see that they would directly violate them: http://www.fast-downward.org/ForDevelopers/Whitespace
There is no example on what happens when the function call is too large to fit into one line even when breaking after the opening parenthesis. I used this style you see here not only here...

src/search/pdbs/pattern_generator_rcg.cc
silvansievers: the line would actually be too long

src/search/pdbs/pattern_generator_rcg.cc
silvansievers: too long

src/search/pdbs/pattern_generator_rcg.cc
silvansievers: this one does

src/search/pdbs/pattern_collection_generator_rcg.cc
silvansievers: I reworked the option structure entirely with the introduction of the common base class.

src/search/pdbs/rcg.cc
FlorianPommerening: Yes, I think it makes more sense there. You could also extend `is_expired` by adding an optional `logger` argument.

src/search/pdbs/pattern_collection_generator_multiple.h
```diff
@@ -0,0 +1,69 @@
...
+    virtual PatternInformation compute_pattern(
+        int max_pdb_size,
+        double max_time,
+        const std::shared_ptr<utils::RandomNumberGenerator> &rng,
+        const std::shared_ptr<AbstractTask> &task,
+        FactPair goal,
+        std::unordered_set<int> &&blacklisted_variables) = 0;
+public:
+    explicit PatternCollectionGeneratorMultiple(options::Options &opts);
+    virtual ~PatternCollectionGeneratorMultiple() override = default;
```
FlorianPommerening, line 61: remove the default destructor.

src/search/pdbs/pattern_generator_rcg.cc
FlorianPommerening: I interpreted the first example to continue like this:

```
int foobar1 = example_of_breaking_after_parenthesis(
    some_argument, foo + bar, baz, more_arguments,
    even_more_arguments, final_argument);
```

but you're right, it doesn't explicitly say so. I'm fine with this anyway, I like having one parameter per line so I don't want to argue too hard against it.

src/search/pdbs/pattern_generator_rcg.cc
FlorianPommerening: Your recent change fixed the first of these comments but not the second. should we call the plugins "random_pattern(s)"?

FlorianPommerening: Nevermind, I looked at an outdated diff.

src/search/pdbs/pattern_collection_generator_multiple.cc
```diff
@@ -0,0 +1,310 @@
...
+    if (generated_patterns.insert(move(pattern)).second) {
+        // RCG generated a new pattern. Reset stagnation_start_time.
+        stagnation_start_time = -1;
+        shared_ptr<PatternDatabase> pdb = pattern_info.get_pdb();
+        remaining_collection_size -= pdb->get_size();
+        generated_pdbs->push_back(move(pdb));
+    } else {
+        // Pattern is not new. Set stagnation start time if not already set.
+        if (stagnation_start_time == -1) {
+            stagnation_start_time = timer.get_elapsed_time();
```
FlorianPommerening, line 95: I don't think it matters too much but it feels like the stagnation time is shifted by one evaluation: If I find pattern P 10 seconds into the algorithm, then find P again at 20 seconds, and then again at 30 seconds, I'd say, the method was stuck for 30 - 10 = 20 seconds. But here the time would start counting at 20, so it would consider the method not progressing for 20 - 30 = 10 seconds.

src/search/pdbs/pattern_collection_generator_multiple.cc
```diff
@@ -0,0 +1,310 @@
...
+            if (verbosity >= utils::Verbosity::NORMAL) {
+                utils::g_log << "stagnation limit reached, terminating" << endl;
+            }
+            return true;
+        }
+    }
+    return false;
+}
+
+static PatternCollectionInformation get_pattern_collection(
```
FlorianPommerening, line 158: This looks like a general utility function for PDBs. Is this useful elsewhere?

src/search/pdbs/pattern_collection_generator_multiple.cc
```diff
@@ -0,0 +1,310 @@
...
+        double remaining_time =
+            min(static_cast<double>(timer.get_remaining_time()), pattern_generation_max_time);
+
+        PatternInformation pattern_info = compute_pattern(
+            remaining_pdb_size,
+            remaining_time,
+            pattern_computation_rng,
+            task,
+            goals[goal_index],
+            move(blacklisted_variables));
```
FlorianPommerening, line 226: I didn't look at how this is used yet, but it seems strange that we move `blacklisted_variables` instead of passing a const reference.

src/search/pdbs/pattern_collection_generator_multiple.cc
```diff
@@ -0,0 +1,310 @@
...
+        "total_max_time",
+        "maximum time in seconds for this pattern collection generator. The "
+        "algorithm will always execute at least one iteration, i.e., call the "
+        "algorithm for computing a single pattern at least once.",
+        "100.0",
+        Bounds("0.0", "infinity"));
+    parser.add_option<double>(
+        "stagnation_limit",
+        "maximum time in seconds the algorithm allows without generating a "
+        "new pattern. The multiple algorithm terminates prematurely if this "
```
FlorianPommerening, line 290: "The multiple algorithm" sounds strange, especially without context. Maybe just replace this with "Pattern generation terminates ...".

src/search/pdbs/pattern_collection_generator_multiple.cc
```diff
@@ -0,0 +1,310 @@
...
+        Bounds("1.0", "infinity"));
+    parser.add_option<double>(
+        "blacklist_trigger_percentage",
+        "percentage of total_max_time after which the algorithm enables "
+        "blacklisting for diversification",
+        "0.75",
+        Bounds("0.0", "1.0"));
+    parser.add_option<bool>(
+        "enable_blacklist_on_stagnation",
+        "if true, the multiple algorithm will enable blacklisting "
```
FlorianPommerening, line 302: Same here. It could just be "if true, blacklisting is enabled ..."

src/search/pdbs/pattern_collection_generator_multiple_random.cc
```diff
@@ -0,0 +1,91 @@
...
+}
+
+PatternInformation PatternCollectionGeneratorMultipleRandom::compute_pattern(
+    int max_pdb_size,
+    double max_time,
+    const shared_ptr<utils::RandomNumberGenerator> &rng,
+    const shared_ptr<AbstractTask> &task,
+    FactPair goal,
+    unordered_set<int> &&) {
+    // TODO: add support for blacklisting in single RCG?
```
FlorianPommerening, line 46: While this is not yet supported, should we check for it here and warn the user if it is used?

src/search/pdbs/pattern_collection_generator_multiple_random.cc
```diff
@@ -0,0 +1,91 @@
...
+    const utils::Verbosity random_pattern_verbosity(utils::Verbosity::SILENT);
```
FlorianPommerening, line 47: I don't think this needs a local variable: it is only used once and the name of the constant (`SILENT`) is more self-documenting than the name of the variable (`random_pattern_verbosity`).

src/search/pdbs/pattern_collection_generator_multiple_random.cc
```diff
@@ -0,0 +1,91 @@
+#include "pattern_collection_generator_multiple_random.h"
+#include "pattern_database.h"
+#include "random_pattern.h"
+#include "utils.h"
+
+#include "../option_parser.h"
+#include "../plugin.h"
+#include "../task_proxy.h"
+
+#include "../task_utils/causal_graph.h"
```
FlorianPommerening, line 10: Do we still need all of those includes? looks like most should now be included in the base class (probbaly the same for cegar).

src/search/pdbs/pattern_collection_generator_single_cegar.cc
```diff
@@ -56,7 +56,28 @@ static shared_ptr<PatternCollectionGenerator> _parse(
             "AAAI Press",
             "2019"));
     add_implementation_notes_to_parser(parser);
-    add_cegar_options_to_parser(parser);
+    parser.add_option<int>(
```
FlorianPommerening, line 59: Could you add a TODO and an issue to move these options to the base class of pattern generators? I think we really want them everywhere.

src/search/pdbs/rcg.cc
silvansievers: I looked into the usages of `utils::CountdownTimer::is_expired()` and would argue to not add anything to `utils::coutdown_timer.h` that has to do with output. All users of `utils::CountdownTimer::is_expired()` print their own message and possibly do additional things in the if-clause. We could pass both `Log logger` and `string message` to the timer class(es), but I don't really think it belongs there.

src/search/pdbs/pattern_collection_generator_multiple.h
silvansievers: I think it must be there because it is a base class: http://www.fast-downward.org/ForDevelopers/CodingConventions

src/search/pdbs/pattern_collection_generator_multiple.cc
```diff
@@ -0,0 +1,310 @@
...
+    vector<int> &non_goal_variables) {
+    unordered_set<int> blacklisted_variables;
+    if (blacklisting && !non_goal_variables.empty()) {
+        /*
+          Randomize the number of non-goal variables for blacklisting.
+          We want to choose at least 1 non-goal variable and up to the
+          entire set of non-goal variables.
+        */
+        int blacklist_size = (*rng)(non_goal_variables.size());
+        ++blacklist_size; // [1, |non-goal variables|]
```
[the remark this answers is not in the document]
silvansievers: We want to have a random int in the interval [1, number of non goal variables].

src/search/pdbs/random_pattern.cc
```diff
@@ -0,0 +1,79 @@
...
+            }
+        }
+
+        if (neighbor_var != -1 && utils::is_product_within_limit(
+                pdb_size, variables[neighbor_var].get_domain_size(), max_pdb_size)) {
+            pdb_size *= variables[neighbor_var].get_domain_size();
+            visited_vars.insert(neighbor_var);
+            current_var = neighbor_var;
+        } else {
+            break;
```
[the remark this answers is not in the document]
silvansievers: Good suggestion! I will implement this later and run an experiment to test the impact of the change.

src/search/pdbs/pattern_collection_generator_multiple.cc
silvansievers: I agree. Changing this, however, would mean to run experiments with the multiple CEGAR algorithm, too, so I'm not too eager to do this, but I could be persuaded.

src/search/pdbs/pattern_collection_generator_multiple.cc
silvansievers: I'm not aware of any place, but we could still move it to `utils`, if you like.

src/search/pdbs/pattern_collection_generator_multiple.cc
silvansievers: I agree, but we left this like this because CEGAR not only wants to store this set of variables, but also extends it.

src/search/pdbs/rcg.cc
FlorianPommerening: OK, fine with me. Maybe you could go in the other direction and make the log message here more specific as well. What kind of time limit was reached, what does this mean for the computation? That sort of thing.

src/search/pdbs/pattern_collection_generator_multiple.h
FlorianPommerening: Ahh, right. I forgot about this.

src/search/pdbs/pattern_collection_generator_multiple.cc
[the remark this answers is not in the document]
FlorianPommerening: I see. I assume there is no method for this in the rng? It seems like a common problem.
Anyway, I think the comment is not clear. It also is almost covered by the comment in lines 56-57.
How about changing 
```
 We want to choose at least 1 non-goal variable and up to the entire set of non-goal variables.
````
to 
```
 We want to choose at least 1 non-goal variable, so we pick a random value in the range [1, |non-goal variables|].
````
and then removing the comment on line 60?

File: src/search/pdbs/pattern_collection_generator_multiple.cc (hunk 18 above)
Comment (FlorianPommerening) [reply src/search/pdbs/pattern_collection_generator_multiple.cc:95]: If it is possible to implement this without complicating the code too much, I'd be in favor of changing it but if it requires complex changes to keep track of the start value, I don't think it is necessary. If you leave it like this, I think this should be described in the documentation of the parameter because it doesn't do what you'd expect.
On the plus side, changing it could actually be beneficial, as a stagnation is recognized earlier.

File: src/search/pdbs/pattern_collection_generator_multiple.cc (hunk 19 above)
Comment (FlorianPommerening) [reply src/search/pdbs/pattern_collection_generator_multiple.cc:158]: Yes, I think it would be a better place for it.

File: src/search/pdbs/utils.cc [hunk 29]
```diff
@@ -67,6 +68,37 @@ vector<int> get_non_goal_variables(const TaskProxy &task_proxy) {
     return non_goal_variables;
 }
 
+vector<vector<int>> compute_cg_neighbors(
```
thomaskeller79, line 71: I would expect this to be part of the namespace causal_graph, where I would make it slightly more general by splitting the bool bidirectional to a bool add_predeccessors and a bool add_successors (i.e., also supporting the case where only successors are requested).

thomaskeller79 (commented):
I like these changes very much, in particular the creation of the ```PatternCollectionGeneratorMultiple``` class which makes the existing code much clearer.

I'm not sure I fully understood the different interactions between the classes yet and would like to suggest to briefly discuss this. I kind of feel like the random* classes and the cegar* classes should somehow be comparable and something like the *manual* classes should also interact in the same way, but it seems like there is some small irregularity (for instance, I'm not sure why pattern_collection_generator_single_cegar exists but not pattern_collection_generator_single_random, or it seems like you should be able to call pattern_generator_random from within pattern_collection_generator_multiple_random). If you don't think this is necessary, I'm also fine with merging this though.

src/search/pdbs/pattern_collection_generator_multiple.cc
silvansievers: I implemented the change. If anything, I think that the logic got slightly easier. Please have a look at the commit.

src/search/pdbs/pattern_collection_generator_multiple.h
```diff
@@ -0,0 +1,69 @@
...
+    void handle_generated_pattern(
+        PatternInformation &&pattern_info,
+        std::set<Pattern> &generated_patterns,
+        std::shared_ptr<PDBCollection> &generated_pdbs,
+        const utils::CountdownTimer &timer);
+    bool collection_size_limit_reached() const;
+    bool time_limit_reached(const utils::CountdownTimer &timer) const;
+    bool check_for_stagnation(const utils::CountdownTimer &timer);
+protected:
+    virtual std::string get_name() const = 0;
```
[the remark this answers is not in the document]
silvansievers: It's not so much about changing this; I'm not quite sure if having a member is a nicer solution and if we have any conventions. I'm also not sure how the subclass would initialize this member: directly in the initializer list? Or rather by passing an argument to a superclass' constructor?

src/search/pdbs/pattern_collection_generator_multiple.cc
```diff
@@ -0,0 +1,311 @@
...
+unordered_set<int> PatternCollectionGeneratorMultiple::get_blacklisted_variables(
+    vector<int> &non_goal_variables) {
+    unordered_set<int> blacklisted_variables;
+    if (blacklisting && !non_goal_variables.empty()) {
+        /*
+          Randomize the number of non-goal variables for blacklisting.
+          We want to choose at least 1 non-goal variable and up to the
+          entire set of non-goal variables.
+        */
+        int blacklist_size = (*rng)(non_goal_variables.size());
```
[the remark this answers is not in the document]
silvansievers: I agree that this is ugly to read. You could ask in Discord if there is any reason for this interface and if otherwise, it is worth opening an issue for this.

src/search/pdbs/pattern_collection_generator_multiple_random.cc
```diff
@@ -0,0 +1,85 @@
+#include "pattern_collection_generator_multiple_random.h"
+
+#include "random_pattern.h"
+#include "utils.h"
+
+#include "../option_parser.h"
+#include "../plugin.h"
+#include "../task_proxy.h"
```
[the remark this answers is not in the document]
silvansievers: It is unnecessary because `utils.h`, which is also included here, includes `task_proxy.h`. But note that this class uses `TaskProxy`. I don't think we have convention for this, but I sometimes like including things that should be included for the obvious reason even if there are already included via several levels of other includes because that isn't obvious, and most of the time is a flaw because header files shouldn't include other header files ideally.

src/search/pdbs/pattern_generator_random.cc
```diff
@@ -0,0 +1,103 @@
+#include "pattern_generator_random.h"
+
+#include "pattern_database.h"
+#include "random_pattern.h"
+#include "utils.h"
+
+#include "../option_parser.h"
+#include "../plugin.h"
+#include "../task_proxy.h"
```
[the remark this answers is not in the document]
silvansievers: same as above

src/search/pdbs/pattern_collection_generator_multiple.cc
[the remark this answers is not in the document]
maltehelmert: Agreed, I don't see a point in making these two separate updates to `blacklist_size`.

src/search/pdbs/pattern_collection_generator_multiple.h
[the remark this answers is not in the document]
thomaskeller79: I would pass it to the superclasses constructor, i.e. change that to 

```PatternCollectionGeneratorMultiple(options::Options &opts, const std::string &name)```

Regarding coding convention, I couldn't find anything (in a very brief 2 minute grep-search) in the code that is directly comparable, except for maybe the ```description``` string of the evaluator class (which is passed as a string in the constructor).

src/search/pdbs/pattern_collection_generator_multiple.cc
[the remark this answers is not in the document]
thomaskeller79: I put this question to discord and close it here.

src/search/pdbs/utils.cc
silvansievers: I'm not so sure. The `CausalGraph` class itself already provides the members `get_successors` and `get_predecessors`. Providing a function to store them doesn't make too much sense to me. This is seems specific to the application we have here, which stores the neighbors to be able to shuffle them.

silvansievers: We decided to not change anything here.

thomaskeller79:
I had a look at all the commits since my last review. Everything looks very good and I think this is ready to get merged.

PR #52, opened 2021-06-23
issue751: store g-values in GEvaluator

jendrikseipp:
For now, I only tackled eager search.


src/search/search_engines/plugin_astar.cc
```diff
@@ -38,9 +42,14 @@ static shared_ptr<SearchEngine> _parse(OptionParser &parser) {
 
     shared_ptr<eager_search::EagerSearch> engine;
     if (!parser.dry_run()) {
-        auto temp = search_common::create_astar_open_list_factory_and_f_eval(opts);
-        opts.set("open", temp.first);
-        opts.set("f_eval", temp.second);
+        shared_ptr<Evaluator> g = make_shared<g_evaluator::GEvaluator>(
```
FlorianPommerening, line 45: This change seems unrelated to the issue. Pulling out the computation of "f = g + h" defeats the purpose of a function that specifically creates the open list of A*, in my opinion, i.e., that part is what makes it an A* open list..

src/search/search_engines/plugin_astar.cc
```diff
@@ -38,9 +42,14 @@ static shared_ptr<SearchEngine> _parse(OptionParser &parser) {
...
+            tasks::g_root_task);
```
FlorianPommerening, line 46: Is it necessary to use the root task here?

src/search/search_engines/search_common.cc
```diff
@@ -98,7 +99,7 @@ shared_ptr<OpenListFactory> create_wastar_open_list_factory(
         options.get_list<shared_ptr<Evaluator>>("evals");
     int w = options.get<int>("w");
 
-    shared_ptr<GEval> g_eval = make_shared<GEval>();
+    shared_ptr<GEval> g_eval = make_shared<GEval>(tasks::g_root_task);
```
FlorianPommerening, line 102: see above

src/search/search_engines/search_common.h
```diff
@@ -68,15 +68,13 @@ extern std::shared_ptr<OpenListFactory> create_wastar_open_list_factory(
...
 
 /*
-  Create open list factory and f_evaluator (used for displaying progress
-  statistics) for A* search.
+  Create open list factory for A* search.
 
   The resulting open list factory produces a tie-breaking open list
-  ordered primarily on g + h and secondarily on h. Uses "eval" from
-  the passed-in Options object as the h evaluator.
+  ordered primarily on g + h and secondarily on h.
```
FlorianPommerening, line 74: This is the part where I don't like the way the new function is split up: the comment says that the open list is ordered by g+ h, but the parameters of the functions are h and f, so the comment is only true if the user puts in the correct f.

src/search/search_node_info.cc
```diff
@@ -1,10 +1,6 @@
...
 
-static const int pointer_bytes = sizeof(void *);
-static const int info_bytes = 3 * sizeof(int) + sizeof(StateID);
-static const int padding_bytes = info_bytes % pointer_bytes;
-
 static_assert(
-    sizeof(SearchNodeInfo) == info_bytes + padding_bytes,
-    "The size of SearchNodeInfo is larger than expected. This probably means "
-    "that packing two fields into one integer using bitfields is not supported.");
+    sizeof(SearchNodeInfo) == 4 + sizeof(StateID) + sizeof(OperatorID),
```
FlorianPommerening, line 4: This ignores word alignment now.

src/search/search_node_info.h
```diff
@@ -10,15 +10,14 @@
 struct SearchNodeInfo {
     enum NodeStatus {NEW = 0, OPEN = 1, CLOSED = 2, DEAD_END = 3};
 
-    unsigned int status : 2;
-    int g : 30;
+    // TODO: pack this into 8 bytes.
```
FlorianPommerening, line 13: This might be more tricky than expected with the custom ID classes.

src/search/search_node_info.h
```diff
@@ -10,15 +10,14 @@
...
+    NodeStatus status;
     StateID parent_state_id;
     OperatorID creating_operator;
-    int real_g;
 
     SearchNodeInfo()
-        : status(NEW), g(-1), parent_state_id(StateID::no_state),
-          creating_operator(-1), real_g(-1) {
+        : status(NEW), parent_state_id(StateID::no_state),
+          creating_operator(OperatorID::no_operator) {
```
FlorianPommerening, line 20: Nice, we should have done this when we introduced the class.

src/search/evaluators/g_evaluator.cc
```diff
@@ -5,15 +5,40 @@
...
 
+#include "../tasks/root_task.h"
+
 using namespace std;
 
 namespace g_evaluator {
+static const int INFTY = numeric_limits<int>::max();
+
+GEvaluator::GEvaluator(const shared_ptr<AbstractTask> &task)
+    : task(task),
```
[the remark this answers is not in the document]
jendrikseipp: I agree. Do you have a solution for this?

src/search/search_engine.cc
```diff
@@ -60,6 +60,9 @@ SearchEngine::SearchEngine(const Options &opts)
         utils::exit_with(ExitCode::SEARCH_INPUT_ERROR);
     }
     bound = opts.get<int>("bound");
+    if (bound != numeric_limits<int>::max()) {
+        real_g_evaluator = make_shared<g_evaluator::GEvaluator>(task);
```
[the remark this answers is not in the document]
jendrikseipp: Good point! I'll add a special case for that.

src/search/search_engines/eager_search.cc
```diff
@@ -182,28 +200,34 @@ SearchStatus EagerSearch::step() {
...
         bool is_preferred = preferred_operators.contains(op_id);
 
-        SearchNode succ_node = search_space.get_node(succ_state);
-
+        int succ_g_old = EvaluationContext(succ_state).get_evaluator_value_or_infinity(
+            g_evaluator.get());
         for (Evaluator *evaluator : path_dependent_evaluators) {
             evaluator->notify_state_transition(s, op_id, succ_state);
         }
+        EvaluationContext succ_eval_context(succ_state, is_preferred, &statistics);
```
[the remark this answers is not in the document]
jendrikseipp: I also thought so. But I didn't see a clean workaround. Do you?

src/search/search_engines/plugin_astar.cc
jendrikseipp: I split it because I didn't see a good way to return the open list, the f-evaluator and the g-evaluator from the function. So you see a good way to do so? Or should I pass the g-evaluator to the function (removing the need to return it)?

src/search/search_engines/plugin_astar.cc
jendrikseipp: We can't use the task from the search engine, because it is hard-coded (to g_root_task) in the search engine ctor, and not passed to the search engine via the options.

src/search/search_node_info.cc
jendrikseipp: I thought so as well, but this version of the code compiles. My guess is that there will be padding once more than one SearchNodeInfo is concatenated in memory.

src/search/search_engines/eager_search.cc
```diff
@@ -33,6 +38,15 @@ EagerSearch::EagerSearch(const Options &opts)
...
         utils::exit_with(utils::ExitCode::SEARCH_INPUT_ERROR);
     }
+    // Even for GBFS, we need a g-evaluator for reporting progress.
+    if (!g_evaluator) {
+        shared_ptr<AbstractTask> task = tasks::g_root_task;
+        OperatorCost cost_type = opts.get<OperatorCost>("cost_type");
+        if (cost_type != OperatorCost::NORMAL) {
+            task = make_shared<tasks::CostAdaptedTask>(task, cost_type);
+        }
+        g_evaluator = make_shared<g_evaluator::GEvaluator>(task);
```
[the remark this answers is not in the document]
jendrikseipp: I agree, this is definitely problematic in the current patch. I see two options: 1) we remove cost_type from the options and expose g_eval on the commandline (breaking backwards compatibility of the configs). 2) we make sure that users can never pass cost_type and g_eval at the same time. I tried to use 2) in the patch, but I forgot to use the cost_type for creating the g_eval. I'll fix that. I'd like to avoid 1) to avoid breaking old configs, but maybe we'll see later that we have to break them.

src/search/search_engines/plugin_astar.cc
jendrikseipp: I'm working on a different solution that restores the original function and directly inserts the g evaluators into the options object.

src/search/evaluators/g_evaluator.cc
```diff
@@ -22,7 +47,7 @@ static shared_ptr<Evaluator> _parse(OptionParser &parser) {
     if (parser.dry_run())
         return nullptr;
     else
-        return make_shared<GEvaluator>();
+        return make_shared<GEvaluator>(tasks::g_root_task);
```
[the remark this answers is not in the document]
jendrikseipp: This code is only relevant if users use g() on the commandline. It is unrelated to the cost_type option of the search engines. It can still lead to confusion though, I agree.

src/search/evaluators/g_evaluator.cc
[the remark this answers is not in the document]
jendrikseipp: We decided to make GEvaluator a subclass of Heuristic (which will be renamed to CachingEvaluator), which should solved this problem.

src/search/search_engines/eager_search.cc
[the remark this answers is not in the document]
jendrikseipp: We decided to switch to option 1).

src/search/search_engines/eager_search.cc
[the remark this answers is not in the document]
jendrikseipp: This will be resolved once GEvaluator caches its estimates and knows which ones are "dirty".

ClemensBuechner:
We decided to close this pull request because it is outdated (see issue tracker: https://issues.fast-downward.org/issue751).

PR #53, opened 2021-06-28
Issue1024

roeger:


README.md
```diff
@@ -1,4 +1,4 @@
-# Fast Downward
+<img src="misc/images/fast-downward.png" width="800" alt="Fast Downward">
```
PatrickFerber, line 1: On mobile the image size is perfect, on desktop the image is quite large. If something adapting is possible that could be nice, but I can live with it as it is.

PatrickFerber: I remember that we had the convention modules names are written with underscores, scripts with hyphen. What is our convention on other file types?
Now that you write `fast-downward.png` with a hyphen I am a bit worried that people will copy the spelling from this (at the same time there is a hyphen in our url and it is more likely that people copy the name from there).

roeger: I don't think many people will look at the file name. I did not find something analogous in the repo, so I took the hyphen over from other files. Regarding the image size, I had a look what other projects do. I also found it quite large at the beginning, but to stick out in the github GUI I actually makes sense.

PatrickFerber (dismissed):

README.md
```diff
@@ -1,4 +1,4 @@
...
```
PatrickFerber, line 2: I was just wondering, if we want to use the SVG image on the github page. It seems to be possible (https://stackoverflow.com/questions/13808020/include-an-svg-hosted-on-github-in-markdown). Have you tried it?

I like vector graphics, as they are always sharp and this prevents the issue that people will just download the png from the github page instead of downloading the svg under misc/images/... (and most times, I think it is preferable to use the SVG file and even MS PowerPoint seems to suport SVG now).

roeger: I tried it and the drill was drilling the right end of the 'O'. Afterwards, I simplified the svg a bit and at least it now shows up ok if I open it in the browser. Let's give it another try.

PatrickFerber: It is so sharp now, I zoomed in and cut myself on the edge of the drill.

PatrickFerber (approved):

PR #60, opened 2021-07-20
PDDL Parser: more explicit error when facing empty expressions

victorpaleologue:
Supersedes https://github.com/aibasel/downward/pull/43

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,8 @@ after the corresponding tracker issues.
 
 ## Changes since the last release
 
-- none yet
+- Better error message for empty goal.
+<http://issues.fast-downward.org/issue1030>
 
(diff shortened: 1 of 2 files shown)


victorpaleologue:
Discussion are ongoing here: http://issues.fast-downward.org/issue1029

victorpaleologue:
As discussed in http://issues.fast-downward.org/issue1030, we do the check only for goals, where the absence of a goal is clearly not supported (and perhaps should not). For other conditions, the spec is not clear so nothing was changed.

maltehelmert:
Superseded by #155, so I'm closing this. But thank you for bringing this to our attention! This was what ultimately triggered the more extensive rewrite of the PDDL parser.

PR #61, opened 2021-07-20
Do not alter compiler look-up in CMake

victorpaleologue:
Protect the environment, leave it untouched.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -30,9 +30,11 @@
 # Version 2.8.3 introduces CMakeParseArguments.
 cmake_minimum_required(VERSION 2.8.3)
 
-# Respect the PATH environment variable when searching for compilers.
-find_program(CMAKE_C_COMPILER NAMES $ENV{CC} gcc PATHS ENV PATH NO_DEFAULT_PATH)
-find_program(CMAKE_CXX_COMPILER NAMES $ENV{CXX} g++ PATHS ENV PATH NO_DEFAULT_PATH)
+# Select a default compiler because CMake does not respect the PATH environment variable for some generators,
+# e.g. used by existing compute servers in Basel.
+# See https://stackoverflow.com/a/45934279 and http://issues.fast-downward.org/issue1031 for details.
+find_program(CMAKE_C_COMPILER NAMES $ENV{CC} cc gcc PATHS ENV PATH NO_DEFAULT_PATH)
+find_program(CMAKE_CXX_COMPILER NAMES $ENV{CXX} c++ g++ PATHS ENV PATH NO_DEFAULT_PATH)
 
 # Path containing custom CMake modules
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake_modules)


victorpaleologue:
Supersedes https://github.com/aibasel/downward/pull/44

maltehelmert:
Hi Victor, if you want us to make this change, we need to understand why it is better than the current situation. You already started a discussion with Florian about this in the superseded pull request 44, perhaps you can continue this discussion here?

victorpaleologue:
We have started a discussion on https://issues.fast-downward.org/, but it is down so often that this is very hard to keep it up. We are discussing the best solution, and I still plan to update my PR when we reach a conclusion. Do you suggest we move that discussion back to GitHub? Do you know what happens with https://issues.fast-downward.org/ ?

FlorianPommerening:
I thought we had reached a conclusion in the issue tracker to leave the code as is and add a comment to explain the reason. This could just refer to the relevant issue (issue1031).

I have not seen any outages of the tracker recently but I also don't check it every day. It should send emails whenever something changes. Did you not receive those? Also the tracker is still on http://issues.fast-downward.org not https://issues.fast-downward.org/issue1031. We are planning to move it to https but this is not done yet.

victorpaleologue:
Maybe this is the recent update of Firefox that prevents me from reaching the http version =/
I thought the line of code was slightly different from the original. Also the comment must explain what actually fails without it, but nobody knows! So there is some investigation involved to figure it out.

maltehelmert:
I think I am on the most recent version of Firefox with no problems. Perhaps you have an extension enabled that forces https? Anyway, we hope to migrate to https soon, but there isn't currently a timeline.

FlorianPommerening:
> I thought the line of code was slightly different from the original.

Yes, we wanted to add `cc` and `c++`  between the environment variables and the gnu compilers but my point is that we reached a conclusion about what to change.

>Also the comment must explain what actually fails without it, but nobody knows! So there is some investigation involved to figure it out.

If you want to do this investigation, the combinations to test would be the valid combinations of operating system (Windows, Linux, Mac), compiler (g++, clang, MSVC), and CMake generator (Makefile, NMake, Ninja, Visual Studio Project, XCode Project). Not all of the 45 combinations are possible but it is still a lot to test. You could also look through the history and issue tracker discussion to find the issue that originally caused us to introduce the line.

If you want to get this merged quickly, I think a comment that points to the stackoverflow page with the general issue would also be sufficient. Something like this:
`We explicitly select a default compiler because CMake doesn't respect environment variables in all generators.
See https://stackoverflow.com/a/45934279 and http://issues.fast-downward.org/issue1031 for details.`

victorpaleologue:
> I think I am on the most recent version of Firefox with no problems. Perhaps you have an extension enabled that forces https? Anyway, we hope to migrate to https soon, but there isn't currently a timeline.

I had enabled an HTTPS-only mode in Firefox, by setting `dom.security.https_only_mode` to `true` in `about:config`. Firefox would accept HTTP if there wasn't an HTTPS site available. In this precise configuration, Firefox is stuck. It works fine when the option is set to `false` (the default value).

Discussion goes on http://issues.fast-downward.org/issue1031, and I plan to update this PR soon.

maltehelmert:
I suspected something like this. The fact that https does respond (but with an error message) is a problem for https-everywhere-style configurations because there is no clear indication to the browser that https is not actually set up for this host. :-( This is clearly a problem with the setup on our side, and we're increasing our efforts to make https available. But it will likely still take quite a bit of time because university is planning to migrate these servers in the near future, and while that is ongoing, the setup is supposed to remain frozen. We'll try to make it as quick as we can.

src/CMakeLists.txt
```diff
@@ -30,9 +30,11 @@
 # Version 2.8.3 introduces CMakeParseArguments.
 cmake_minimum_required(VERSION 2.8.3)
 
-# Respect the PATH environment variable when searching for compilers.
-find_program(CMAKE_C_COMPILER NAMES $ENV{CC} gcc PATHS ENV PATH NO_DEFAULT_PATH)
-find_program(CMAKE_CXX_COMPILER NAMES $ENV{CXX} g++ PATHS ENV PATH NO_DEFAULT_PATH)
+# Select a default compiler when CMake does not respect the PATH environment variable for some generators,
```
FlorianPommerening, line 33: Could you change the comment to
```# Select a default compiler because CMake does not respect the PATH environment variable for some generators,
# e..g. on the compute servers in Basel.```

FlorianPommerening: Sorry, my first comment had a formatting issue, so if you got this by email, it might have only the second line. I updated  it on Github.

victorpaleologue: Ok, done.

PR #62, opened 2021-08-03
Issues1032

silvansievers:


src/search/task_utils/variable_order_finder.h
```diff
@@ -24,7 +28,7 @@ extern void dump_variable_order_type(VariableOrderType variable_order_type);
   least as long as the variable order finder.
 */
 class VariableOrderFinder {
-    TaskProxy task_proxy;
+    const TaskProxy &task_proxy;
```
jendrikseipp, line 31: I had to look this up :-), but our conventions recommend against a reference here (http://www.fast-downward.org/ForDevelopers/CodingConventions).

src/search/task_utils/variable_order_finder.cc
```diff
@@ -2,6 +2,7 @@
 
 #include "../task_utils/causal_graph.h"
 #include "../utils/logging.h"
+#include "../utils/rng.h"
 #include "../utils/system.h"
 
 #include <algorithm>
```
jendrikseipp, line 8: Is the include still needed?

src/search/pdbs/pattern_collection_generator_combo.cc
```diff
@@ -29,15 +32,16 @@ PatternCollectionInformation PatternCollectionGeneratorCombo::generate(
...
     shared_ptr<PatternCollection> patterns = make_shared<PatternCollection>();
 
-    PatternGeneratorGreedy large_pattern_generator(max_states);
-    const Pattern &large_pattern = large_pattern_generator.generate(task).get_pattern();
-    patterns->push_back(large_pattern);
-
+    PatternGeneratorGreedy large_pattern_generator(max_states, rng);
+    Pattern large_pattern = large_pattern_generator.generate(task).get_pattern();
     set<int> used_vars(large_pattern.begin(), large_pattern.end());
+    patterns->emplace_back(move(large_pattern));
```
jendrikseipp, line 38: I think we generally use push_back() in cases like this (and emplace_back() only if the object is constructed, not just moved).

jendrikseipp (changes requested):

PR #63, opened 2021-08-18
[issue1033] driver: slightly generalize domain file finding scheme

silvansievers:


driver/util.py
```diff
@@ -24,10 +24,11 @@ def find_domain_filename(task_filename):
     Find domain filename for the given task using automatic naming rules.
     """
     dirname, basename = os.path.split(task_filename)
+    basename_root, _ = os.path.splitext(basename)
 
     domain_basenames = [
         "domain.pddl",
-        basename[:3] + "-domain.pddl",
+        basename_root + "-domain.pddl",
```
maltehelmert, line 31: As discussed in the tracker, I think it would be better to have the different cases handled consistently, replacing line 27 with `basename_root, ext = os.path.splitext(basename)` and replacing this line with `basename_root + "-domain.pddl" + ext,`.

PR #70, opened 2021-09-14
Issue1036

ClemensBuechner:
Make the landmark graph progression deterministic and flawless (as described in our HSDIP paper https://ai.dmi.unibas.ch/papers/buechner-keller-icaps2021whsdip.pdf).


src/search/landmarks/landmark_status_manager.cc
```diff
@@ -8,164 +8,216 @@ using namespace std;
...
             }
         }
     }
-
-    return true;
 }
 
-void LandmarkStatusManager::update_lm_status(const State &ancestor_state) {
-    const BitsetView reached = get_reached_landmarks(ancestor_state);
+void LandmarkStatusManager::progress_goal(
```
salome-eriksson, line 157: It might be more efficient to call this function only for goal landmarks (this would require to store either pointers or indices of goal landmarks in the status manager).

src/search/landmarks/landmark_heuristic.cc
```diff
@@ -74,8 +76,7 @@ void LandmarkHeuristic::compute_landmark_graph(const plugins::Options &opts) {
         log << "Generating landmark graph..." << endl;
     }
 
-    shared_ptr<LandmarkFactory> lm_graph_factory =
-        opts.get<shared_ptr<LandmarkFactory>>("lm_factory");
+    auto lm_graph_factory = opts.get<shared_ptr<LandmarkFactory>>("lm_factory");
```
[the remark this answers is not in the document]
ClemensBuechner: My IDE or rather clang-tidy seems to recommend to "use `auto` when initializing with a timplate cast to avoid duplicating the type name". I don't have a strong opinion myself, so I'm fine with changing it back.

src/search/landmarks/landmark_status_manager.cc
```diff
@@ -8,164 +8,216 @@ using namespace std;
...
-    }
+    /*
+      TODO: The old code did some logging in this function. For example, it
+       printed the number of landmarks that hold in the initial state or the
+       goal. I personally don't think this is generally relevant information,
+       but if we want to keep that output, it should be easy to add it again.
+
+      TODO: If we decide to get rid of the logs, we can also remove the logger
+       from the function header.
+    */
```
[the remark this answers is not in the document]
ClemensBuechner: We do have access to the task and therefore also to the initial state in the landmark generation, yes. Do you think I should add it there or just drop it entirely?

src/search/landmarks/landmark_status_manager.cc
```diff
@@ -8,164 +8,216 @@ using namespace std;
...
+            progress_reasonable_ordering(id, past, fut);
+        }
+    }
+}
+
+void LandmarkStatusManager::progress_basic(
+    const BitsetView &parent_past, const BitsetView &parent_fut,
+    const State &parent_ancestor_state, BitsetView &past,
+    BitsetView &fut, const State &ancestor_state) {
+    utils::unused_variable(parent_past);
```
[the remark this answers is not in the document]
ClemensBuechner: When I first implemented this function, it looked quite different and just asserted something for the variables that are here claimed to be unused. Now, `parent_past` is actually used (in line 149 below), but the assertion for `parent_ancestor_state` doesn't make much sense anymore and therefore I'll just remove it.

src/search/landmarks/landmark_status_manager.cc
ClemensBuechner: Similarly, it might be more efficient to loop only over greedy-necessary orderings and reasonable orderings in the according progression functions. Should I compute lookups for these things when the status manager is initialized and store them? Or do you have an even better idea? Doing something in this direction could certainly speed things up in practice, but it doesn't affect big-O.

src/search/landmarks/landmark_status_manager.h
```diff
@@ -9,31 +9,45 @@ namespace landmarks {
...
 
-    BitsetView get_reached_landmarks(const State &state);
+    BitsetView get_past_landmarks(const State &state);
+    BitsetView get_future_landmarks(const State &state);
 
     void update_lm_status(const State &ancestor_state);
 
     void process_initial_state(
         const State &initial_state, utils::LogProxy &log);
-    bool process_state_transition(
```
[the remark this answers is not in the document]
ClemensBuechner: Looking at the implementation, the return value seems to be `false` if the transition induces a self-loop and `true` otherwise. I'm not sure where this information could come in handy, but given that (a) it was never used and (b) it is nowhere documented, I think it's fine to drop the return value entirely.

src/search/landmarks/landmark_status_manager.h
```diff
@@ -9,31 +9,45 @@ namespace landmarks {
...
-        const State &parent_ancestor_state, OperatorID op_id,
-        const State &ancestor_state);
+    void progress(
+        const State & parent_ancestor_state, OperatorID,
```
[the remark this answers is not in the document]
ClemensBuechner: Good question, I don't understand how this got lost...

src/search/landmarks/landmark_status_manager.cc
```diff
@@ -1,171 +1,223 @@
...
-
-    // Mark landmarks reached right now as "reached" (if they are "leaves").
-    for (int id = 0; id < num_landmarks; ++id) {
-        if (!reached.test(id)) {
-            LandmarkNode *node = lm_graph.get_node(id);
-            if (node->get_landmark().is_true_in_state(ancestor_state)) {
-                if (landmark_is_leaf(*node, reached)) {
-                    reached.set(id);
+    ConstBitsetView parent_past =
+        get_past_landmarks_const(parent_ancestor_state);
```
maltehelmert, line 136: The more idiomatic way to handle different const-ness and allow using a const version when we have non-const access is to have both methods called get_past_landmarks (etc.) and implement an automatic conversion from BitsetView to ConstBitsetView.

src/search/landmarks/landmark_status_manager.cc
```diff
@@ -1,171 +1,223 @@
...
+        const Landmark &lm = node->get_landmark();
+        if (lm.is_true_in_state(initial_state)) {
+            assert(past.test(id));
+            for (auto &parent : node->parents) {
+                /*
+                  We assume here that no natural orderings A->B are generated
+                  such that B holds in the initial state. Such orderings would
+                  only be valid in unsolvable problems.
+                */
+                assert(parent.second <= EdgeType::REASONABLE);
```
maltehelmert, line 109: This should perhaps become "==" as we're dropping obedient reasonable orders. Might as well use this now?

src/search/landmarks/landmark_status_manager.cc
```diff
@@ -1,171 +1,223 @@
...
+                /*
+                  Reasonable orderings A->B for which both A and B hold in the
+                  initial state are immediately satisfied. We assume such
+                  orderings are not generated in the first place.
+                */
+                assert(!parent.first->get_landmark().is_true_in_state(
+                           initial_state));
+                utils::unused_variable(parent);
+                fut.set(id);
```
maltehelmert, line 118: Doesn't this loop waste a lot of time in non-debug mode if there are multiple such parents? I find it nonintuitive to have a loop (at least in non-debug mode) where the loop body doesn't depend on the loop variable, even when this is not time-critical, and would perhaps prefer doing the loop purely for the debug test if we want it separately from the fut.set.id() without a loop (instead testing parents for emptiness).

src/search/landmarks/landmark_status_manager.cc
```diff
@@ -1,171 +1,223 @@
...
+        }
+    }
+    return goals;
+}
+
+static vector<pair<LandmarkNode *, vector<LandmarkNode *>>> get_greedy_necessary_children(
+    const LandmarkGraph &graph) {
+    vector<pair<LandmarkNode *, vector<LandmarkNode *>>> orderings;
+    for (auto &node : graph.get_nodes()) {
+        vector<LandmarkNode *> greedy_necessary_children{};
```
maltehelmert, line 23: We don't elsewhere call the default constructor explicitly when this is also what happens implicitly without the {}.

src/search/landmarks/landmark_status_manager.cc
```diff
@@ -1,171 +1,223 @@
...
+    return orderings;
+}
+
 /*
-  By default we mark all landmarks as reached, since we do an intersection when
-  computing new landmark information.
+  By default we mark all landmarks past, since it is the neutral element for
+  set intersection which we emulate in the progression. Similarly, no landmarks
+  are future which is the neutral element for set union.
 */
```
maltehelmert, line 57: I think this comment would work better close to the two relevant lines of code.

src/search/landmarks/landmark_heuristic.cc
```diff
@@ -150,34 +152,42 @@ void LandmarkHeuristic::generate_preferred_operators(
...
+        + utils::format_conference_reference(
+            {"Clemens Büchner", "Thomas Keller", "Salomé Eriksson", "Malte Helmert"},
+            "Landmarks Progression in Heuristic Search",
+            "https://ai.dmi.unibas.ch/papers/buechner-et-al-icaps2023.pdf",
+            "Proceedings of the Thirty-Third International Conference on "
+            "Automated Planning and Scheduling (ICAPS 2023)",
+            "TODO", // TODO: Pages missing.
+            "AAAI Press",
+            "2023"));
+
```
maltehelmert, line 190: > This TODO still needs to be addressed. Do we follow the paper 100%, or is there a difference worth mentioning?

src/search/landmarks/landmark_factory_reasonable_orders_hps.cc
```diff
@@ -394,6 +397,19 @@ class LandmarkFactoryReasonableOrdersHPSFeature : public plugins::TypedFeature<L
...
+        add_option<bool>(
+            "use_obedient_reasonable",
+            "Choose whether obedient-reasonable orderings should also be "
+            "computed. Note that we are unsure whether and how they could be "
+            "used for optimal planning (i.e., in admissible heuristics). "
+            "Furthermore, our experiments suggest that they are not really "
+            "helpful for satisficing planning either, hence the default value "
+            "*false*.",
+            // FIXME: change default value to *false*!
+            "true");
```
maltehelmert, line 412: I wouldn't add this option at this stage just to next rip out the feature again. I'd either leave it unchanged in this issue and rip it out later, or rip it out now.

src/search/landmarks/landmark_cost_assignment.cc
```diff
@@ -174,10 +172,14 @@ lp::LinearProgram LandmarkEfficientOptimalSharedCostAssignment::build_initial_lp
 }
 
 double LandmarkEfficientOptimalSharedCostAssignment::cost_sharing_h_value(
```
maltehelmert, line 174: Have you tested this heuristic?

src/search/landmarks/landmark_cost_assignment.cc
```diff
@@ -19,18 +19,17 @@ using namespace std;
 namespace landmarks {
 LandmarkCostAssignment::LandmarkCostAssignment(
     const vector<int> &operator_costs, const LandmarkGraph &graph)
-    : empty(), lm_graph(graph), operator_costs(operator_costs) {
+    : lm_graph(graph), operator_costs(operator_costs) {
 }
 
 const set<int> &LandmarkCostAssignment::get_achievers(
```
maltehelmert, line 25: I would find the opposite parameter order more logical. But feel free to leave the way it is.

src/search/landmarks/landmark_factory_reasonable_orders_hps.cc
ClemensBuechner: I agree, that's wasted effort. We also have the `TODO` a few lines above for printing a warning when using obedient-reasonable orderings in admissible heuristics. I assume it's okay if I also remove that?

maltehelmert: Yes.

src/search/landmarks/landmark_cost_assignment.cc
ClemensBuechner: Yes, https://ai.dmi.unibas.ch/_experiments/ai/downward/issue1036/data/issue1036-base-optimal-eval/issue1036-base-optimal-compare.html has a triple of `*-seq-opt-bjolp-opt` configurations which use this heuristic with CPLEX. Would you like to see some more tests?

maltehelmert:
No, if you're happy with it I'm good.

Am 9. Juli 2023 19:21:45 MESZ schrieb ClemensBuechner ***@***.***>:
>@ClemensBuechner commented on this pull request.
>
>
>
>> @@ -174,10 +172,14 @@ lp::LinearProgram LandmarkEfficientOptimalSharedCostAssignment::build_initial_lp
> }
> 
> double LandmarkEfficientOptimalSharedCostAssignment::cost_sharing_h_value(
>
>Yes, https://ai.dmi.unibas.ch/_experiments/ai/downward/issue1036/data/issue1036-base-optimal-eval/issue1036-base-optimal-compare.html has a triple of `*-seq-opt-bjolp-opt` configurations which use this heuristic with CPLEX. Would you like to see some more tests?
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/aibasel/downward/pull/70#discussion_r1257516223
>You are receiving this because you commented.
>
>Message ID: ***@***.***>
-- 
Sent from my phone.

src/search/landmarks/landmark_heuristic.cc
ClemensBuechner: I'm not aware of any differences. Also, I've added the pages which are now published.

src/search/landmarks/landmark_status_manager.cc
salome-eriksson: By automatic conversion, do you mean adding a constructor `ConstBitsetView::ConstBitsetView(const BitsetView &other)`? I just tried that and it works, but it leads to some other ugliness. For starters, this constructor requires `ConstBitsetView` to have access to `BitsetView`'s private members. I solved this with a fried declaration (which required me to forward declare `ConstBitsetView`). But then we need to do the entire thing again for `(Const)ArrayView` since this is one of the members of `(Const)BitsetView.

I can push that change so you can see it better, but I first wanted to ask if this is roughly what you had in mind.

maltehelmert: There are two ways to implement automatic conversions from class A to class B: class B can have a non-explicit constructor from class A, or class A can implement an "operator B". In the latter case, you would have access to the private members of class A. So in this case, BitsetView would implement `ConstBitsetView operator ConstBitsetView() const {return ...;}`. It would probably be a good idea for the implementation to be in the header file so that it is inlinable and there is no performance cost compared to the current code.

maltehelmert: If you want to speed this up, we can look at the code together over Zoom (applies to both of you).

src/search/landmarks/landmark_status_manager.cc
```diff
@@ -71,6 +66,11 @@ LandmarkStatusManager::LandmarkStatusManager(
           progress_reasonable_orderings
           ? get_reasonable_parents(graph)
           : vector<pair<LandmarkNode *, vector<LandmarkNode *>>>{}),
+      /*
+        By default we mark all landmarks past, since it is the neutral element
+        for set intersection which we emulate in the progression. Similarly, no
+        landmarks are future which is the neutral element for set union.
```
maltehelmert, line 72: I think something got mangled at some point in this issue, or perhaps a previous issue (not in this commit). I think these sentences don't make sense any more, the "it" should reference the words "true" and "false", but they are not there any more. It is not clear to me what the "it" is supposed to mean now in these two sentences.

maltehelmert: Suggestion: `We initialize to true in past_landmarks because true is the neutral element of conjunction/set intersection. We initialize to false in future_landmarks because false is neutral for disjunction/set union.`

src/search/landmarks/landmark_status_manager.cc
maltehelmert: @salome-eriksson Clemens and I looked at this together, and Clemens will push a commit.

PR #87, opened 2022-02-02
Issue937

ClemensBuechner:
In unsolvable tasks, it is possible that cycles of only natural (and stronger) orderings occur in the landmark graph. Previously, the implementation did not expect that when calling `mk_acyclic_graph` and failed to remove a reasonable (or weaker) ordering, failing an assertion. Our solution is conformant with the idea to make the graph acyclic (even though this means dropping usefol information, but this flaw lies in making the graph acyclic in the first place) and we just remove the weakest ordering in each cycle, even if it is natural (or stronger). If the cycle consists of only natural (and stronger) orderings, we additionally mark all landmarks involved in the cycle to have empty first achievers, which means the heuristic will detect it cannot achieve any of them, which results in a heuristic estimate in LM-count (and should do the same in every other hypothetical landmark heuristic). This requires that (first) achievers are calculated before calling `mk_acyclic_graph`, because otherwise this finding is overwritten in `calc_achievers`.


src/search/landmarks/landmark_factory.cc
```diff
@@ -148,35 +153,36 @@ void LandmarkFactory::mk_acyclic_graph() {
...
+            from = it2->first;
+            to = next(it2)->first;
+            weakest_edge = edge;
+        }
+        if (weakest_edge == EdgeType::OBEDIENT_REASONABLE) {
+            break;
+        }
+    }
+    if (weakest_edge > EdgeType::REASONABLE) {
+        // Cycle of natural orderings
```
salome-eriksson, line 175: I would elaborate this comment a bit, saying that a cycle of natural orderings implies unsolvability, which we signal by clearing the set of first achievers

src/search/landmarks/landmark_factory_rpg_sasp.cc
```diff
@@ -165,7 +165,7 @@ int LandmarkFactoryRpgSasp::min_cost_for_landmark(
       with "astar(lmcount(lm_rhw()))" on mystery/prob07.pddl in debug mode.
       See issue 467
     */
-    assert(min_cost < numeric_limits<int>::max());
+    //assert(min_cost < numeric_limits<int>::max());
```
salome-eriksson, line 168: This is commented out because on the test instance for this issue this assertion fires (this is due to the unrelated issue467). It would be ideal if we merge issue467 before merging this, but we should at the very least make sure that both issues are merged during this sprint.

src/search/landmarks/landmark_factory.cc
```diff
@@ -148,35 +153,36 @@ void LandmarkFactory::mk_acyclic_graph() {
...
+        for (list<pair<LandmarkNode *, EdgeType>>::iterator it2 = it;
+             it2 != path.end(); ++it2) {
+            it2->first->get_landmark().first_achievers.clear();
+            // TODO: maybe also clear *possible_achievers*?
```
salome-eriksson, line 179: I don't have a strong preference here but I would not clear possible achievers: it doesn't gain us anything and I understand the semantics of possible achievers as "any action that can make the landmark true, regardless of initial state, whether we are on the path of a plan or anything else".

src/search/landmarks/landmark_factory_rpg_sasp.cc
ClemensBuechner: I just commented this back in, because as you've written, this is not part of another issue. It will not influence the experiments as the experiments run in release mode anyway. I don't think it's necessary to wait for issue467 to be fixed first, and I don't think we should wait with merging once we agree this is ready to be merged.

src/search/landmarks/landmark_factory.cc
ClemensBuechner: I agree and removed it.

src/search/landmarks/landmark_factory_rpg_sasp.cc
salome-eriksson: Good point. I'd still try to make sure issuei467 is merged soon after, but since this should easily be doable during this sprint I don't see a problem :)

salome-eriksson (approved):

src/search/landmarks/landmark_factory.cc
```diff
@@ -148,35 +153,44 @@ void LandmarkFactory::mk_acyclic_graph() {
...
+          If the weakest ordering is natural (or stronger), this means
+          the entire cycle consists of only natural (or stronger)
+          orderings which indicates that the task is unsolvable. We
+          signal this by clearing the *first_achievers* of all landmarks
+          in the cycle.
+          Note that *calc_achievers* should never be called after
+          *mk_acyclic_graph* because of this, because otherwise this
+          information is lost.
+        */
+        for (list<pair<LandmarkNode *, EdgeType>>::iterator it2 = it;
```
maltehelmert, line 185: This could probably (and hence) should be a range-based for loop, I think. I guess it's the way it is because it was that way before, just a note for similar things in the future. (In the diff it shows up as "new code", but I guess that's because the diff couldn't match the old to the new perfectly here.)

src/search/landmarks/landmark_factory_relaxation.cc
```diff
@@ -15,10 +15,11 @@ void LandmarkFactoryRelaxation::generate_landmarks(const shared_ptr<AbstractTask
     postprocess(task_proxy, exploration);
 }
 
-void LandmarkFactoryRelaxation::postprocess(const TaskProxy &task_proxy, Exploration &exploration) {
+void LandmarkFactoryRelaxation::postprocess(
+    const TaskProxy &task_proxy, Exploration &exploration) {
     lm_graph->set_landmark_ids();
-    mk_acyclic_graph();
     calc_achievers(task_proxy, exploration);
+    mk_acyclic_graph();
```
maltehelmert, line 22: This kind of ordering dependency is a code smell; best not to have it, or at least have the code notice itself when something is wrong via an assertion or similar. The duplicated comments on this are an indication that there is a DRY violation here. Not a problem for now, just a pointer for the future. When I work on code like this, I generally try to leave a comment near the code smell to indicate that we're aware of the problem. (That is, if I have to copy the same comment to multiple files, I'd mention in the comments themselves that this is the case and that we're aware it's a problem.) But no need to change this at this late stage in the issue.

ClemensBuechner: We discussed a solution like having a flag `achievers_calculated` or something similar and assert this in `mk_acyclic_graph`. I'm also happy to still add this to this pull request. After all, there was no dependency before we worked on this issue, but now there is one, so we should probably aim to make it less smelly as long as we have ideas to do so.

I don't understand your suggestion concerning the comment(s) yet. Do you mean you would have them point to each other, i.e., the comment for `LandmarkFactoryHM::calc_achievers` should point to the (currently identical) comment for `LandmarkFactoryRelaxation::calc_achievers`? Or to the comment for `mk_acyclic_graph`? Or should it just point out that we are aware that this is not an ideal solution, not pointing to any other code?

maltehelmert: A flag could be a good idea, but if you prefer to merge with no further changes, that's fine too. After adding a flag, I'd do a few tests locally in debug mode with the relevant configurations to verify that the assertions don't trigger. Regarding the comments: there were two identical comments "Achievers should be calculated..." that can go away once there is an assertion, or be shortened to a single line next to the assertion once we have it. The smell is to have the same comment in multiple places. If that stays, I would change the comments themselves to add something like "DRY: We have the same comment also in XXX; perhaps some refactoring is needed."

maltehelmert: Depends on the cost/benefit and really requires a case by case decision. I said "could be a good idea" rather than something stronger because I think one can argue either way. The cost in time and space is miniscule and not worth worrying about. The cost in code complexity is hard to judge ahead of time. There could also be a benefit in code complexity because I think we can then remove or drastically shorten several comments. There is also a benefit in preventing mistakes and in explicitly highlighting a design flaw; the very fact that a class does require a certain protocol "first you have to do this, then that" that isn't automatically enforced by things like control flow or type system is generally a sign of bad design. Having this very clear in the code is generally a good thing.

PR #88, opened 2022-02-03
Issue1041

ClemensBuechner:


src/search/landmarks/exploration.cc
```diff
@@ -124,11 +124,19 @@ void Exploration::build_unary_operators(const OperatorProxy &op) {
...
-                                          bool use_h_max) {
+/*
+  This function initializes the first layer of the relaxed planning task and the
+  priority queue for unrolling the relaxed task graph. Unary operators are
+  excluded if they achieve an excluded proposition or their operator ID is in
+  *excluded_op_ids*.
+
+  *excluded_op_ids* should contain the operators that achieve an excluded
+  proposition unconditionally. Currently, the only place where this is used is
+  in *LandmarkFactoryRelaxation::relaxed_task_solvable()*.
```
maltehelmert, line 135: From our Zoom discussion, we think that the last sentence is not correct.

src/search/landmarks/exploration.cc
```diff
@@ -124,11 +124,19 @@ void Exploration::build_unary_operators(const OperatorProxy &op) {
...
 }
 
-// heuristic computation
-void Exploration::setup_exploration_queue(const State &state,
-                                          const vector<FactPair> &excluded_props,
-                                          const unordered_set<int> &excluded_op_ids,
-                                          bool use_h_max) {
+/*
+  This function initializes the first layer of the relaxed planning task and the
+  priority queue for unrolling the relaxed task graph. Unary operators are
```
maltehelmert, line 129: I would change the first sentence to:

This function initializes the priority queue and the information associated with propositions and unary operators for the relaxed exploration. Unary operators that are not allowed to be applied due to exclusions are marked by setting their h_add_cost to -2 (sigh).

maltehelmert: Similarly, the h_add_cost of excluded unary propositions is set to -2.

CHANGES.md
```diff
@@ -133,6 +133,9 @@ after the corresponding tracker issues.
   add_log_options_to_parser and get_log_from_options to obtain their
   local log object.
 
+- Fix a bug in checking relaxed solvability when excluding landmarks in the
+  presence of conditional effects.
+  <https://issues.fast-downward.org/issue1041>
```
maltehelmert, line 138: I commented on this on Discord. I guess I should have commented here. :-)

src/search/landmarks/exploration.cc
```diff
@@ -124,11 +124,30 @@ void Exploration::build_unary_operators(const OperatorProxy &op) {
...
+  *compute_reachability_with_excludes()* (and hence this function) is used:
+  (a) in *LandmarkFactoryRelaxation::relaxed_task_solvable()* where indeed
+      all operators are excluded if they achieve an excluded proposition
+      unconditionally; and
+  (b) in *LandmarkFactoryRelaxation::is_causal_landmark()* where
+      *excluded_props* is empty and hence no operators achieve an excluded
+      proposition. (*excluded_op_ids* "additionally" contains operators that
+      have a landmark as their precondition in this context.)
+
+  TODO: Issue 1045 aims at moving the logic to exclude operators that achieve
```
maltehelmert, line 145: Can you write "issue1045" instead of "Issue 1045"? That's our general convention, and I often grep for things like these to find TODOs like these. I don't think it's a problem to start a sentence with a lower-case letter in cases like these. We do the same in the previous paragraph.

CHANGES.md
ClemensBuechner: I tried again with a more high-level and user-oriented changelog entry. Does this go in the direction you were thinking? Otherwise, I may have still not understood what you're aiming at and it would be easiest if you come up with a proposal that I can just insert here.

maltehelmert: Much improved, thanks!

PR #89, opened 2022-02-04
[issue1042] add LimitedPruning class and move statistics to base class

silvansievers:


src/search/pruning/limited_pruning.cc
```diff
@@ -0,0 +1,78 @@
...
+    utils::g_log << "pruning method: limited" << endl;
+}
+
+void LimitedPruning::prune_operators(
+    const State &state, vector<OperatorID> &op_ids) {
+    if (is_pruning_disabled) {
+        return;
+    }
+    if (min_required_pruning_ratio > 0. &&
+        num_pruning_calls == num_expansions_before_checking_pruning_ratio) {
```
salome-eriksson, line 31: Is the intention to only check this once during the entire search, not iteratively after x*num_expansions_before_checking_pruning_ratio?
If so, might it make sense to set min_required_pruning_ratio to 0 inside the if-statement to make the if return false faster in the future? (Genuine question, I don't have an intuition.)

src/search/pruning/limited_pruning.cc
```diff
@@ -0,0 +1,78 @@
...
+    ++num_pruning_calls;
+    pruning_method->prune_operators(state, op_ids);
+}
+
+static shared_ptr<PruningMethod> _parse(OptionParser &parser) {
+    parser.document_synopsis(
+        "Limited pruning",
+        "Limited pruning applies another pruning method and switches it off "
+        "after a fixed number of expansions if the pruning ratio is below a "
+        "given value. The pruning ratio is defined as 1 - (A/B), where B and A "
```
salome-eriksson, line 53: While this is how it is calculated in the code, I think it would be more straightforward to say something like:
"The pruning ratio is the sum of all pruned operators divided by the sum of all operators before pruning, considering all previous expansions."

src/search/pruning_method.h
```diff
@@ -2,29 +2,35 @@
...
+    utils::Timer timer;
+    friend class limited_pruning::LimitedPruning;
+
+    virtual void prune_operators(
+        const State &state, std::vector<OperatorID> &op_ids) = 0;
 protected:
     std::shared_ptr<AbstractTask> task;
-
+    long num_unpruned_successors_generated;
+    long num_pruned_successors_generated;
```
salome-eriksson, line 27: I find these two names a bit unfortunate and misunderstood them at first:
 - `num_unpruned_successors_generated` sounds like the number of successors that were not pruned
 -  `num_pruned_successors_generated`sounds like these were the successors that were pruned, but why where they generated then?
 
I would rather suggest `num_successors_before_pruning` and `ǹum_successors_after_pruning`

src/search/pruning_method.cc
```diff
@@ -13,6 +16,31 @@ PruningMethod::PruningMethod()
...
+    timer.resume();
+    int num_ops_before_pruning = op_ids.size();
+    prune_operators(state, op_ids);
+    num_unpruned_successors_generated += num_ops_before_pruning;
+    num_pruned_successors_generated += op_ids.size();
+    timer.stop();
+}
+
+void PruningMethod::print_statistics() const {
+    utils::g_log << "total successors before partial-order reduction: "
```
salome-eriksson, line 35: Are all pruning methods partial-order reductions? Even if this is (currently) the case, what speaks against simply saying "total successors before/after pruning"?

src/search/pruning_method.cc
```diff
@@ -13,6 +16,31 @@ PruningMethod::PruningMethod()
 void PruningMethod::initialize(const shared_ptr<AbstractTask> &task_) {
     assert(!task);
     task = task_;
+    num_unpruned_successors_generated = 0;
+    num_pruned_successors_generated = 0;
+}
+
+void PruningMethod::prune_op_ids(
+    const State &state, vector<OperatorID> &op_ids) {
+    assert(!task_properties::is_goal_state(TaskProxy(*task), state));
```
salome-eriksson, line 25: It is not clear to me why we should not call this method in goal states, so maybe a comment would be good here. It's not necessary though (and I like it a lot that we don't delegate this to the derived classes anymore! :))

src/search/pruning_method.h
```diff
@@ -2,29 +2,35 @@
...
 class AbstractTask;
+class State;
+
+namespace limited_pruning {
+class LimitedPruning;
+}
 
 class PruningMethod {
+    utils::Timer timer;
+    friend class limited_pruning::LimitedPruning;
```
salome-eriksson, line 20: Why does LimitedPruning need to be a friend class?

src/search/pruning/null_pruning_method.h
```diff
@@ -9,7 +9,6 @@ class NullPruningMethod : public PruningMethod {
     virtual void initialize(const std::shared_ptr<AbstractTask> &) override;
     virtual void prune_operators(const State &,
```
salome-eriksson, line 10: Shouldn't `prune_operators` be moved outside public?

src/search/pruning_method.cc
```diff
@@ -13,6 +16,31 @@ PruningMethod::PruningMethod()
 void PruningMethod::initialize(const shared_ptr<AbstractTask> &task_) {
     assert(!task);
     task = task_;
+    num_unpruned_successors_generated = 0;
+    num_pruned_successors_generated = 0;
+}
```
roeger, line 21: If we use `LimitedPruning`, we keep these three variables for the wrapper as well as for actual pruning method but we count only in `LimitedPruning`. Looks like this is the downside of the new design. Any ideas how to avoid it (without making limitation an option again)?

src/search/pruning_method.h
```diff
@@ -2,29 +2,35 @@
...
-
     virtual void initialize(const std::shared_ptr<AbstractTask> &task);
-
-    /* This method must not be called for goal states. This can be checked
-       with assertions in derived classes. */
-    virtual void prune_operators(const State &state,
-                                 std::vector<OperatorID> &op_ids) = 0;
-
-    virtual void print_statistics() const = 0;
+    void prune_op_ids(const State &state, std::vector<OperatorID> &op_ids);
```
roeger, line 32: Hahaha, nice trick to free the name `prune_operators` for the actual internal pruning. But a much uglier name in the public interface. Is there another reason to rename it (e.g. to align it with other names)? Otherwise I would rather keep the public name and find a more precise one for the "internal" method.

roeger (commented):
I looked at the changes and left some comments.

src/search/pruning/limited_pruning.cc
silvansievers: Yes, this is indeed the intention. I wouldn't expect it to cause a large overhead. I would not want to change `min_required_pruning_ratio` because that is logically a const variable (and it is actually const). However, to still do something about this: what do you think about ordering `num_pruning_calls == num_expansions_before_checking_pruning_ratio` first? This evaluates to true exactly once.

src/search/pruning_method.h
silvansievers: Good catch; it doesn't need to be.

src/search/pruning_method.cc
silvansievers: To be honest, I don't know. Maybe something specific to stubborn sets? Maybe because the list of operators are empty for goal states? Although I don't see why this should cause problems.

src/search/pruning_method.h
silvansievers: Wait, LimitedPruning was not compiled with the wrong `DEPENDENCY_ONLY` flag noticed by Gabi. Since it directly wants to use the private member `prune_operators` of pruning methods, LImitedPruning has to be a friend class.

src/search/pruning/null_pruning_method.h
```diff
@@ -9,7 +9,6 @@ class NullPruningMethod : public PruningMethod {
...
                                  std::vector<OperatorID> &) override {}
-    virtual void print_statistics() const override {}
```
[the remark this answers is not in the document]
silvansievers: You are right. What do you think is the better way of addressing this: making `print_statistics` `virtual` so that the null pruning method can override it or not printing any pruning statistics if there was no pruning? From an experiments perspective, I would prefer to always have the same output, also if there was no pruning with a non-null pruning method. But overriding to suppress doesn't seem that great either.

src/search/pruning_method.cc
silvansievers: Good point. I don't have any ideas to avoid this problem.

src/search/pruning_method.h
silvansievers: No, just the reason you found. Any suggestion for the name of the internal method?

src/search/pruning/limited_pruning.cc
salome-eriksson: Ah I missed that it was const, good point. I think we can also leave it as is, especially if you planned on not running experiments ;).

src/search/pruning_method.cc
salome-eriksson: Maybe it's also that we in general shouldn't do anything with goal states. But as I said, an explanatory comment isn't really necessary anyway, so I'll just mark this as resolved.

src/search/pruning/stubborn_sets.h
```diff
@@ -94,16 +81,9 @@ class StubbornSets : public PruningMethod {
     virtual void initialize_stubborn_set(const State &state) = 0;
     virtual void handle_stubborn_operator(const State &state, int op_no) = 0;
 public:
-    explicit StubbornSets(const options::Options &opts);
+    StubbornSets();
```
maltehelmert, line 84: Not needed any more?

src/search/pruning/stubborn_sets.cc
```diff
@@ -32,13 +31,7 @@ bool contain_conflicting_fact(const vector<FactPair> &facts1,
...
 }
 
-StubbornSets::StubbornSets(const options::Options &opts)
-    : min_required_pruning_ratio(opts.get<double>("min_required_pruning_ratio")),
-      num_expansions_before_checking_pruning_ratio(
-          opts.get<int>("expansions_before_checking_pruning_ratio")),
-      num_pruning_calls(0),
-      is_pruning_disabled(false),
-      timer(false) {
+StubbornSets::StubbornSets() : num_operators(0) {
```
maltehelmert, line 34: Is this the best value? I see this was previously uninitialized. Rather than knowing we have 0 operators, it rather seems to be the case that we have an uninitialized attribute here. Perhaps an illegal value like -1 would communicate this more clearly.

src/search/pruning/null_pruning_method.h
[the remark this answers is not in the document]
FlorianPommerening: Do we need the `NullPruningMethod` at all? It has been a long time since we implemented this and I don't quite remember why we added it. But working on the option parser now, I saw there is an option `OptionParser::NONE` that can be used as a default value for optional arguments without a default value. It could be that we just added `NullPruningMethod` because we were not aware of this option and did not know how to specify a default value for the option. The downside of using `NONE` is that within the search, the pruning method would then potentially `nullptr` and you'd have to add checks for that. Silvan, I saw this was used for label reduction in M&S, so you probably have experience with the trade-off.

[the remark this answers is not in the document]
silvansievers: Good point. I think that the search calls the pruning method exactly once, so this would require a single additional if-clause. Definitely good trade-off.

src/search/pruning/stubborn_sets.h
silvansievers: Depending on if we want to manually initialize `num_operators` or not.

src/search/search_engines/eager_search.cc
```diff
@@ -20,6 +20,13 @@
 using namespace std;
 
 namespace eager_search {
+static shared_ptr<PruningMethod> get_pruning_method_from_options(const Options &opts) {
+    if (opts.contains("pruning")) {
+        return opts.get<shared_ptr<PruningMethod>>("pruning");
+    }
+    return nullptr;
+}
+
```
jendrikseipp, line 29: A shorthand for this is `return opts.get<shared_ptr<PruningMethod>>("pruning", nullptr);`. Not sure if this needs a separate function `get_pruning_method_from_options` then.

silvansievers: Ah, nice, I didn't know the get method could take a default value!

CHANGES.md
```diff
@@ -125,6 +125,14 @@ after the corresponding tracker issues.
...
   <http://issues.fast-downward.org/issue1031>
 
+- For users: new LimitedPruning class replaces previous limitation
+  options of individual pruning methods
+  <http://issues.fast-downward.org/issue1042>
+  A previous command line option using this feature, such as
+  --search "astar(lmcut(),pruning=atom_centric_stubborn_sets(min_required_pruning_ratio=0.2,expansions_before_checking_pruning_ratio=1000))"
+  is now changed to
+  --search "astar(lmcut(),pruning=limited_pruning(pruning=atom_centric_stubborn_sets,min_required_pruning_ratio=0.2,expansions_before_checking_pruning_ratio=1000))
+
```
maltehelmert, line 135: I think this should be `pruning=atom_centric_stubborn_sets()`. The fact that the parentheses are currently optional is an option parser bug and shouldn't work anymore after the option parser rewrite. (Unless those involved in the rewrite thought it was a feature and explicitly added support. ;-) But I think we shouldn't.) I wouldn't write "for users"; see https://issues.fast-downward.org/msg10563.

src/search/pruning/limited_pruning.cc
```diff
@@ -0,0 +1,78 @@
...
+namespace limited_pruning {
+LimitedPruning::LimitedPruning(const Options &opts)
+    : pruning_method(opts.get<shared_ptr<PruningMethod>>("pruning")),
+      min_required_pruning_ratio(opts.get<double>("min_required_pruning_ratio")),
+      num_expansions_before_checking_pruning_ratio(
+          opts.get<int>("expansions_before_checking_pruning_ratio")),
+      num_pruning_calls(0),
+      is_pruning_disabled(false) {
+}
+void LimitedPruning::initialize(const shared_ptr<AbstractTask> &task) {
```
maltehelmert, line 19: Should have an empty line between methods.

src/search/pruning/stubborn_sets.cc
```diff
@@ -110,30 +101,8 @@ bool StubbornSets::mark_as_stubborn(int op_no) {
     return false;
 }
 
-void StubbornSets::prune_operators(
+void StubbornSets::prune(
```
maltehelmert, line 104: Looks like a short line to have a line break, but perhaps I'm misjudging.

src/search/pruning/stubborn_sets.h
```diff
@@ -1,29 +1,15 @@
...
 
-#include "../abstract_task.h"
 #include "../pruning_method.h"
 
-#include "../utils/timer.h"
-
-namespace options {
-class OptionParser;
-}
+#include "../task_proxy.h"
```
maltehelmert, line 6: Shouldn't have a blank line between the includes. The previous blank line was because of the different subdirectories involved.

src/search/pruning/stubborn_sets.cc
silvansievers: It would be too long otherwise due to the `override`.

PR #90, opened 2022-02-07
Issue467

ClemensBuechner:


src/search/landmarks/landmark.h
```diff
@@ -31,7 +31,10 @@ class Landmark {
     bool is_true_in_goal;
     bool is_derived;
 
-    // Cost of achieving the landmark (as determined by the landmark factory)
+    /*
+      Cost of achieving the landmark (as determined by the landmark factory),
+      can be infinite (currently only if *first_achievers* is empty).
```
maltehelmert, line 36: Can you explain what "infinite" means? Integers don't have a notion of infinite as such; we have different conventions in different parts of the code which values to use as a substitute (-1 vs. the maximum int value).

CHANGES.md
```diff
@@ -125,6 +125,9 @@ after the corresponding tracker issues.
   override this choice.
   <http://issues.fast-downward.org/issue1031>
 
+- Fix a bug in the computation of landmark costs
+  <http://issues.fast-downward.org/issue467>
```
maltehelmert, line 129: Similar to my comment for the other issue, we should write these from the perspective of what the users need to know about this change. Landmark costs are a code-internal concept. We don't want to scare everyone that uses landmarks in any way that their past experiment results are now invalidated.

Assuming that the issue title is complete, things we should consider mentioning include: this only affects the lmcount heuristic (not, for example, LM-Cut, or operator-counting constraints using landmarks), only a certain landmark factory (which implies lmcount, so perhaps making the previous point unnecessary), only unsolvable task, and the mode of failure was a failed assertion. (Then people know that if they didn't work with this factory or with unsolvable tasks and if they didn't see a failed assertion, then they were not affected.)

Of course it's always good to be skeptical about issue titles and consider what other failure modes might have existed, e.g. on build were assertions were disabled.

The basic idea is: include enough information for users to make an informed judgement whether this issue is relevant to them, how the changed behaviour affects them, and whether they need to dig into the details on the tracker.

PR #91, opened 2022-02-07
Issue1043

silvansievers:


src/search/pdbs/pattern_collection_generator_hillclimbing.cc
```diff
@@ -253,8 +254,10 @@ pair<int, int> PatternCollectionGeneratorHillclimbing::find_best_improving_pdb(
             best_pdb_index = i;
         }
         if (count > 0) {
```
PatrickFerber, line 256: merge both conditions to one. (count > 0 && verbosity >= ...)

src/search/pdbs/pattern_collection_generator_hillclimbing.cc
```diff
@@ -342,14 +349,23 @@ void PatternCollectionGeneratorHillclimbing::hill_climbing(
         while (true) {
             ++num_iterations;
             int init_h = current_pdbs->get_value(initial_state);
-            utils::g_log << "current collection size is "
-                         << current_pdbs->get_size() << endl;
-            utils::g_log << "current initial h value: ";
+            if (verbosity >= utils::Verbosity::VERBOSE) {
+                utils::g_log << "current collection size is "
+                             << current_pdbs->get_size() << endl;
+                utils::g_log << "current initial h value: ";
```
PatrickFerber, line 355: no need to repeat the second utils::g_log

PatrickFerber: how expensive is the is_dead_end(state) check? 
If it is cheap, the code would be cleaner, if you directly chain:
` << (pdbs->is_dead_end(state)? "inftiy": iinit_h << endl;`
Then, the next if block is simplified and the else block can be removed.

src/search/pdbs/pattern_collection_generator_hillclimbing.cc
```diff
@@ -389,30 +411,39 @@ void PatternCollectionGeneratorHillclimbing::hill_climbing(
...
-    utils::g_log << "Hill climbing maximum PDB size: " << max_pdb_size << endl;
-    utils::g_log << "Hill climbing time: "
-                 << hill_climbing_timer->get_elapsed_time() << endl;
+    if (verbosity >= utils::Verbosity::NORMAL) {
+        utils::g_log << "Hill climbing iterations: " << num_iterations << endl;
+        utils::g_log << "Hill climbing generated patterns: " << generated_patterns.size() << endl;
+        utils::g_log << "Hill climbing rejected patterns: " << num_rejected << endl;
+        utils::g_log << "Hill climbing maximum PDB size: " << max_pdb_size << endl;
+        utils::g_log << "Hill climbing time: "
+                     << hill_climbing_timer->get_elapsed_time() << endl;
```
PatrickFerber, line 432: I prefer chaining << over repeating utils::g_log

src/search/pdbs/pattern_generator.h
```diff
@@ -6,24 +6,49 @@
 #include "types.h"
 
 #include <memory>
```
PatrickFerber, line 8: Might not be related to this issue, but as we take a look currently at those generators:

The code for path the Pattern and the PatternCollection generators is almost a perfect copy. If this is expected to stay, the code can be simplified using templates.

1. create a templated generator : `template<typename T> Generator{...; virtual T compute_patterns(..) = 0;
2. class PatternGenerator : public Generator<PatternInformation>{};
3. class PatternCollectionGenerator : public Generator<PatternCollectionInformation>{};

The code in the cc files is also an almost perfect copy. The only difference are the dump_pattern(_collection)_statistics, but this can be easily done with overloading.

Is there a reason to have those duplication?

src/search/pdbs/pattern_generator_manual.cc
```diff
@@ -14,10 +14,13 @@ using namespace std;
...
+}
+string PatternGeneratorManual::name() const {
+    return "manual pattern generator";
 }
 
-PatternInformation PatternGeneratorManual::generate(
+PatternInformation PatternGeneratorManual::compute_pattern(
     const shared_ptr<AbstractTask> &task) {
     PatternInformation pattern_info(TaskProxy(*task), move(pattern));
     utils::g_log << "Manual pattern: " << pattern_info.get_pattern() << endl;
```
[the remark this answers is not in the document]
silvansievers: Good catch, yes!

src/search/pdbs/pattern_collection_generator_hillclimbing.cc
silvansievers: It is an iteration over all patterns/PDBs and a lookup for each. I could simply store a boolean to avoid recomputation and simplify the output as suggested.

src/search/pdbs/pattern_collection_generator_hillclimbing.cc
silvansievers: I prefer repeating the stream after each endl, i.e., I dislike chaining over `endl`. I couldn't find anything regarding this in our coding conventions, but I think I have never seen something like `cout << "bla" << endl << "foo" << endl` in Fast Downward.

src/search/pdbs/pattern_generator.h
silvansievers: I don't know if there is a good reason other than not wanting to use templates. I'm somewhat indifferent on this, maybe someone else has an opinion. Can you ask on Discord?

src/search/pdbs/cegar.cc
```diff
@@ -197,24 +197,24 @@ CEGAR::CEGAR(
...
-    utils::g_log << "]" << endl;
+    log << "]" << endl;
 }
 
 bool CEGAR::time_limit_reached(
     const utils::CountdownTimer &timer) const {
     if (timer.is_expired()) {
-        if (verbosity >= utils::Verbosity::NORMAL) {
-            utils::g_log << "CEGAR time limit reached" << endl;
+        if (log.is_at_least_normal()) {
```
maltehelmert, line 216: We already had the bike-shedding discussion about these names, right?
It seems a long name, but I don't have an awesome alternative. And it's not what this issue is about. I'm just interested.

src/search/pdbs/cegar.cc
```diff
@@ -521,78 +521,63 @@ void CEGAR::refine(const FlawList &flaws) {
...
-        }
-        utils::g_log << endl;
-        utils::g_log << "goal variables: ";
+    if (log.is_at_least_normal()) {
+        log << "CEGAR options:" << endl;
+        log << "max pdb size: " << max_pdb_size << endl;
+        log << "max collection size: " << max_collection_size << endl;
+        log << "max time: " << max_time << endl;
+        log << "wildcard plans: " << use_wildcard_plans << endl;
+        log.dump_options();
```
maltehelmert, line 566: No need to change this, but I'm not sure if we actually want this here going forward. This might end up being a complex log configuration, possibly for a log shared with other components of the planner although we don't currently support that, and it's not clear to me whether we then want to print a complex log configuration in a place like this. Just something to potentially consider in the future as the log functionality evolves further.

src/search/pdbs/pattern_collection_generator_multiple.cc
```diff
@@ -132,59 +131,46 @@ bool PatternCollectionGeneratorMultiple::check_for_stagnation(
...
+    if (log.is_at_least_normal()) {
+        log << "max pdb size: " << max_pdb_size << endl;
+        log << "max collection size: " << remaining_collection_size << endl;
+        log << "max time: " << total_max_time << endl;
+        log << "stagnation time limit: " << stagnation_limit << endl;
+        log << "timer after which blacklisting is enabled: "
+            << blacklisting_start_time << endl;
+        log << "enable blacklisting after stagnation: "
+            << enable_blacklist_on_stagnation << endl;
+        log.dump_options();
```
maltehelmert, line 173: (Same comment applies in all these places.)

src/search/pdbs/pattern_collection_generator_systematic.cc
```diff
@@ -258,13 +263,16 @@ void PatternCollectionGeneratorSystematic::build_patterns_naive(
         next_patterns.clear();
     }
 
-    utils::g_log << "Found " << patterns->size() << " patterns." << endl;
+    if (log.is_at_least_normal()) {
+        log << "Found " << patterns->size() << " patterns." << endl;
+    }
+}
```
maltehelmert, line 269: Missing new line between methods.

src/search/pdbs/pattern_generator_manual.cc
```diff
@@ -14,13 +14,18 @@ using namespace std;
 
 namespace pdbs {
 PatternGeneratorManual::PatternGeneratorManual(const Options &opts)
-    : pattern(opts.get_list<int>("pattern")) {
+    : PatternGenerator(opts), pattern(opts.get_list<int>("pattern")) {
+}
```
maltehelmert, line 18: missing blank line

PR #96, opened 2022-02-14
Issue921

silvansievers:


src/search/cegar/abstraction.cc
```diff
@@ -137,8 +137,8 @@ pair<int, int> Abstraction::refine(
 }
 
 void Abstraction::print_statistics() const {
```
PatrickFerber, line 139: always log this or add an llog.is_at_least_normal?

src/search/cegar/cegar.cc
```diff
@@ -270,7 +289,7 @@ unique_ptr<Flaw> CEGAR::find_flaw(const Solution &solution) {
 void CEGAR::print_statistics() {
     abstraction->print_statistics();
     int init_id = abstraction->get_initial_state().get_id();
```
PatrickFerber, line 291: print always?

src/search/cegar/cost_saturation.cc
```diff
@@ -241,13 +241,13 @@ void CostSaturation::build_abstractions(
...
 void CostSaturation::print_statistics(utils::Duration init_time) const {
-    utils::g_log << "Done initializing additive Cartesian heuristic" << endl;
-    utils::g_log << "Time for initializing additive Cartesian heuristic: "
-                 << init_time << endl;
-    utils::g_log << "Cartesian abstractions built: " << num_abstractions << endl;
-    utils::g_log << "Cartesian states: " << num_states << endl;
-    utils::g_log << "Total number of non-looping transitions: "
-                 << num_non_looping_transitions << endl;
-    utils::g_log << endl;
+    log << "Done initializing additive Cartesian heuristic" << endl;
```
PatrickFerber, line 244: Across all files, decided if we want to have a condition on the `print_statistics` output like at least normal.

src/search/landmarks/landmark_count_heuristic.cc
```diff
@@ -37,8 +37,11 @@ LandmarkCountHeuristic::LandmarkCountHeuristic(const options::Options &opts)
           admissible ||
           (!task_properties::has_axioms(task_proxy) &&
            (!task_properties::has_conditional_effects(task_proxy) || conditional_effects_supported))),
+      log(utils::get_log_from_options(opts)),
```
PatrickFerber, line 40: I saw in every heuristic that the log option is added to its parsing function and to its constructor. Should we move this just to the base class `Heuristic` or even `Evaluator`?

src/search/landmarks/landmark_factory_h_m.cc
```diff
@@ -350,70 +352,70 @@ void LandmarkFactoryHM::print_pm_op(const VariablesProxy &variables, const PMOp
         cond_eff.clear();
         int pm_fluent;
         size_t j;
```
PatrickFerber, line 354: Do I set it correctly that this function is only called within a "log.at_least_X" condition? If yes that is good, otherwise, I am wondering if we always want to log this.

PatrickFerber (commented):
Looks good. There are a few questions which might lead to no code changes :)

src/search/cegar/abstraction.cc
silvansievers: Currently, this is only called from `cegar.cc`, where it is ensured that logging is at least normal if calling the method. Are you in favor of always having that if-clause?

src/search/cegar/cegar.cc
silvansievers: see previous reply

src/search/landmarks/landmark_count_heuristic.cc
silvansievers: Yes, I think having the log in the `Heuristic` class should be the goal. But I would only do it once all heuristics actually use a local log instead of `g_log`. I don't think `Evaluator` should have a log. None of the existing evaluators produce output themselves. Only the base class has two `print` methods which are called by the search with the log passed as a parameter by the search. I think this makes sense because search can decide on its own if to call this method and how to configure the log for that.

src/search/landmarks/landmark_factory_h_m.cc
silvansievers: This is basically the same thing again: do we want to if-guard all `print_*` methods or do we say that whoever calls such a method should know that this method will produce output. I think this is a question we could discuss with the others.

src/search/cegar/abstraction.cc
PatrickFerber: For other functions I observed what you described (that the if is called around the function call). There I did not write a comment, because it was clear that the purpose of the function is just to pull the output generation away from the meaty parts of the function. To be honest, the function is called `print_statistics` I am fine with defining that if I call this function, then the statistics are printed, independent of the logging level.

jendrikseipp: I agree that checking the log level is the responsibility of the caller.

src/search/landmarks/landmark_count_heuristic.cc
maltehelmert: The distinction "heuristic" vs. "evaluator" is not really that meaningful any more. The names don't match what the difference between them is. The class "Heuristic" now is the class "Evaluator" with some extra bits and pieces that don't really belong together and that are not that closely related to an intuitive distinction between what a heuristic is vs. what a generic evaluator is. We did a lot of refactoring there, but didn't yet complete it. In the meantime, I would avoid adding more distinctions between "Evaluator" and "Heuristic". If the solution for now is to put all logging into the individual heuristics, that works for me.

silvansievers: In that case I would prefer having the log in the Evaluator base class. Adding one to each heuristic would be very tedious.

maltehelmert: Works for me.

src/search/pdbs/utils.cc
```diff
@@ -88,40 +88,48 @@ vector<vector<int>> compute_cg_neighbors(
...
-                 << endl;
-    utils::g_log << identifier << " total PDB size: ";
-    utils::g_log << compute_total_pdb_size(
-        pci.get_task_proxy(), pattern_collection) << endl;
-    utils::g_log << identifier << " computation time: " << runtime << endl;
+    if (log.is_at_least_normal()) {
+        log << identifier << " number of patterns: " << pattern_collection.size()
+            << endl;
+        log << identifier << " total PDB size: ";
+        log << compute_total_pdb_size(
```
ClemensBuechner, line 129: To be more consistent, you could "append" this to the previous line.

src/search/pdbs/validation.cc
```diff
@@ -40,18 +41,19 @@ void validate_and_normalize_pattern(const TaskProxy &task_proxy,
...
     */
     for (Pattern &pattern : patterns)
-        validate_and_normalize_pattern(task_proxy, pattern);
+        validate_and_normalize_pattern(task_proxy, pattern, log);
     PatternCollection sorted_patterns(patterns);
     sort(sorted_patterns.begin(), sorted_patterns.end());
     auto it = unique(sorted_patterns.begin(), sorted_patterns.end());
     if (it != sorted_patterns.end()) {
-        utils::g_log << "Warning: duplicate patterns have been detected" << endl;
+        log << "Warning: duplicate patterns have been detected" << endl;
```
ClemensBuechner, line 56: Use log-level for warnings and to be more consistent over the entire codebase write WARNING in all caps (or adapt everywhere else).

src/search/search_engine.cc
```diff
@@ -173,11 +173,11 @@ void SearchEngine::add_succ_order_options(OptionParser &parser) {
 }
 
 void print_initial_evaluator_values(
-    const EvaluationContext &eval_context, utils::LogProxy &log) {
+    const EvaluationContext &eval_context) {
```
ClemensBuechner, line 176: With the decision to have an if-guard at the lowest level (where prints actually happen), we should probably re-insert handing the logger to this function and also to the ones it calles that use the logger as well.

src/search/search_engines/search_common.cc
```diff
@@ -80,16 +80,28 @@ shared_ptr<OpenListFactory> create_greedy_open_list_factory(
   If w = 0, we omit the h-evaluator altogether:
   we use g instead of g + 0 * h.
 */
-static shared_ptr<Evaluator> create_wastar_eval(const shared_ptr<GEval> &g_eval, int w,
+static shared_ptr<Evaluator> create_wastar_eval(const Options &options,
+                                                const shared_ptr<GEval> &g_eval, int w,
                                                 const shared_ptr<Evaluator> &h_eval) {
     if (w == 0)
```
ClemensBuechner, line 86: After adding curly braces below for the cases `w == 1` and `else`, I personally would prefer to have curly braces here as well. (Though, I always prefer to have them, so it's fine if you decide against it.)

src/search/search_engines/search_common.cc
```diff
@@ -80,16 +80,28 @@ shared_ptr<OpenListFactory> create_greedy_open_list_factory(
...
         return g_eval;
     shared_ptr<Evaluator> w_h_eval = nullptr;
-    if (w == 1)
+    if (w == 1) {
         w_h_eval = h_eval;
-    else
-        w_h_eval = make_shared<WeightedEval>(h_eval, w);
-    return make_shared<SumEval>(vector<shared_ptr<Evaluator>>({g_eval, w_h_eval}));
+    } else {
```
ClemensBuechner, line 91: This entire `else` case seems a bit tedious, I definitely preferred the way it was before for this particular piece of code. Unfortunately, I don't see a better solution for now...

ClemensBuechner (commented):
I have found many little things but nothing too critical. There are many comments talking about opportunities to make the codebase more consistent with regard to the way we write log messages. I don't insist on changing these, but to me, this issue sounds like a good fit for things like that. 
I am in favor of introducing another log-level `warning`, so this is another thing I would change while we're at it.

src/search/heuristics/cg_cache.cc
```diff
@@ -64,7 +66,9 @@ CGCache::CGCache(const TaskProxy &task_proxy, int max_cache_size)
         }
     }
 
-    utils::g_log << "done!" << endl;
+    if (log.is_at_least_normal()) {
```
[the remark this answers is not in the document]
maltehelmert: The point is/was that we flush earlier, and then this line can take a lot of time. This doesn't really work so well with our now more line-based logging; for example, it's not really that clear what the time stamp that every message has refers to (the first or last half of the line?), and I'm not even sure our implementation currently supports flushing in the intended way. (But probably yes.)

src/search/heuristics/hm_heuristic.cc
```diff
@@ -18,10 +18,12 @@ HMHeuristic::HMHeuristic(const Options &opts)
       m(opts.get<int>("m")),
       has_cond_effects(task_properties::has_conditional_effects(task_proxy)),
       goals(task_properties::get_fact_pairs(task_proxy.get_goals())) {
-    utils::g_log << "Using h^" << m << "." << endl;
-    utils::g_log << "The implementation of the h^m heuristic is preliminary." << endl
-                 << "It is SLOOOOOOOOOOOW." << endl
-                 << "Please do not use this for comparison!" << endl;
+    if (log.is_at_least_normal()) {
+        log << "Using h^" << m << "." << endl;
+        log << "The implementation of the h^m heuristic is preliminary." << endl
```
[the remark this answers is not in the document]
ClemensBuechner: OK, that makes sense indeed. Would it be an option to replace the `endl` of the second message (except the last one) by `\n` to (1) only flush this message once and (2) make your interpretation more obvious?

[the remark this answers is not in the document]
ClemensBuechner: I see, thanks for the explanation. Let's just leave it as is, then.

src/search/cegar/split_selector.cc
```diff
@@ -102,7 +102,7 @@ double SplitSelector::rate_split(const AbstractState &state, const Split &split)
         rating = get_max_hadd_value(var_id, values);
         break;
     default:
-        utils::g_log << "Invalid pick strategy: " << static_cast<int>(pick) << endl;
+        cerr << "Invalid pick strategy: " << static_cast<int>(pick) << endl;
```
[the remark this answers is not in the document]
silvansievers: Correct.

src/search/evaluators/combining_evaluator.cc
```diff
@@ -2,13 +2,15 @@
...
 
 using namespace std;
 
 namespace combining_evaluator {
 CombiningEvaluator::CombiningEvaluator(
-    const vector<shared_ptr<Evaluator>> &subevaluators_)
-    : subevaluators(subevaluators_) {
+    const options::Options &opts)
+    : Evaluator(opts),
+      subevaluators(opts.get_list<shared_ptr<Evaluator>>("evals")) {
```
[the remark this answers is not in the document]
silvansievers: Exactly.

src/search/evaluators/g_evaluator.cc
```diff
@@ -8,6 +8,10 @@
 using namespace std;
 
 namespace g_evaluator {
+GEvaluator::GEvaluator(const options::Options &opts)
```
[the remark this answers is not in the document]
silvansievers: I have never heard of this. Do you know of any use in Fast Downward?

src/search/evaluators/weighted_evaluator.cc
```diff
@@ -12,14 +12,11 @@ using namespace std;
 
 namespace weighted_evaluator {
 WeightedEvaluator::WeightedEvaluator(const Options &opts)
-    : evaluator(opts.get<shared_ptr<Evaluator>>("eval")),
+    : Evaluator(opts),
+      evaluator(opts.get<shared_ptr<Evaluator>>("eval")),
       w(opts.get<int>("weight")) {
 }
 
-WeightedEvaluator::WeightedEvaluator(const shared_ptr<Evaluator> &eval, int weight)
```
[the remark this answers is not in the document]
silvansievers: It has. We now need to construct evaluators with an options object because the base class needs it. Thus having a second constructor without the options object is no longer possible.

src/search/heuristic.cc
```diff
@@ -16,7 +16,7 @@
 using namespace std;
 
 Heuristic::Heuristic(const Options &opts)
-    : Evaluator(opts.get_unparsed_config(), true, true, true),
+    : Evaluator(opts, opts.get_unparsed_config(), true, true, true),
```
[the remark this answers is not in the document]
silvansievers: Good point.

src/search/heuristics/additive_heuristic.cc
```diff
@@ -18,15 +18,17 @@ const int AdditiveHeuristic::MAX_COST_VALUE;
...
+    }
 }
 
 void AdditiveHeuristic::write_overflow_warning() {
     if (!did_write_overflow_warning) {
         // TODO: Should have a planner-wide warning mechanism to handle
         // things like this.
-        utils::g_log << "WARNING: overflow on h^add! Costs clamped to "
-                     << MAX_COST_VALUE << endl;
+        log << "WARNING: overflow on h^add! Costs clamped to "
```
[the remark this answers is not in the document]
silvansievers: Yes. I'm not sure I agree it is useful to do this now, but fine by me. What is weird, though, is that here we have a warning and "a warning on stderr" right below. Should we remove one of these?

src/search/landmarks/landmark_factory.cc
```diff
@@ -18,6 +18,10 @@
 using namespace std;
 
 namespace landmarks {
+LandmarkFactory::LandmarkFactory(const options::Options &opts)
+    : log(utils::get_log_from_options(opts)), lm_graph(nullptr) {
```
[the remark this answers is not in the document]
silvansievers: It is not necessary but explicit.

src/search/evaluators/g_evaluator.cc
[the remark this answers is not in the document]
maltehelmert: Much of our code is older than C++11, and also I'm not sure how often the scenario arises in our code where we have a nontrivial constructor that we would like to inherit, and nothing "new" needs to be done in the derived class constructor. I guess usually there would be something in the derived class that we need to take care of in the constructor.

It could also be the case that we want to inherit some, but not all constructors for some reason, and in that case I don't think `using` is an option. Maybe also some people have a preference for the more explicit form so that they can more easily see how an object is constructed. In any case, I'm not aware of any objections that someone has raised against inheriting constructors.

src/search/landmarks/util.cc
```diff
@@ -79,67 +79,79 @@ int get_operator_or_axiom_id(const OperatorProxy &op) {
...
-        if (!first) {
-            if (landmark.disjunctive) {
-                cout << " | ";
-            } else if (landmark.conjunctive) {
-                cout << " & ";
+static void dump_node(
+    const TaskProxy &task_proxy,
+    const LandmarkNode &node,
+    utils::LogProxy &log) {
+    if (log.is_at_least_debug()) {
```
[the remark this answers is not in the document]
silvansievers: This is what we decided in the meeting two days ago: every logging must be guarded. Someone could just call dump_node from outside dump_landmark_graph.

src/search/merge_and_shrink/merge_and_shrink_algorithm.cc
```diff
@@ -496,9 +488,10 @@ void handle_shrink_limit_options_defaults(Options &opts) {
         }
     }
 
+    utils::LogProxy log = utils::get_log_from_options(opts);
     if (max_states_before_merge > max_states) {
-        utils::g_log << "warning: max_states_before_merge exceeds max_states, "
-                     << "correcting." << endl;
+        log << "warning: max_states_before_merge exceeds max_states, "
```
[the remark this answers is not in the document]
silvansievers: I don't like using CAPS at all.

src/search/heuristics/additive_heuristic.cc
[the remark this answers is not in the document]
maltehelmert: The motivation is that we want to have it on `cerr` so that it shows up (for example) in the "unexpected errors" of the lab report, but we also want it to show up in the log (which it wouldn't if it were only printed on `cerr`). I'd leave this as-is for now, but it's something that we could address in a better way by improving our logging framework. For example, warning-level (or worse) log messages could *always* be printed on both streams, or this could be configurable.

src/search/merge_and_shrink/utils.cc
```diff
@@ -127,7 +127,7 @@ bool shrink_before_merge_step(
         shrink_threshold_before_merge,
         shrink_strategy,
         log);
-    if (log.is_at_least_verbose() && shrunk1) {
+    if (shrunk1) {
```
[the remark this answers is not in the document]
silvansievers: Here we only print statistics if something related to merge-and-shrink happened. This has nothing to do with the log level.

src/search/landmarks/util.cc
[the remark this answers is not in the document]
maltehelmert: Exactly, it will go away in the redesign. Having the `at_least_debug` right here will help us in that revision to figure out which method we need to call (such as `log.debug`) without having to think. In the redesign, there won't be a way to log without specifying the level of the log message, so we will need to know then what the log level of the message is.

src/search/pdbs/dominance_pruning.cc
```diff
@@ -204,17 +210,18 @@ void prune_dominated_cliques(
     }
 
     int num_pruned_collections = num_cliques - remaining_pattern_cliques.size();
```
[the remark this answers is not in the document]
silvansievers: No, because the containers are swapped right after this line.

src/search/pdbs/pattern_database.cc
```diff
@@ -100,8 +102,6 @@ PatternDatabase::PatternDatabase(
         }
     }
     create_pdb(task_proxy, operator_costs, compute_plan, rng, compute_wildcard_plan);
-    if (dump)
-        utils::g_log << "PDB construction time: " << timer << endl;
```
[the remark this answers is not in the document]
silvansievers: I thought that it isn't: the reason is that unless we use a single PDB heuristic, we construct thousands of PDBs and having thousands of these lines doesn't help at all, nor can or should we parse them. If someone using a single PDB heuristic really wants to know how long the PDB took to compute, then they could simply measure the time outside of the PDB constructor. Also, with the planned separation of PDB construction from the class PatternDatabase, this output should be gone anyway.

src/search/search_engine.cc
silvansievers: I don't understand how this is related. Now each Evaluator has its own log, so we don't need to hand it in anymore. Instead, if search wants to modify logging of its evaluators, then the evaluators should be created with the right option. (I agree that it feels more natural to use the logger of the search, but implementing hierarchical loggers is for the future.)

src/search/search_engines/search_common.cc
silvansievers: Objects which are created from the command line should not be created from code, in principle. But getting rid of these things is only possible after solving the component interaction problem.

maltehelmert: We had similar discussions in the past. There are several drawbacks to using `Options` objects in this way, which is why we had these other "direct" constructors. These are some of the main points again `Options`:

1. Requiring `Options` loses static type safety (i.e. checking at compile-time that we have the right names and types of parameters). Some things are not even checked dynamically. For example, if we add an option that doesn't exist, we won't notice.
2. Requiring `Options` introduces unnecessary overhead due to the hash-map operations and type conversions to and from `Any` objects involved.
3. Everything now has a dependency on `Options`. Even people that only want to use a fixed planner configuration for which they could in principle hard-code the constructor have to import significant parts of our option-parsing code. This can be viewed as algorithmic code depending on user-interface code, which is generally bad. (Although it's a matter of opinion whether `Options` should could as user-interface code.)
4. The code that needs to construct `Options` manually for things like here is tedious.
 
Some of these can be addressed by having two constructors like we did in these cases in the past, one taking its arguments directly and one taking `Options`. It's not something we wanted to do in all cases because of the boiler-plate involved. Also, the non-Options-based constructor calls can be hard to read when there are many arguments due to C++'s lack of support for something like keyword arguments.

It is worth reconsidering these arguments after substantial changes to the parts of the planner that influenced our decision at the time. For example, perhaps this is something we should rediscuss once we introduce builders. Perhaps builders should use Options objects, and the objects they build should not. If you think this might be the case, perhaps we can save parts of this comment and its context in a relevant place of the issue tracker.

maltehelmert: I forgot to mention: it might very well be possible to address 4. by working on the interface of Options objects. In principle, I think it might be possible to support something like

`Options options {
  {"w", 3},
  {"evaluator", eval},
  {"use_foobar", true} };`

although care would need to be taken that C++ derives the correct types.

src/search/heuristics/additive_heuristic.cc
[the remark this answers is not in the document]
silvansievers: I thought our convention was to exit after printing on `cerr`.

src/search/search_engines/search_common.cc
silvansievers: I agree that using Options is bad. To circumvent this problem here, we would also need to have a second constructor in the evaluator base class that takes the LogProxy instead of the Options object. Is this the preferred solution?

maltehelmert: I think it's OK to use `Options` here, I don't think it's the only place like this. But we should look at it again once we implement the builders.

src/search/landmarks/util.cc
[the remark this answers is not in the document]
ClemensBuechner: Will there be an option to still exclude the `dump` functions from outside? Because otherwise this would still loop over the entire landmark graph just to find that it should only log this in debug mode.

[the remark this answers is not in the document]
maltehelmert: Log level tests will still be possible everywhere. I'm not sure what you mean with "this"; this function in paticular doesn't loop over landmark nodes, and if the log level isn't at least debug, it will exit immediately. But if there is a caller that triggers a lot of logging, of course it can include its own guards.

For what it's worth, a name like `dump_...` or `print_...` doesn't work that well with the new model because I think there would be an expectation that it prints unconditionally. Probably a name like `log_...` would be better.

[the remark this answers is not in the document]
maltehelmert: Perhaps it's worth adding a section on logging to ForDevelopers, either as part of the coding conventions or on a subpage to combine these things.

PR #97, opened 2022-02-24
issue1044

ClemensBuechner:


src/search/landmarks/exploration.cc
```diff
@@ -147,124 +122,91 @@ void Exploration::build_unary_operators(const OperatorProxy &op) {
...
         }
-        op.h_add_cost = op.base_cost; // will be increased by precondition costs
         op.h_max_cost = op.base_cost;
-        op.depth = -1;
 
         if (op.unsatisfied_preconditions == 0) {
-            op.depth = 0;
-            int depth = op.is_induced_by_axiom(task_proxy) ? 0 : 1;
-            enqueue_if_necessary(op.effect, op.base_cost, depth, &op, use_h_max);
+            enqueue_if_necessary(op.effect, op.base_cost, &op);
```
maltehelmert, line 160: I think the operator costs (`base_cost`) can also go away in the next round.

src/search/landmarks/exploration.cc
```diff
@@ -147,124 +122,91 @@ void Exploration::build_unary_operators(const OperatorProxy &op) {
...
-                    enqueue_if_necessary(unary_op->effect, unary_op->h_max_cost,
-                                         depth, unary_op, use_h_max);
-                else
-                    enqueue_if_necessary(unary_op->effect, unary_op->h_add_cost,
-                                         depth, unary_op, use_h_max);
+                enqueue_if_necessary(unary_op->effect, unary_op->h_max_cost,
+                                     unary_op);
             }
         }
     }
```
maltehelmert, line 189: We already have an implementation of (qualitative) relaxed reachability in the LM-Cut code, I think. In revising this, it might make sense to look into that because it's more optimized. IIRC, it is a depth-first search implemented via recursion rather than an open list.

src/search/landmarks/exploration.cc
```diff
@@ -147,124 +122,91 @@ void Exploration::build_unary_operators(const OperatorProxy &op) {
...
-    if (use_h_max && (prop->h_max_cost == -1 || prop->h_max_cost > cost)) {
+    if (prop->h_max_cost == -1 || prop->h_max_cost > cost) {
         prop->h_max_cost = cost;
-        prop->depth = depth;
-        prop->reached_by = op;
-        prop_queue.push(cost, prop);
-    } else if (!use_h_max && (prop->h_add_cost == -1 || prop->h_add_cost > cost)) {
-        prop->h_add_cost = cost;
-        prop->depth = depth;
         prop->reached_by = op;
```
maltehelmert, line 197: Do we still use `reached_by` for anything? (Perhaps this code is still used to generate preferred operators?)

src/search/landmarks/exploration.cc
```diff
@@ -147,124 +122,91 @@ void Exploration::build_unary_operators(const OperatorProxy &op) {
...
+    vector<vector<int>> &lvl_var,
+    const vector<FactPair> &excluded_props,
+    const unordered_set<int> &excluded_op_ids) {
     // Perform exploration using h_max-values
-    setup_exploration_queue(task_proxy.get_initial_state(), excluded_props, excluded_op_ids, true);
-    relaxed_exploration(true, level_out);
+    setup_exploration_queue(task_proxy.get_initial_state(), excluded_props, excluded_op_ids);
+    relaxed_exploration();
 
     // Copy reachability information into lvl_var and lvl_op
```
maltehelmert, line 211: That comment is no longer correct. Also needs a full stop at the end.

src/search/landmarks/exploration.h
```diff
@@ -50,12 +42,12 @@ struct ExUnaryOperator {
     int base_cost; // 0 for axioms, 1 for regular operators
```
maltehelmert, line 42: I think this comment has been wrong ever since we started supporting action costs. Or is it really always 1 for regular operators? (But I suppose that's unrelated to this issue.)

src/search/landmarks/exploration.h
```diff
@@ -50,12 +42,12 @@ struct ExUnaryOperator {
...
     int unsatisfied_preconditions;
-    int h_add_cost;
+    bool excluded;
     int h_max_cost;
-    int depth;
     ExUnaryOperator(const std::vector<ExProposition *> &pre, ExProposition *eff,
                     int op_or_axiom_id, int base)
-        : op_or_axiom_id(op_or_axiom_id), precondition(pre), effect(eff), base_cost(base) {}
+        : op_or_axiom_id(op_or_axiom_id), precondition(pre), effect(eff),
+          base_cost(base), excluded(false) {}
```
maltehelmert, line 50: `base` is a bad name for `operator cost`. But I expect this will go away soon anyway. And I suppose it's not relevant to this issue.

src/search/landmarks/landmark_factory_relaxation.cc
```diff
@@ -93,8 +93,7 @@ void LandmarkFactoryRelaxation::calc_achievers(
 
         vector<vector<int>> lvl_var;
         vector<utils::HashMap<FactPair, int>> lvl_op;
```
maltehelmert, line 95: Is `lvl_op` still needed? If not, I wonder why this doesn't generate a warning.

src/search/landmarks/exploration.cc
```diff
@@ -147,124 +122,91 @@ void Exploration::build_unary_operators(const OperatorProxy &op) {
...
+        enqueue_if_necessary(init_prop, 0, 0);
     }
 
     // Initialize operator data, deal with precondition-free operators/axioms.
     for (ExUnaryOperator &op : unary_operators) {
+        op.excluded = false; // Reset from previous exploration
         op.unsatisfied_preconditions = op.precondition.size();
-        if (op.effect->h_add_cost == -2
+        if (op.effect->excluded
             || excluded_op_ids.count(op.op_or_axiom_id)) {
```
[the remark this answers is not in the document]
ClemensBuechner: Unfortunately yes, without it we're 1 character over the limit.

src/search/landmarks/exploration.cc
ClemensBuechner: Apparently not, I removed it.

src/search/landmarks/exploration.h
ClemensBuechner: Are you fine with just removing it completely?

src/search/landmarks/landmark_factory_relaxation.cc
ClemensBuechner: No, it's not. I'm also surprised there was no warning about this... It makes me wonder whether there are other instances that we overlooked.

src/search/landmarks/exploration.h
maltehelmert: Yes, as it doesn't matter for testing relaxed reachability. But if we do more changes like these in the same issue, I'd suggest rerunning the experiments because it is always good to have experimental results for the things that we merge. If running another experiment is too onerous, that may be sufficient reason to leave it to the next issue.

maltehelmert: I see. Sure, just remove the comment. Regarding when we need experiments: in this case we do the experiments to a large extent to test the performance impact, not just to test the correctness of the modification. For this, many of the individual changes had an impact.

src/search/landmarks/exploration.cc
```diff
@@ -147,124 +122,91 @@ void Exploration::build_unary_operators(const OperatorProxy &op) {
...
     // Deal with current state.
     for (FactProxy fact : state) {
         ExProposition *init_prop = &propositions[fact.get_variable().get_id()][fact.get_value()];
-        enqueue_if_necessary(init_prop, 0, 0, 0, use_h_max);
+        enqueue_if_necessary(init_prop, 0, 0);
     }
 
     // Initialize operator data, deal with precondition-free operators/axioms.
     for (ExUnaryOperator &op : unary_operators) {
+        op.excluded = false; // Reset from previous exploration
```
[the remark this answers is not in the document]
ClemensBuechner: It was needed in the old code, but happened only after the for-loop (it was line 182 before we started: `op.h_add_cost = op.base_cost`). The reason it moved is probably because we removed it completely in a first approach and only later realized that it must be reset whenever we start a new exploration.

src/search/landmarks/exploration.cc
```diff
@@ -147,124 +122,91 @@ void Exploration::build_unary_operators(const OperatorProxy &op) {
...
-    const unordered_set<int> &excluded_op_ids, bool use_h_max) {
+    const unordered_set<int> &excluded_op_ids) {
     prop_queue.clear();
 
     for (size_t var_id = 0; var_id < propositions.size(); ++var_id) {
         for (size_t value = 0; value < propositions[var_id].size(); ++value) {
             ExProposition &prop = propositions[var_id][value];
-            prop.h_add_cost = -1;
+            prop.excluded = false;
             prop.h_max_cost = -1;
```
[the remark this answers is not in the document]
ClemensBuechner: This is exactly what you already suggested in issue1045 which we will probably tackle next, so it won't be forgotten if we resolve this.

src/search/landmarks/exploration.cc
```diff
@@ -147,124 +122,91 @@ void Exploration::build_unary_operators(const OperatorProxy &op) {
...
-            prop.depth = -1;
-            prop.marked = false;
         }
     }
 
     for (const FactPair &fact : excluded_props) {
         ExProposition &prop = propositions[fact.var][fact.value];
-        prop.h_add_cost = -2;
+        prop.excluded = true;
     }
```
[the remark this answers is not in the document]
ClemensBuechner: I'm wondering when would be the best time to do this. I see it fitting in this issue rather than a follow up, but we wouldn't get rid of the loop over all propositions completeley (only the setting `excluded` to false part) since we need to reset the `h_max_cost`. I therefore consider just writing a comment for now, describing your suggestion and leaving it for when we work on issue1045 to replace the `h_max_cost` with a `reached` flag anyway. I believe the `reached` would still need a reset in every exploration, but maybe there's another smart way to avoid doing that for all propositions, I didn't think about it for too long.

src/search/landmarks/exploration.cc
ClemensBuechner: Will probably happen when we replace `h_max_cost` by a boolean `reached` in issue1045.

src/search/landmarks/exploration.cc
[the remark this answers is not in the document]
maltehelmert: Sure, there may still be other O(|atoms|) processing here. But less processing will be faster. I have no strong preference regarding when to look into this, with a weak preference to merging soon provided we keep the momentum and continue working on this.

src/search/landmarks/exploration.cc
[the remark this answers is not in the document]
maltehelmert: I see, thanks. I think that logic is a bit convoluted, both the split initialization where we already essentially have an if/else branch for excluded/non-excluded actions below anyway, and also the fact that the logic below uses `continue` rather than s normal `if/else`. Would be good to clean this up further, but no need for further changes now.

src/search/landmarks/exploration.cc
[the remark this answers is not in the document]
ClemensBuechner: I decided to address this right away and made the change within this issue. I've pushed the changes a minute ago, so you should be able to see the change now. This requires another experiment, so I'm setting this up now.

src/search/landmarks/exploration.cc
```diff
@@ -147,158 +121,99 @@ void Exploration::build_unary_operators(const OperatorProxy &op) {
...
     const State &state, const vector<FactPair> &excluded_props,
-    const unordered_set<int> &excluded_op_ids, bool use_h_max) {
+    const unordered_set<int> &excluded_op_ids) {
     prop_queue.clear();
 
-    for (size_t var_id = 0; var_id < propositions.size(); ++var_id) {
-        for (size_t value = 0; value < propositions[var_id].size(); ++value) {
-            ExProposition &prop = propositions[var_id][value];
-            prop.h_add_cost = -1;
+    for (auto &proposition : propositions) {
```
maltehelmert, line 127: `proposition` is not a fitting name, the first-level elements of `propositions` are nor propositions, only the second-level elements are. The fact that `prop` is an element of `proposition` makes it clear that there's something fishy here, I think. :-) Perhaps a name like `variable`, `var_propositions`, `propositions_for_variable` or similar instead.

maltehelmert: Hi Clemens, das ist als "resolved" markiert, aber die Namen sind noch dieselben. Sollten wir das noch besprechen?

PR #99, opened 2022-04-01
Issue1045

salome-eriksson:
Refactor the Exploration Class in the landmarks code.

Main changes:
 - `Exploration` now only computes reachability information instead of h^max values.
 - `reached` (previously `lvl_var`) is built within `Exploration` rather than passed from outside.
 - The computation of which operators should be excluded because of excluded propositions now happens inside `Exploration`.
 - `UnaryOperator` only stores the number of preconditions but no references to the actual preconditions anymore.


src/search/landmarks/exploration.cc
```diff
@@ -14,21 +14,17 @@
...
+  Integration Note: this class is the same as (rich man's) FF heuristic
+  (taken from hector branch) except for the following:
+  - Added-on functionality for excluding certain operators from the relaxed
+    exploration. (These operators are never applied, as necessary for landmark
+    computation.)
+  - Unary operators are not simplified, because this may conflict with excluded
+    operators. (For an example, consider that unary operator o1 is thrown out
+    during simplify() because it is dominated by unary operator o2, but then o2
+    is excluded during an exploration ==> the shared effect of o1 and o2 is
+    wrongly never reached in the exploration.)
```
salome-eriksson, line 27: I wonder if we should keep this comment at all; I don't know if anyone can still reproduce the FF heuristic from hector branch, and we changed a lot in the class in this revision.

src/search/landmarks/exploration.cc
```diff
@@ -39,19 +35,25 @@ Exploration::Exploration(const TaskProxy &task_proxy)
...
-    for (FactProxy goal_fact : task_proxy.get_goals()) {
-        int var_id = goal_fact.get_variable().get_id();
-        int value = goal_fact.get_value();
-        goal_propositions.push_back(&propositions[var_id][value]);
-        termination_propositions.push_back(&propositions[var_id][value]);
+    /*
+      Reserve vector size unary operators. This is needed because we
+      cross-reference to the memory address of elements of the vector while
+      building it; meaning a resize would invalidate all references.
+    */
```
salome-eriksson, line 48: It would probably be better to not use raw pointers to vector entries at all and for example use IDs instead.  Maybe a good follow-up issue?

src/search/landmarks/exploration.cc
```diff
@@ -104,115 +105,117 @@ void Exploration::build_unary_operators(const OperatorProxy &op) {
...
     const vector<FactPair> &excluded_props,
     const unordered_set<int> &excluded_op_ids) {
-    // Perform exploration using h_max-values.
-    setup_exploration_queue(task_proxy.get_initial_state(), excluded_props, excluded_op_ids);
+    setup_exploration_queue(task_proxy.get_initial_state(),
+                            excluded_props, excluded_op_ids);
     relaxed_exploration();
 
-    // Copy reachability information into lvl_var.
+    // Bundle reachability information into the return data structure.
```
salome-eriksson, line 208: One follow-up issue Clemens and me discussed is having a proper class for this reachability information. This class could then potentially take on some responsibilities that are currently in the rhw factory and util, such as `possibly_reaches_lm` from util

src/search/landmarks/exploration.cc
maltehelmert: The "FF heuristic from hector branch" is our FF heuristic.

I think the first half of the comment is no longer useful and should be removed. I think the warning regarding "simplify" is still useful because it mentions a dangerous trap. I suggest rewriting the comment to only mention this, i.e., say that the simplification method in our delete relaxation (RelaxationHeuristic::simplify) cannot be used here for the reasons mentioned.

src/search/landmarks/exploration.cc
maltehelmert: Using IDs can be better or worse than using pointers, it's hard to predict. I would only create a follow-up issue if you intend to work on it immediately, since I would say for now it's at most a "possibly nice to have".

src/search/landmarks/exploration.cc
maltehelmert: Does this work with the plan of making landmarks abstract and polymorphic? That would normally imply that logic related to things like "does something reach a landmark?" needs to live with the landmarks so that different kinds of landmarks can do this differently.

src/search/landmarks/landmark_factory_rpg_sasp.cc
```diff
@@ -480,28 +495,30 @@ void LandmarkFactoryRpgSasp::generate_relaxed_landmarks(
...
+      at the moment we don't know which of these var-val pairs will be LMs).
+    */
     VariablesProxy variables = task_proxy.get_variables();
-    find_forward_orders(variables, lvl_var, lmp);
+    find_forward_orders(variables, reached, lmp);
 
-    // Use domain transition graphs to find further orders. Only possible if lmp is
-    // a simple landmark.
+    // Use domain transition graphs to find further orders. Only possible
+    // if lmp is a simple landmark.
```
remochristen, line 510: This and the following three comments could be in the new multi-line style as well.

remochristen (commented):
I looked through the changes and added some very minor comments, mostly about naming and grammar in the code comments. The changes are definitely an improvement in terms of readability.

src/search/landmarks/exploration.cc
```diff
@@ -39,39 +31,38 @@ Exploration::Exploration(const TaskProxy &task_proxy)
     // Build propositions.
     for (VariableProxy var : task_proxy.get_variables()) {
         int var_id = var.get_id();
-        propositions.push_back(vector<ExProposition>(var.get_domain_size()));
+        propositions.push_back(vector<Proposition>(var.get_domain_size()));
```
silvansievers, line 34: Use emplace instead of push_back?

PR #104, opened 2022-06-08
Issue1009

ClemensBuechner:


src/search/landmarks/landmark_count_heuristic.cc
```diff
@@ -374,8 +421,21 @@ static shared_ptr<Heuristic> _parse(OptionParser &parser) {
                             "(see OptionCaveats#Using_preferred_operators_"
                             "with_the_lmcount_heuristic)", "false");
     parser.add_option<bool>("alm", "use action landmarks", "true");
+    parser.add_option<int>("derived_lm_cost",
+                           "the cost to use for derived landmarks",
+                           "1", Bounds("0", "1"));
```
ClemensBuechner, line 426: What's the behaviour if I call `lmcount` with `admissible=true` and `derived_lm_cost=1`? Is it still admissible? I didn't follow the discussions in the issue tracker and on Discord too closely, so you probably have explained it there already.
Personally, I think the default "0" is more intuitive and in my opinion it wouldn't hurt to make the alias explicitly specify that we use derived landmark costs "1", but I can live with either option.

src/search/landmarks/landmark_count_heuristic.cc
```diff
@@ -374,8 +421,21 @@ static shared_ptr<Heuristic> _parse(OptionParser &parser) {
...
+
+    parser.document_note("Note on performance for satisficing planning",
+                         "The cost of a landmark is based on the cost of the "
+                         "operators that achieve it. For satisficing search "
+                         "this can be counterproductive since it is often "
+                         "better to focus on distance from goal "
+                         "(i.e. length of the plan) rather than cost."
+                         "In experiments we achieved the best performance using"
+                         "options 'transform=adapt_costs(one)' and "
+                         "'derived_lm_cost=1'.");
```
ClemensBuechner, line 438: Good phrasing.

src/search/landmarks/landmark_count_heuristic.h
```diff
@@ -23,11 +23,21 @@ class LandmarkCountHeuristic : public Heuristic {
...
     std::unique_ptr<LandmarkStatusManager> lm_status_manager;
     std::unique_ptr<LandmarkCostAssignment> lm_cost_assignment;
     std::unique_ptr<successor_generator::SuccessorGenerator> successor_generator;
 
+    /*
+      TODO: The following vectors are only used in the non-admissible case. This
+       is bad design, but given this is already the case for
+       *lm_cost_assignment* above, this is just another indicator to change this
+       in the future.
+    */
```
ClemensBuechner, line 37: Agreed, we should split the admissible and the inadmissible landmark heuristics in a future issue.

src/search/landmarks/landmark_status_manager.h
```diff
@@ -24,16 +22,12 @@ class LandmarkStatusManager {
...
-
-    bool is_initial_state_dead_end() const;
-    void compute_unachievable_landmark_ids();
 public:
     explicit LandmarkStatusManager(LandmarkGraph &graph);
 
     BitsetView get_reached_landmarks(const State &state);
 
     void update_lm_status(const State &ancestor_state);
-    bool dead_end_exists() const;
```
ClemensBuechner, line 36: Very nice that we could get rid of all of these members. :-)

ClemensBuechner (commented):
I believe this pull request is in good shape and quality. I agree with the changes and didn't find anything that needs to be changed before it can be merged. I left some comments for small improvements and raised some discussions, but there's really nothing that I deem important enough to keep us from merging this before the end of the sprint.

src/search/landmarks/landmark_cost_assignment.cc
```diff
@@ -61,7 +61,8 @@ double LandmarkUniformSharedCostAssignment::cost_sharing_h_value(
         if (lmn_status != lm_reached) {
             const set<int> &achievers =
                 get_achievers(lmn_status, node->get_landmark());
-            assert(!achievers.empty());
+            if (achievers.empty())
+                return numeric_limits<double>::max();
```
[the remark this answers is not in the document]
salome-eriksson: We could also have kept the old handling of dead-ends but this approach seemed nicer to us and we didn't observe any performance drop in experiments.

src/search/landmarks/landmark_count_heuristic.cc
```diff
@@ -108,6 +112,45 @@ LandmarkCountHeuristic::LandmarkCountHeuristic(const options::Options &opts)
...
+       *min_first_achiever_costs* and *min_possible_achiever_costs*
+       at index i corresponds to the entry for the landmark node with ID i.
+    */
+    min_first_achiever_costs.reserve(lgraph->get_num_landmarks());
+    min_possible_achiever_costs.reserve(lgraph->get_num_landmarks());
+    for (auto &node : lgraph->get_nodes()) {
+        /*
+           Currently derived landmarks are unsupported in the admissible
+           case and ignored in the inadmissible case. We ensure the latter
+           by setting minimum achiever costs to 0.
```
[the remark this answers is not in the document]
salome-eriksson: The admissible code cannot handle derived landmarks; it exits with SEARCH_UNSUPPORTED if axioms are detected. But it is admittedly not obvious that this option is only for the inadmissible setting. I suggest to add a short comment to the option documentation, e.g. "the cost to use for derived landmarks (only applicable when admissible=false, since the admissible lmcount heuristic does not support axioms)".

src/search/landmarks/landmark_count_heuristic.cc
salome-eriksson: (for the first part, see my comment above)
I was on the fence whether 0 or 1 should be the default and decided for 1 because it performs better empirically. I also hope that once we get around to thinking about how to deal with derived landmarks properly we can get rid of the option again.

src/search/landmarks/landmark_count_heuristic.cc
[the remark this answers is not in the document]
ClemensBuechner: I don't understand the connection between my comment and your reaction. I believe this function is only executed in the inadmissible case (maybe add an assertiaon?). Hence, it doesn't really matter here, that the admissible case cannot deal with axioms (for now).
My comment aimed at the recent change that derived landmarks aren't simply ignored anymore, but one can use the option `derived_lm_cost` to decide whether derived landmarks should be considered or not. More specifically, in lines 141--142 you set the achiever costs to this value and not always to 0, so there's an inconsistency between the code and the comment above, isn't there?

src/search/landmarks/landmark_count_heuristic.cc
ClemensBuechner: As mentioned on Discord: leaving 1 as default because it's the better performing option is reasonable. I suggest to add a comment to explain why this default performs better.

PR #116, opened 2022-07-13
Issue1058

silvansievers:


src/search/pruning/limited_pruning.cc
```diff
@@ -20,7 +21,7 @@ LimitedPruning::LimitedPruning(const Options &opts)
 void LimitedPruning::initialize(const shared_ptr<AbstractTask> &task) {
     PruningMethod::initialize(task);
     pruning_method->initialize(task);
-    utils::g_log << "pruning method: limited" << endl;
+    log << "pruning method: limited" << endl;
```
maltehelmert, line 24: We probably don't want to print this in silent mode?

src/search/pruning/null_pruning_method.cc
```diff
@@ -8,22 +8,28 @@
...
 
 namespace null_pruning_method {
+NullPruningMethod::NullPruningMethod(const Options &opts)
+    : PruningMethod(opts) {
+}
+
 void NullPruningMethod::initialize(const shared_ptr<AbstractTask> &task) {
     PruningMethod::initialize(task);
-    utils::g_log << "pruning method: none" << endl;
+    log << "pruning method: none" << endl;
```
maltehelmert, line 17: and here

src/search/pruning/stubborn_sets_atom_centric.cc
```diff
@@ -11,13 +11,14 @@ using namespace std;
...
-    : use_sibling_shortcut(opts.get<bool>("use_sibling_shortcut")),
+    : StubbornSets(opts),
+      use_sibling_shortcut(opts.get<bool>("use_sibling_shortcut")),
       atom_selection_strategy(opts.get<AtomSelectionStrategy>("atom_selection_strategy")) {
 }
 
 void StubbornSetsAtomCentric::initialize(const shared_ptr<AbstractTask> &task) {
     StubbornSets::initialize(task);
-    utils::g_log << "pruning method: atom-centric stubborn sets" << endl;
+    log << "pruning method: atom-centric stubborn sets" << endl;
```
maltehelmert, line 21: and here

src/search/pruning/stubborn_sets_ec.cc
```diff
@@ -122,7 +126,7 @@ void StubbornSetsEC::initialize(const shared_ptr<AbstractTask> &task) {
     disabled.resize(num_operators);
     disabled_computed.resize(num_operators, false);
 
-    utils::g_log << "pruning method: stubborn sets ec" << endl;
+    log << "pruning method: stubborn sets ec" << endl;
```
maltehelmert, line 129: and here

src/search/pruning/stubborn_sets_simple.cc
```diff
@@ -10,11 +10,15 @@
...
+StubbornSetsSimple::StubbornSetsSimple(const options::Options &opts)
+    : StubbornSets(opts) {
+}
+
 void StubbornSetsSimple::initialize(const shared_ptr<AbstractTask> &task) {
     StubbornSets::initialize(task);
     interference_relation.resize(num_operators);
     interference_relation_computed.resize(num_operators, false);
-    utils::g_log << "pruning method: stubborn sets simple" << endl;
+    log << "pruning method: stubborn sets simple" << endl;
```
maltehelmert, line 21: and here

src/search/pruning_method.cc
```diff
@@ -23,24 +27,47 @@ void PruningMethod::initialize(const shared_ptr<AbstractTask> &task_) {
 void PruningMethod::prune_operators(
     const State &state, vector<OperatorID> &op_ids) {
     assert(!task_properties::is_goal_state(TaskProxy(*task), state));
-    timer.resume();
+    /*
+      We only measure time with verbose verbosity level because time
+      measurements induce a significant overhead, up to 30% for configurations
+      like blind search. See issue1042 and issue1058.
```
maltehelmert, line 33: Perhaps add "with the NullPruning method" after "like blind search"?

src/search/pruning_method.h
```diff
@@ -15,22 +16,30 @@ namespace limited_pruning {
...
+}
+
 class PruningMethod {
     utils::Timer timer;
     friend class limited_pruning::LimitedPruning;
 
     virtual void prune(
         const State &state, std::vector<OperatorID> &op_ids) = 0;
 protected:
+    mutable utils::LogProxy log;
```
maltehelmert, line 31: I assume this does have to be mutable? It's not pretty.

silvansievers: Yes, unfortunately, because output streams are modified when used.

maltehelmert: If we think it is OK to do output on const logs, it would probably be good to encapsulate this inside the LogProxy class. But that's something for another issue then.

PR #118, opened 2022-07-13
Explicitly qualifies all calls to std::move

vulder:
In future version of the C++ standard unqualified calls to move can
conflict with other functions called move (e.g., std::views::move) in
the standard library. So, to prevent future compile errors,
incompatabilities, or bugs we should qualify all calls to std::move.

Newer compiler version already warn about this usage pattern:
-Wunqualified-std-cast-call


vulder:
@jendrikseipp
Hi,
you might remember me from the talk Christian Kaltenecker and I gave in Saarbrücken.
As I was checking out the FDW codebase I found this small issue and cleaned it up.
Could you please take a look at the PR.

kind regards,
Florian

roeger:
Hi Florian! Sorry that this PR rested so long. With this change the style checks fail because they explicitly require that there is no std:: in .cc files. So I cannot simply merge this into main without further discussion with the rest of the dev team.

From what compiler and version did you get this warning? I currently update our test infrastructure to somewhat newer systems and tools and even with our newest tested versions I did not see a warning, yet.

vulder:
As far as I know, `clang-15` is the first compiler release that includes the new warning.

The underlying problems (selecting the wrong move) that could appear, however, could start also with earlier compilers versions, basically with every compiler/standard library that has C++20 ranges support.

maltehelmert:
We discussed this on our internal Discord some time ago, but then I forgot to get back to you. :-( Sorry! We've decided not to make this change for now. Unless I'm overlooking something, we won't get compiler errors now or in the future unless we explicitly add another `using` declaration like `using namespace std::views`. We'll look into compiler warnings once we start encountering them. For now, making this change doesn't have a good cost/benefit ratio for us. We're generally sympathetic towards moving to more modern C++ style sensibilities, such as getting rid of `using namespace std` entirely. But it's not something we want to do right now.

vulder:
Ok

PR #125, opened 2022-10-20
Issue1067

roeger:


src/search/cegar/transition_system.cc
```diff
@@ -326,11 +326,15 @@ int TransitionSystem::get_num_loops() const {
...
+#endif
         int total_outgoing_transitions = 0;
         int total_loops = 0;
         for (int state_id = 0; state_id < get_num_states(); ++state_id) {
+#ifndef NDEBUG
             total_incoming_transitions += incoming[state_id].size();
+#endif
             total_outgoing_transitions += outgoing[state_id].size();
             total_loops += loops[state_id].size();
         }
```
jendrikseipp, line 340: We could use `utils::unused_variable(total_incoming_transitions);` instead of the preprocessor directives. Same below.

.github/workflows/ubuntu.yml
```diff
@@ -19,20 +19,24 @@ jobs:
...
-        ubuntu-version: [ubuntu-18.04, ubuntu-20.04]
+        ubuntu-version: [ubuntu-20.04, ubuntu-22.04]
         compiler-version:
           - {cc: gcc, cxx: g++}
-          - {cc: gcc-10, cxx: g++-10}
+          - {cc: gcc-12, cxx: g++-12}
           - {cc: clang, cxx: clang++}
-          - {cc: clang-11, cxx: clang++-11}
-        python-version: [3.6]
+# For ubuntu 22.04 we do not have to explicitly include the newest clang
```
jendrikseipp, line 27:
```suggestion
# For Ubuntu 22.04 we do not have to explicitly include the newest Clang
```

.github/workflows/ubuntu.yml
```diff
@@ -19,20 +19,24 @@ jobs:
...
+# because it happens to be the default version
```
jendrikseipp, line 28:
```suggestion
# because it happens to be the default version.
```

.github/workflows/ubuntu.yml
```diff
@@ -19,20 +19,24 @@ jobs:
...
         # Unfortunately, we couldn't figure out a way to name the
         # compiler versions so that we don't have to copy them here.
         exclude:
-          - ubuntu-version: ubuntu-18.04
-            compiler-version: {cc: gcc-10, cxx: g++-10}
-          - ubuntu-version: ubuntu-18.04
-            compiler-version: {cc: clang-11, cxx: clang++-11}
+          - ubuntu-version: ubuntu-20.04
+            compiler-version: {cc: gcc-12, cxx: g++-12}
+# If we explicitly include the newest clang above, we need to exclude
```
jendrikseipp, line 36:
```suggestion
# If we explicitly include the newest Clang above, we need to exclude
```

.github/workflows/ubuntu.yml
```diff
@@ -19,20 +19,24 @@ jobs:
...
+# it here for the older ubuntu version
```
jendrikseipp, line 37:
```suggestion
# it here for the older Ubuntu version.
```

.github/workflows/mac.yml
```diff
@@ -11,15 +11,15 @@ jobs:
   test:
     name: Compile and test planner
     timeout-minutes: 60
-    runs-on: macos-10.15
+    runs-on: macos-11
```
jendrikseipp, line 14:
```suggestion
    runs-on: [macos-11, macos-12]
```
Or use macos-12 instead of macos-11?

.github/workflows/style.yml
```diff
@@ -10,20 +10,20 @@ on:
 jobs:
   style:
     name: Test code style
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
```
jendrikseipp, line 13: These changes also require a changelog entry and updating README.md.

jendrikseipp (changes requested):
Thanks for tackling this! We're moving forward :-)

.github/workflows/mac.yml
roeger: I guess there is a reason why macos-latest is still pointing to macos-11. Could it be the case that it is still the more common version? And I am not convinced that we have sufficient benefit of running two versions with the increased environmental footprint.

.github/workflows/style.yml
roeger: Thanks for the hint to README.md, I updated the information there. Since this is only about our test environment, I don't think we need a changelog entry (and I haven't found such an entry for earlier such updates).

PR #141, opened 2022-12-15
Issue1070

ClemensBuechner:


src/search/landmarks/landmark_count_heuristic.cc
```diff
@@ -297,22 +316,6 @@ bool LandmarkCountHeuristic::dead_ends_are_reliable() const {
...
-    LandmarkNodeSet landmark_node_set;
-    for (int i = 0; i < landmark_bitset.size(); ++i) {
-        if (landmark_bitset.test(i)) {
-            landmark_node_set.insert(lgraph->get_node(i));
-        }
-    }
-    return landmark_node_set;
-}
-
-
```
jendrikseipp, line 315: :heart:

src/search/landmarks/landmark_count_heuristic.cc
```diff
@@ -263,20 +254,48 @@ bool LandmarkCountHeuristic::generate_helpful_actions(
...
-    if (num_reached != lgraph->get_num_landmarks()) {
-        if (reached.find(&lm_node) != reached.end())
-            return false;
-        else
-            return !check_node_orders_disobeyed(lm_node, reached);
+    const State &/*state*/, LandmarkNode &lm_node) const {
+    /*
+      A landmark is interesting if it hasn't been reached before and
+      its parents have all been reached, or if all landmarks have been
+      reached before, the LM is a goal, and it's not true at moment.
```
jendrikseipp, line 261:
```suggestion
      reached before, the LM is a goal, and it's not true at the moment.
```

src/search/landmarks/landmark_count_heuristic.cc
```diff
@@ -263,20 +254,48 @@ bool LandmarkCountHeuristic::generate_helpful_actions(
...
-        if (reached.find(&lm_node) != reached.end())
-            return false;
-        else
-            return !check_node_orders_disobeyed(lm_node, reached);
+    const State & /*state*/, LandmarkNode &lm_node) const {
+    /*
+      A landmark is interesting if it hasn't been reached before and
+      its parents have all been reached, or if all landmarks have been
+      reached before, the LM is a goal, and it's not true at the moment.
+    */
```
salome-eriksson, line 262: This comment is not true anymore. Either remove it or (better) replace it with something like this:
A landmark is interesting if it must be achieved in the future and all its parents have been reached.
I would then also remove the comment for the `lm_not_reached` case below

src/search/landmarks/landmark_count_heuristic.cc
```diff
@@ -199,30 +201,19 @@ int LandmarkCountHeuristic::compute_heuristic(const State &ancestor_state) {
...
-     return false. If a simple landmark can be achieved, return only operators
-     that achieve simple landmarks, else return operators that achieve
-     disjunctive landmarks */
+bool LandmarkCountHeuristic::generate_helpful_actions(const State &state) {
+    /*
+      Find actions that achieve new landmark leaves. If no such action exist,
+      return false. If a simple landmark can be achieved, return only operators
+      that achieve simple landmarks, else return operators that achieve
+      disjunctive landmarks.
+    */
```
maltehelmert, line 216: This comment contradicts the change log message. The change log message suggests that we are happy with achieving *any* landmark. But here the first sentence says "landmark leaves", which means landmarks whose parents have all been accepted.

src/search/landmarks/landmark_count_heuristic.cc
```diff
@@ -199,30 +201,19 @@ int LandmarkCountHeuristic::compute_heuristic(const State &ancestor_state) {
...
-    return false;
-}
-
-bool LandmarkCountHeuristic::generate_helpful_actions(
-    const State &state, const LandmarkNodeSet &reached) {
-    /* Find actions that achieve new landmark leaves. If no such action exist,
-     return false. If a simple landmark can be achieved, return only operators
-     that achieve simple landmarks, else return operators that achieve
-     disjunctive landmarks */
+bool LandmarkCountHeuristic::generate_helpful_actions(const State &state) {
```
maltehelmert, line 210: We should clean up the signature: there is no longer a reason to have this be a non-void function. The general preference is for mutating functions like these ones to be void; the return value is not obvious from the name and is unused. Also, the function name should be generate_preferred_operators; "helpful actions" is an EHC/FF term. If we make this void, the following comment also should be adjusted.

src/search/landmarks/landmark_count_heuristic.cc
maltehelmert: Another error in the comment: it implies the false dichotomy that landmarks are either simple or disjunctive; there is the third option of being conjunctive. From the comment, it is not clear how conjunctive landmarks are handled. (I assume this is because this code predates the addition of conjunctive landmarks.)

src/search/landmarks/landmark_count_heuristic.cc
```diff
@@ -263,20 +254,44 @@ bool LandmarkCountHeuristic::generate_helpful_actions(
 }
```
maltehelmert, line 254: github doesn't let me comment on it directly because it's not close enough to a modified line, so I'm adding this comment here: The two lines with the `return false` case above this block no longer make sense and should be removed. There is no need to add special code to avoid iterating over an empty vector.

src/search/landmarks/landmark_count_heuristic.cc
maltehelmert: This also entails changing the variable names that use "ha".

src/search/landmarks/landmark_count_heuristic.cc
maltehelmert: Actually, I wonder how the code handles conjunctive landmarks. From looking at this a bit more, I think it either ignores them or handles them wrong. Can we discuss this?

src/search/landmarks/landmark_count_heuristic.cc
```diff
@@ -263,20 +254,44 @@ bool LandmarkCountHeuristic::generate_helpful_actions(
...
+    case lm_needed_again:
+        return all_of(lm_node.parents.begin(), lm_node.parents.end(),
+                      [&](const pair<LandmarkNode *, EdgeType> &parent) {
+                          int parent_id = parent.first->get_id();
+                          return lm_status_manager->is_reached(parent_id);
+                      });
+    case lm_reached:
+    default:
+        return false;
     }
```
maltehelmert, line 294: I think a `switch` that only really looks for one specific case (albeit a disjunctive one) is more cleanly expressed as an `if`. I'm not sure the `all_of` is easier to understand/read than a direct `for` loop, but if you prefer it, fine.

CHANGES.md
```diff
@@ -11,7 +11,14 @@ after the corresponding tracker issues.
...
 ## Changes since the last release
 
-- none so far
+- landmarks: When computing preferred operators, the `lmcount` heuristic
+  now considers all landmarks that have to be achieved in the future,
+  not only those that have not yet been accepted. This speeds up the
+  first iteration of LAMA with preferred operators and leads to an 
+  increased coverage on the IPC benchmarks, but does not affect other 
+  statistics (e.g., plan quality) significantly. (Holds also for the 
+  full anytime LAMA.)
```
maltehelmert, line 20: Is this change log entry up to date? You wrote on the tracker that there are no more semantic changes.

src/search/landmarks/landmark_count_heuristic.cc
maltehelmert: Looks like Jendrik's comment is not yet resolved.

src/search/landmarks/landmark_count_heuristic.cc
```diff
@@ -237,46 +242,49 @@ bool LandmarkCountHeuristic::generate_helpful_actions(
...
+    if (all_lms_reached) {
+        const Landmark &landmark = lm_node.get_landmark();
+        return landmark.is_true_in_goal && !landmark.is_true_in_state(state);
+    } else if (reached.test(lm_node.get_id())) {
+        return false;
+    } else {
+        return all_of(lm_node.parents.begin(), lm_node.parents.end(),
+                      [&](const pair<LandmarkNode *, EdgeType> parent) {
+                          return reached.test(parent.first->get_id());
+                      });
```
maltehelmert, line 286: To me, based on the comment above, this would be clearer as a two-case distinction rather than a three-case distinction, with the second case being "!reached.test(...) && all_of(...)". Common advice would perhaps also be to have the if/else order match the order of the two cases in the explaining comment; currently it's opposite. Since another common advice is to avoid negative conditions, this perhaps would entail changing the order in the comment, not in the code.

PR #143, opened 2023-01-27
Issue1071

ClemensBuechner:


src/search/landmarks/landmark_heuristic.h
```diff
@@ -0,0 +1,64 @@
...
+# include "../heuristic.h"
+
+class BitsetView;
+
+namespace successor_generator {
+class SuccessorGenerator;
+}
+
+namespace landmarks {
+class LandmarkFactory;
```
salome-eriksson, line 13: I don't think we need to know about LandmarkFactory here, only in the .cc file (where the header is included).

src/search/landmarks/landmark_heuristic.cc
```diff
@@ -0,0 +1,217 @@
...
+    utils::Timer lm_graph_timer;
+    if (log.is_at_least_normal()) {
+        log << "Generating landmark graph..." << endl;
+    }
+    shared_ptr<LandmarkFactory> lm_graph_factory =
+        opts.get<shared_ptr<LandmarkFactory>>("lm_factory");
+
+    /* TODO: Admissible case checks here that no reasonable orderings
+        are computed. In the long term, they should allow computing
+        reasonable orderings, though. */
```
salome-eriksson, line 48: Remove this comment

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
```diff
@@ -0,0 +1,159 @@
...
+       This could also be done by checking the landmark graph itself whether it
+       has any reasonable orderings at all.
+
+       In any case, doing this check only here means the landmark graph was
+       already computed by the base class, which would not be necessary if we
+       could ask the factory first, whether it computes reasonable orderings in
+       the first place. Since this all happens in the constructor anyway and we
+       abort if reasonable orderings are present, I don't think this is a big
+       issue, though.
+    */
```
salome-eriksson, line 39: I would shorten this comment: remove the second paragraph and the last sentence of the first, and instead add a different sentence:
TODO: The landmark graph is already constructed by the base class for
       \*LandmarkHeuristic\*. The only reason why we get the landmark factory from
       the options again is to check whether it produces reasonable orderings.
       The check for reasonable orders can be removed once landmark progression 
      correctly handles them (issue1036).

I don't know how long computing the landmark graph takes, but I can imagine it can take some time for large tasks, so it can be annoying to get this error message only after a while even though we could check it much faster. We could instead add a function "check_options_consistency" or similar to the base class (either abstract or returns true), which is called right at the beginning of the base class constructor. Then every derived class can define custom consistency checks. (In this case the comment becomes obsolete)

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
```diff
@@ -0,0 +1,159 @@
...
+
+        document_property(
+            "admissible",
+            "yes");
+        /* TODO: This was "yes with admissible=true and optimal cost
+            partitioning; otherwise no" before. Can we answer this now that
+            the heuristic only cares about admissible? */
+        document_property(
+            "consistent",
+            "complicated; needs further thought");
```
salome-eriksson, line 146: I discussed this with Malte. Any cost partitioning is consistent. However, the definition of consistent does not really make sense for path-dependent heuristics. For example if we expand a state s with h=5 then all children will be consistent with this heuristic estimate since we incorporate the landmark information from s in its successors. But we might later find a different path to s that increases its heuristic value to let's say 10. Then the heuristic can become inconsistent as long as we don't reevaluate the heuristic values of the successors (and this can lead to a chain reaction). It also depends on whether or not you use lazy_evaluator.

Suggestion (not discussed with Malte): We write "(yes), see document note about consistency" here and add a document_note, maybe something like this:
Consistency
The heuristic is consistent along single paths if it is set as lazy_evaluator; i.e. when expanding s then we have h(s) <= h(s')+cost(a) for all successors s' of s reached with a. But newly found paths to s can increase h(s), at which point the above inequality might not hold anymore."

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
```diff
@@ -0,0 +1,159 @@
...
+            "optimal",
+            "use optimal (LP-based) cost sharing",
+            "false");
+        add_option<bool>("alm", "use action landmarks", "true");
+        lp::add_lp_solver_option_to_feature(*this);
+
+        document_note(
+            "Optimal search",
+            "You probably also want to add this heuristic as a lazy_evaluator "
+            "in the A* algorithm to improve heuristic estimates.");
```
salome-eriksson, line 122: I don't like the phrasing here. Suggestion:
"Usage with A*"
"We recommend to add this heuristic as lazy_evaluator when using it in the A* algorithm. This way, the heuristic is recomputed before a state is expanded, leading to improved estimates that incorporate all knowledge gained from paths that were found after the state was inserted into the open list.

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
ClemensBuechner: This is also how I explained it to myself. 

I like the suggestion to add a document note, but I think "no, see document note about consistency" would be more accurate. My argument also relies on the assumption that when someone looks into the documentation to check whether a heuristic is consistent, they hope to find "consistent=yes" rather than "no". If they find "(yes), see document note ..." it is more tempting to skip the document note than when we say no.

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
ClemensBuechner: Did you check that this is what happens when setting lazy evaluators? I'm always confused because I think there are two different things described in the literature describing lazy evaluation. One fits to what you suggest to write here. The other thing is that states are inserted into the open list with the heuristic estimates of their parents rather than their own ones to avoid computing the heuristic for every generated state, but only compute the heuristic value once a state is expanded.

salome-eriksson: I agree the naming is unfortunate to say the least :). The second thing you mention is *lazy_search*. Lazy_evaluator is only implemented in eager_search, and the only thing it does different is that when a node is removed from the open list it will recompute the heuristic if necessary (checked with the h_dirty flag) and push the node back in open if the heuristic value changed.

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
salome-eriksson: Yeah that's probably better. I thought to add yes to convey that in some sense it is consistent, but the information that in some sense it's not is more imporant I agree.

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
ClemensBuechner: The constructor of the base class complained (legitimately) that it cannot call a virtual function before constructing the object. We discussed this offline and decided to add arguments to the constructor of the base class that tell it whether reasonable orderings, axioms, and conditional effects are supported by the concrete heuristic.

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
ClemensBuechner: OK, that sounds very familiar, then I'll adapt the comment as you suggested.

ClemensBuechner:
The last bunch of tests failed because I cancelled them since the change was very minor and had absolutely no influence on the tested code.

ClemensBuechner:
Here is a suggestion for the commit message:

[issue1071] Split LM-count heuristic in admissible and inadmissible case.

For users: The landmark count heuristic previously offered the boolean option `admissible`. If `false` (default), the heuristic value corresponds to the sum of costs to achieve all unreached or needed-again landmarks, which is generally not admissible. If `true`, a cost partitioning over the landmarks is used to get an admissible heuristic estimate. Since these two ideas are fundamentally different (only related in the fact that they are based on landmarks), we decided to refactor our code to represent the two distinct concepts appropriately.
If you use the `lmcount` heuristic in your calls to Fast Downward, you should check whether you need to adapt your command strings. Any occurrence of `admissible=(true|false)` should be removed and in case you used `admissible=true`, you also need to replace the heuristic `lmcount` by `lmcp`, i.e., use the landmark cost partitioning heuristic instead of the landmark count heuristic.
According to our experiments, the changes improve search time consistently but not significantly, both in satisficing planning with `lmcount` and in optimal planning with `lmcp`.
https://issues.fast-downward.org/issue1071.

silvansievers:
While this summarizes very well what this issue did, this sounds like a too verbose commit message to me. For instance, I wouldn't explain why the change makes sense and what the two classes conceptually have in common, but restrict the message to explain how the use of lmcount is changed, i.e., how users need to adapt their command line options. I also wouldn't mention experiments.

ClemensBuechner:
@silvansievers: I included some words on performance because I was asked to do so in the past. I'm fine with removing them but left them in for now until more people can comment on this.

Update:

[issue1071] Split LM-count heuristic in admissible and inadmissible case.

For users: We removed the boolean option `admissible` of the heuristic `lmcount`. No functionality is removed, though, as we introduce the `lmcp` heuristic that replaces `admissible=true`. If you use `lmcount` in your command strings, you should remove all occurrences of `admissible=(true|false)` and in case you used `admissible=true` you also need to replace the heuristic `lmcount` by `lmcp`, i.e., use the landmark cost partitioning heuristic instead of the landmark count heuristic.
According to our experiments, the changes improve search time consistently but not significantly, both in satisficing planning with `lmcount` and in optimal planning with `lmcp`.
https://issues.fast-downward.org/issue1071.

silvansievers:
Fine with me.

maltehelmert:
Regarding performance numbers: it's difficult to give hard and fast advice. This is the kind of change that sounds from outside essentially like a change in option syntax only, so as a user I would expect performance to be (essentially) unaffected and would not expect a comment if that's the case. From our release perspective, it's always easier to edit something down (removing a comment about performance) than to add information that wasn't present and thus needs to be rediscovered.

ClemensBuechner:
After some more changes and more experiments, we are in need of a new commit message. How about this one:

[issue1071] Split LM-count heuristic in admissible and inadmissible case.

For users: We replace the `lmcount` heuristic with two new heuristics called `landmark_sum` and `landmark_cost_partitioning`. No functionality is removed, though; `landmark_sum` replaces `lmcount` with the option `admissible=false` (changing the name because the heuristic sums over the landmark costs rather than simply counting landmarks) and `landmark_cost_partitioning` replaces `lmcount` with the option `admissible=true`.

PR #144, opened 2023-01-30
Issue1019

silvansievers:


src/search/pdbs/cegar.cc
```diff
@@ -226,9 +227,10 @@ bool CEGAR::time_limit_reached(
 unique_ptr<PatternInfo> CEGAR::compute_pattern_info(Pattern &&pattern) const {
     vector<int> op_cost;
     bool compute_plan = true;
-    shared_ptr<PatternDatabase> pdb =
-        make_shared<PatternDatabase>(task_proxy, pattern, op_cost, compute_plan, rng, use_wildcard_plans);
-    vector<vector<OperatorID>> plan = pdb->extract_wildcard_plan();
+    PatternDatabaseFactory pdb_factory(
```
remochristen, line 230: As mentioned in our quick discussion, it's not so nice that this usage differs from the others.

src/search/pdbs/pattern_database_factory.h
```diff
@@ -0,0 +1,213 @@
...
+    int get_size() const {
+        return num_states;
+    }
+
+    std::vector<std::vector<OperatorID>> && extract_wildcard_plan() {
+        return std::move(wildcard_plan);
+    };
+};
+
+extern std::shared_ptr<PatternDatabase> compute_pdb(
```
remochristen, line 204: This function caught my eye as potentially not being good style. I'm not sure if this is the case, and if yes how to best improve it.

src/search/pdbs/cegar.cc
silvansievers: I fixed this by having two functions to compute a PDB: one with plan and one without.

src/search/pdbs/pattern_database_factory.h
silvansievers: I think the next step (in a new issue) would be to get rid of the factory class and compute the PDB in a functional way. This makes it possibly simpler to group some of the computed data structures into new classes, such as `pattern`, `num_states`, and `hash_multipliers` into a function `Projection` used for perfect hashing.

src/search/pdbs/pattern_database.cc
```diff
@@ -1,360 +1,20 @@
...
-        int var_id = pattern[pattern_var_id];
-        VariableProxy var = variables[var_id];
-        int temp = state_index / hash_multipliers[pattern_var_id];
-        int val = temp % var.get_domain_size();
-        if (val != abstract_goal.value) {
-            return false;
-        }
-    }
-    return true;
+    Pattern &&pattern,
```
FlorianPommerening, line 10: I liked the old interface better because it prevented some wrong usage of the class. With the new interface, the values of `pattern`, `num_states`, `distances`, and `hash_multipliers` could disagree. So as a user of the class, once you pick the first constructor parameter, you have to put specific values in the other parameters. I see forced choices in parameters as a bit of a code smell but I don't have a better suggestion that also moves the computation out of the constructor..

src/search/pdbs/pattern_database_factory.h
```diff
@@ -0,0 +1,218 @@
...
+    const Pattern &get_pattern() const {
+        return pattern;
+    }
+
+    // Returns the size (number of abstract states) of the PDB
+    int get_size() const {
+        return num_states;
+    }
+
+    std::vector<std::vector<OperatorID>> && extract_wildcard_plan() {
```
FlorianPommerening, line 198: If this is called multiple times, the second time will fail.

FlorianPommerening (changes requested):
We discussed the general setup in person and decided to try a more functional approach where instead of the factory class, we have a rather small public interface of just functions that create the PDBs. The implementation of those functions than can pass around the necessary information internally (as functions static in the cc file).

src/cmake_modules/FastDownwardMacros.cmake
```diff
@@ -1,17 +1,21 @@
 include(CMakeParseArguments)
 
+macro(check_and_set_compiler_flag FLAG)
```
FlorianPommerening, line 3: This is now done on main, I guess the diff will go away on a rebase.

src/search/DownwardFiles.cmake
```diff
@@ -761,6 +761,7 @@ fast_downward_plugin(
         pdbs/pattern_collection_generator_multiple_random
         pdbs/pattern_collection_generator_multiple
         pdbs/pattern_collection_generator_systematic
+        pdbs/pattern_database_factory
```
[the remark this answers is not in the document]
FlorianPommerening: I don't remember what we discussed about this. If we decided against the rename, can you comment what the reason against this was and click "resolve conversation", so it doesn't show as an open discussion?

src/search/pdbs/cegar.cc
```diff
@@ -225,10 +226,8 @@ bool CEGAR::time_limit_reached(
 
 unique_ptr<PatternInfo> CEGAR::compute_pattern_info(Pattern &&pattern) const {
     vector<int> op_cost;
-    bool compute_plan = true;
-    shared_ptr<PatternDatabase> pdb =
-        make_shared<PatternDatabase>(task_proxy, pattern, op_cost, compute_plan, rng, use_wildcard_plans);
-    vector<vector<OperatorID>> plan = pdb->extract_wildcard_plan();
+    auto [pdb, plan] = compute_pdb_and_wildcard_plan(
```
FlorianPommerening, line 229: Much nicer this way!

src/search/pdbs/cegar.cc
```diff
@@ -225,10 +226,8 @@ bool CEGAR::time_limit_reached(
...
+        task_proxy, pattern, op_cost, rng, use_wildcard_plans);
```
FlorianPommerening, line 230: I'm still a bit confused with the naming. What does `compute_..._wildcard_plan` do with `use_wildcard_plans=false`?

src/search/pdbs/match_tree.h
```diff
@@ -22,11 +24,9 @@ namespace pdbs {
 */
 
 class MatchTree {
-    TaskProxy task_proxy;
+    const TaskProxy &task_proxy;
```
FlorianPommerening, line 27: Do we need a reference here? `TaskProxy` should be light-weight enough to copy and that would reduce the complexity. For example, in `Heuristic` we also store a `TaskProxy` not a reference. I think we agreed at some point to use const references for passing around proxies but I don't think they are necessary for storage.

src/search/pdbs/match_tree.h
```diff
@@ -22,11 +24,9 @@ namespace pdbs {
...
+    const Projection &projection;
```
FlorianPommerening, line 28: Having a reference here means that the `Projection` has to live longer than the `MatchTree`. This is probably fine but at least it should be documented.

FlorianPommerening: I think it would also be ok to copy it. If it just stores the pattern and multipliers, it should be light-weight enough.

FlorianPommerening (commented):
I looked through the classes I could review on Github. I have a meeting now and will continue with meld for the other files afterwards.

src/search/pdbs/pattern_database_factory.h
```diff
@@ -0,0 +1,55 @@
...
+  step.. That is, a wildcard plan induces a single trace of abstract states
+  like a regular plan, but offers a set of "equivalent" operators for each
+  plan step.
+
+  If compute_wildcard_plan is false, each set of operator IDs of the returned
+  plan contains exactly one operator, thus representing a regular plan. If
+  set to true, each set contains at least one operator ID.
+*/
+extern std::tuple<std::shared_ptr<PatternDatabase>,
+                  std::vector<std::vector<OperatorID>>> compute_pdb_and_wildcard_plan(
```
FlorianPommerening, line 47: I'm still not 100% convinced that this belongs in the factory. The use case is so specific and the data structure so unusual for a global interface (we don't normally use `vector<vector<OperatorID>>` for plans). I don't mind for this issue, as it was strongly coupled to the code before as well. In the long run, we should rethink this. For example, a more general option might be to return the transition system of the PDB and let the caller compute the wildcard plan. Having access to the transition system would be useful anyway, I implemented this in a couple of branches already.

FlorianPommerening (commented):
I added some more comments on the other files, but only minor things.

src/search/DownwardFiles.cmake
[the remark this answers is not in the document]
silvansievers: I think you wanted it to be renamed. I don't have a strong opinion so I'll do it.

src/search/pdbs/cegar.cc
silvansievers: I would like to change it back to `compute_pdb_and_plan` (which you didn't like initially) because I think that this is indeed confusing. If `use_wildcard_plans=false`, the method computes a regular plan, though still using the same data structure that you seem to dislike so much, by having the inner vector containing exactly one operator ID for each plan step.

src/search/pdbs/abstract_operator.cc
```diff
@@ -0,0 +1,36 @@
...
+    : concrete_op_id(concrete_op_id),
+      cost(cost),
+      regression_preconditions(move(regression_preconditions)),
+      hash_effect(hash_effect) {
+}
+
+void AbstractOperator::dump(const Pattern &pattern,
+                            const VariablesProxy &variables,
+                            utils::LogProxy &log) const {
+    log << "AbstractOperator:" << endl;
```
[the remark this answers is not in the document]
silvansievers: This was not intentionally. I think when updating the old pull request (which didn't have Log objects), I might have missed to have this if-clause. Currently, there is no caller of this method, so I think it is fine to leave it like this. To answer your question: I didn't look at or change other dump methods in the PDB code.

src/search/pdbs/match_tree.h
silvansievers: I don't like copying two vectors for no good reason. I'll document that Projection needs to live longer than MatchTree.

src/search/DownwardFiles.cmake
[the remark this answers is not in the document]
FlorianPommerening: My suggestion would be to rename it if it should take the role of a builder in the long term. Otherwise, we'll have to rename it later (either from builder to factory or the other way). If this is supposed to stay a private class, I think sticking with factory is fine.

src/search/pdbs/pattern_database_factory.cc
```diff
@@ -0,0 +1,469 @@
...
+            move(projection),
+            move(distances));
+    }
+
+    vector<vector<OperatorID>> && extract_wildcard_plan() {
+        return move(wildcard_plan);
+    };
+};
+
+static Projection compute_projection(
```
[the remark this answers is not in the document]
silvansievers: I agree. Changed.

src/search/pdbs/cegar.cc
FlorianPommerening: Fine with me. What I dislike is more the place where this code lives. I think this is a very specific computation that is useful for one particular use case and implemented in a general interface. But I'm fine with leaving it as is for this issue.

src/search/pdbs/match_tree.h
FlorianPommerening: In my opinion avoid a life-time dependency between two objects is a good reason to copy two vectors that we can expect to be small. Of course every vector could become so large it is a problem but specifically for patterns of PDBs, if the pattern is so large that copying is a problem, then we'll have other problems. We do not have many such life-time depencies in the code as far as I'm aware. There is one for the task that we accepted and work around by copying a shared pointer to the task everywhere where it should stay alive. There is another one with a log that recently caused problems.

src/search/pdbs/pattern_database.cc
```diff
@@ -1,372 +1,42 @@
...
-        int var_id = pattern[pattern_var_id];
-        VariableProxy var = variables[var_id];
-        int temp = state_index / hash_multipliers[pattern_var_id];
-        int val = temp % var.get_domain_size();
-        if (val != abstract_goal.value) {
-            return false;
-        }
-    }
-    return true;
+int Projection::unrank(int index, int var, int domain_size) const {
```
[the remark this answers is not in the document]
silvansievers: Projection could store the list of domain sizes of the variables in pattern. However, unranking is only required in MatchTree and the distance computation. Match tree nodes store the domain size to know how many successors they have. Since they use arrays, they need that information for their destruction as well. So, storing the domain sizes for the pattern variables would waste some memory once the PDB is computed, and the match tree would need to rely on the fact that the projection has the same domain sizes as those that it uses itself. I'm still not really opposed to the suggestion, as we recently implemented domain abstractions and there we did exactly that: store the domain sizes of all "pattern" variables. Do you think this is a valuable change?

src/search/pdbs/match_tree.h
silvansievers: I agree, that one should avoid this. However, I think both AbstractOperator and MatchTree are classes "internal" to the PDB factory. They are visible only because they are in separate files (which I think is good). Projection is stored as part of the PDB factory, which also in the constructor computes the match tree and uses it to finish computing the PDB. Of course using a match tree in a different context is "dangerous", but it is just not intended to be used for something different. An alternative could be to not store the reference, but pass projection to both public methods of match treee: insertion and retrieval of operators. Would you prefer this?

src/search/pdbs/abstract_operator.cc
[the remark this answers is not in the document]
FlorianPommerening: I'm fine either way. In the long run, I would prefer to implement dump methods consistently (either all checking the log level internally or all externally).

PR #153, opened 2023-02-13
Issue1080

remochristen:


FlorianPommerening:
We decided not to merge this. See https://issues.fast-downward.org/issue1080

PR #154, opened 2023-03-15
Issue1085

maltehelmert:


src/search/lp/lp_internals.cc
```diff
@@ -84,7 +76,7 @@ class ErrorCatchingCoinMessageHandler : public CoinMessageHandler {
 
     virtual void checkSeverity() {
         /*
-          Note that currentMessage_ should be used here but it doesn't help for clpex:
+          Note that currentMessage_ should be used here but it doesn't help for cplex:
```
FlorianPommerening, line 79: Since you are fixing this anyway, shouldn't this be CPLEX?

FlorianPommerening (approved):
Looks fine to me. As far as I know, these are all the places that need to be changed in the code. 

We also have some `#pragma GCC diagnostic ignored` around the includes of the OSI headers in `lp_internals.cc`. If some of those warnings would only be tripped in the Gurobi and CLP headers, we could remove the ignores for them. But I think they occur all over the place and they will also disappear together with OSI in issue1076, so probably not worth spending time on this here.

jendrikseipp:
Closing this PR since merging issue1076 made it redundant.

PR #155, opened 2023-03-27
Issue1030

PatrickFerber:


src/translate/pddl_parser/parsing_functions.py
```diff
@@ -1,31 +1,142 @@
...
+                            f"Type missing after '{TYPED_LIST_SEPARATOR}'.",
+                            alist)
+                    items = alist[:separator_position]
+                    _type = alist[separator_position + 1]
+                    alist = alist[separator_position + 2:]
+                    if not (isinstance(_type, str) or
+                           (_type and _type[0] == "either" and
+                            all(isinstance(_sub_type, str) for _sub_type in _type[1:]))):
+                        context.error("Type value is expected to be a single word "
+                                      "or '(either WORD*)")
```
salome-eriksson, line 132: This message was at first unclear to me. Maybe expand it to the following?

Type value is expected to be a single word, or a choice of words denoted by '(either word1 word2 ...)'

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -42,49 +153,111 @@ def set_supertypes(type_list):
...
+                              syntax=SYNTAX_CONDITION_NOT)
+            negated = not negated
     elif tag in ("forall", "exists"):
-        parameters = parse_typed_list(alist[1])
-        args = alist[2:]
-        assert len(args) == 1
+        if len(alist) != 3:
+            context.error("'forall' and 'exists' expect exactly two arguments.",
+                          syntax=SYNTAX_CONDITION_FORALL_EXISTS)
+        if not isinstance(alist[1], list) or not alist[1]:
```
salome-eriksson, line 234: I'm sure this works in python, but is there a reason why we don't first check for "existence" of the variable and only then whether or not it is a list.

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -188,121 +379,271 @@ def add_effect(tmp_effect, result):
...
+    if not alist:
+        context.error("All (sub-)effects have to be a non-empty blocks.", alist)
     tag = alist[0]
     if tag == "and":
-        return pddl.ConjunctiveEffect(
-            [parse_effect(eff, type_dict, predicate_dict) for eff in alist[1:]])
+        effects = []
+        for eff in alist[1:]:
+            if not isinstance(eff, list):
+                context.error("All sub-effects of a conjunction have to be blocks.",
```
salome-eriksson, line 391: Why not use `expected_list_error`?

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -188,121 +379,271 @@ def add_effect(tmp_effect, result):
...
     elif tag == "when":
-        assert len(alist) == 3
-        condition = parse_condition(
-            alist[1], type_dict, predicate_dict)
-        effect = parse_effect(alist[2], type_dict, predicate_dict)
+        if len(alist) != 3:
+            context.error("'when' effect expects exactly two arguments.",
+                          syntax=SYNTAX_EFFECT_WHEN)
+        if not isinstance(alist[1], list):
+            context.error(
```
salome-eriksson, line 415: Why not use `expected_list_error`?

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -188,121 +379,271 @@ def add_effect(tmp_effect, result):
...
+        if isinstance(exp, list):
+            if len(exp) < 1:
+                context.error("Expression cannot be an empty block.",
+                              syntax=SYNTAX_EXPRESSION)
+            functionsymbol = exp[0]
+            return pddl.PrimitiveNumericExpression(functionsymbol, exp[1:])
+        elif exp.replace(".", "").isdigit() and exp.count(".") <= 1:
+            return pddl.NumericConstant(float(exp))
+        elif exp[0] == "-":
+            context.error("Expression cannot be a negative number",
```
salome-eriksson, line 448: I liked the old phrasing "Negative numbers are not supported" better, mainly because it does not use the word "expression" and thus is still understandable if you're not too familiar with PDDL terminology.

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -313,180 +654,164 @@ def parse_task(domain_pddl, task_pddl):
...
+        for entry in iterator:
+            if isinstance(entry, list) and entry[0] == ":metric":
+                with context.layer("Parsing metric"):
+                    if len(entry) != 3 or not isinstance(entry[2], list) or len(entry[2]) != 1 or entry[1] != "minimize" or entry[2][0] != "total-cost":
+                        context.error("Invalid metric definition.", entry, syntax=SYNTAX_METRIC)
+                    use_metric = True
+        yield use_metric
+
+        for _ in iterator:
+            assert False, "This line should be unreachable"
```
salome-eriksson, line 802: This can happen when we have malformed PDDL, i.e. when someone defines more things after metric, correct? Then we should give an error message.

As an aside, if someone does not specify a metric but other things, do they get swallowed and the PDDL accepted? I will test this.

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -1,31 +1,142 @@
...
+
+    def error(self, message, item=None, syntax=None):
+        error_msg = f"{self}\n{message}"
+        if syntax:
+            error_msg += f"\nSyntax: {syntax}"
+        if item:
+            error_msg += f"\nGot: {item}"
+        raise ParseError(error_msg)
+
+    def expected_word_error(self, name, *args, **kwargs):
```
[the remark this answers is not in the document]
ClemensBuechner: I agree that it would be nicer to have a function do the check and yielding the error (if necessary). This applies to all three cases `expected_word_error`, `expected_list_error`, and `expected_name_block_error`. I also checked the calls to these functions and they all have similar if-clauses preceding them, so it shouldn't be a problem to unify this in a function.

If I understood Malte's comment in the daily sprint meeting correctly, they should preferrably be named `check...` rather than `assert...` because assert should be reserved for code errors rather than user input errors.

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -1,31 +1,142 @@
...
+    @contextlib.contextmanager
+    def layer(self, message: str):
+        self._traceback.append(message)
+        yield
+        assert self._traceback.pop() == message
+
+
+def check_named_block(alist, names):
+    return isinstance(alist, list) and alist and alist[0] in names
+
```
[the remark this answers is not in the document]
ClemensBuechner: Sounds good to me.

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -1,31 +1,142 @@
...
+def assert_named_block(context, alist, names):
+    if not check_named_block(alist, names):
+        context.expected_named_block_error(alist, names)
+
+def construct_typed_object(context, name, _type):
```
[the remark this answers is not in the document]
ClemensBuechner: I agree with your point that `_type` should also be tested. After thinking about it a little bit, I don't see an issue with just doing the checks outside the function. Sure, one could argue that it makes sense to test this whenever trying to construct an object. However, within the constructing function, it would make more sense to assert this in my opinion.

src/translate/pddl_parser/parsing_functions.py
ClemensBuechner: Now from my digging in the issue tracker yesterday it sounded like we don't support `either`. Here I see the opposite and from the code it seems it's supported. We might need to clean that up in the issue tracker then.

That being said, I prefer your formulation.

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -1,31 +1,142 @@
...
+                    _type = alist[separator_position + 1]
+                    alist = alist[separator_position + 2:]
+                    if not (isinstance(_type, str) or
+                           (_type and _type[0] == "either" and
+                            all(isinstance(_sub_type, str) for _sub_type in _type[1:]))):
+                        context.error("Type value is expected to be a single word "
+                                      "or '(either WORD*)")
+                for item in items:
+                    if only_variables and not item.startswith("?"):
+                        context.error("Expected item to be a variable", item)
```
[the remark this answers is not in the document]
ClemensBuechner: Sounds good.

src/translate/pddl_parser/parsing_functions.py
ClemensBuechner: I can imagine that `not alist[1]` is there to prevent empty lists, so even if `alist[1]` is a list, we want to throw the error if that list is empty. Maybe it makes sense to split the error message, though. First check that `alist[1]` is indeed a list and complain about this if it is not the case, and then check that the list is not empty and complain differently if this doesn't hold. Or did I misunderstand something and a "non-empty block" can also be something different than a list?

src/translate/pddl_parser/parsing_functions.py
ClemensBuechner: Yeah, seems a bit inconsistent. Though, this gives clearer instructions where to look for the faulty input. would the context-layer provide enough info about this as well? Then I don't see an issue with turning this into `expected_list_error`.

src/translate/pddl_parser/parsing_functions.py
ClemensBuechner: Sounds reasonable, I support changing it back.

src/translate/pddl_parser/parsing_functions.py
ClemensBuechner: I can neither confirm that this can happen nor do I understand the context to meaningfully comment on this.

src/translate/pddl_parser/parsing_functions.py
maltehelmert: For what it's worth, I think we don't support fractional numbers either. I think it's a mistake that we accept ".".

src/translate/translate.py
```diff
@@ -753,3 +754,6 @@ def handle_sigxcpu(signum, stackframe):
         traceback.print_exc(file=sys.stdout)
         print("=" * 79)
         sys.exit(TRANSLATE_OUT_OF_MEMORY)
+    except pddl_parser.ParseError as e:
+        print(e)
+        sys.exit(TRANSLATE_INPUT_ERROR)
```
maltehelmert, line 759: This changes the meaning of `TRANSLATE_INPUT_ERROR` compared to what we say on the wiki ("Usage error: wrong command line options"). If nobody things we need two different error codes for command-line errors vs. input file errors, the code can stay is is, but then we should update the wiki documentation to cover both cases.

src/translate/pddl_parser/parsing_functions.py
FlorianPommerening: I would also read this as "if the element is no list or an empty list". The order of the tests makes sense to me: the first check is only passed if the object is a list (so for example, `None` wouldn't pass the first check). Then the second test checks if the list is empty. Flipping the checks would first cast the object to a boolean from whatever type it is and check if it is "truthy" before checking if that truthiness means what we expect (checking if list is empty.)

src/translate/pddl_parser/parsing_functions.py
FlorianPommerening: I ran this and found three issues:
```
(
   ...
    (:goal ...  )
    (foo)
    (:metric minimize (total-cost))
    (bar)
)
```
Anything after `(:goal ...)` other than `(:metric ...)` is silently ignored. The assert actually never triggers because the loop above consumes all elements. But there shouldn't be a loop to ignore `(foo)` to begin with, right?

The other issue is this with this code:
```
(
   ...
    (:goal ...  )
    (:metric minimize (total-cost))
    ()
)
```
where the check `entry[0] == ":metric"` fails.

src/translate/pddl_parser/parsing_functions.py
maltehelmert: In Python, if x is a collection, "if x" is the idiomatic way to test for non-emptiness. Splitting this into two tests with separate errors along the lines of "expected a list", "list cannot be empty" would make sense, and I agree they only make sense in this order.

src/translate/pddl/tasks.py
```diff
@@ -68,15 +68,22 @@ def dump(self):
             for axiom in self.axioms:
                 axiom.dump()
 
+
+REQUIREMENT_LABELS = [
```
victorpaleologue, line 72: Since Python 3.8, there's a notion of `Final`: https://peps.python.org/pep-0591/

src/translate/pddl_parser/lisp_parser.py
```diff
@@ -1,20 +1,18 @@
...
-        return self.value
+from .parse_error import ParseError
 
 # Basic functions for parsing PDDL (Lisp) files.
 def parse_nested_list(input_file):
     tokens = tokenize(input_file)
     next_token = next(tokens)
     if next_token != "(":
-        raise ParseError("Expected '(', got %s." % next_token)
+        raise ParseError("Expected '(', got '%s'." % next_token)
```
victorpaleologue, line 10: switch to f-string?

src/translate/pddl_parser/pddl_file.py
```diff
@@ -14,10 +15,10 @@ def parse_pddl_file(type, filename):
         return lisp_parser.parse_nested_list(file_open(filename,
                                                        encoding='ISO-8859-1'))
     except OSError as e:
-        raise SystemExit("Error: Could not read file: %s\nReason: %s." %
+        raise SystemExit("Error: Could not read file: %s\nReason: %s" %
                          (e.filename, e))
-    except lisp_parser.ParseError as e:
-        raise SystemExit("Error: Could not parse %s file: %s\nReason: %s." %
+    except parse_error.ParseError as e:
+        raise parse_error.ParseError("Error: Could not parse %s file: %s\nReason: %s" %
```
victorpaleologue, line 21: or, in a more idiomatic way:

```python
except parse_error.ParseError as e:
        raise parse_error.ParseError(f"Error: Could not parse {type} file: {filename}") from e
```

salome-eriksson (commented):
These comments are from notes I made when testing different ways to specify wrong PDDLs. Unfortunately it was a while ago and I don't always remember the details.
I think basically all of my comments are not handled better by the current code, meaning we can also merge issue1030 as is and tackle these comments at a later time.

maltehelmert:
[issue1030] translator: Improve PDDL error reporting.

We have substantially revised the PDDL syntax checks in the translator. Many PDDL errors are now diagnosed much better and lead to a meaningful error message and exit code where previously they would trigger failed assertions or be silently accepted with unclear semantics.

PR #156, opened 2023-06-27
issue1088

ClemensBuechner:
Assert that LP constraints contain every variable at most once.


src/search/lp/lp_solver.cc
```diff
@@ -37,6 +37,14 @@ using namespace std;
 using utils::ExitCode;
 
 namespace lp {
+#ifndef NDEBUG
+template<class T>
+static bool all_values_are_unique(const vector<T> &v) {
+    set<T> s(v.begin(), v.end());
```
jendrikseipp, line 43:
```suggestion
    unordered_set<T> s(v.begin(), v.end());
```
since we don't require a sorted container.

src/search/lp/lp_solver.cc
```diff
@@ -37,6 +37,14 @@ using namespace std;
 using utils::ExitCode;
 
 namespace lp {
+#ifndef NDEBUG
+template<class T>
+static bool all_values_are_unique(const vector<T> &v) {
```
jendrikseipp, line 42: Move this function next to sorted_unique?

jendrikseipp: I have no strong opinion.

src/search/lp/lp_solver.cc
```diff
@@ -32,11 +32,20 @@
 #include <cassert>
 #include <iostream>
 #include <numeric>
+#include <unordered_set>
 
 using namespace std;
 using utils::ExitCode;
 
 namespace lp {
+#ifndef NDEBUG
```
FlorianPommerening, line 41: Do you get an "unused function" error without this? If not, I'd reduce the number of preprocessor directives and just always define the function (it shouldn't have a performance impact if it is not used). I have no strong opinion, though.

src/search/lp/lp_solver.cc
FlorianPommerening: You currently also have an additional import here (for `unordered_set`). I think treat both the algorithm and the utils namespace as something that may be included everywhere. But same as Jendrik, I'm also fine either way. It certainly has an advantage to make the module as independent from other modules as possible and we could always make the function more public later on if we need it somewhere else.

ClemensBuechner: Indeed, I overlooked that because I first used `set` which didn't need the extra import. ;) 

The disadvantage of defining it locally is that somebody else might define something similar locally unaware of this function here. So I'll just move it to `utils` until somebody has a stronger opinion in a different direction.

src/search/lp/lp_solver.cc
ClemensBuechner: No, I thought that's something we do if something's used only in debug mode. I'll remove it.

PR #157, opened 2023-07-16
Issue1090 (use uncrustify 0.72)

jendrikseipp:


src/search/algorithms/segmented_vector.h
```diff
@@ -182,7 +182,7 @@ class SegmentedArrayVector {
         : elements_per_array((assert(elements_per_array_ > 0),
                               elements_per_array_)),
           arrays_per_segment(
-              std::max(SEGMENT_BYTES / (elements_per_array * sizeof(Element)), size_t(1))),
+              std::max(SEGMENT_BYTES / (elements_per_array * sizeof(Element)), static_cast<size_t>(1))),
```
FlorianPommerening, line 185: This looks better, but is it something the style checks should do?

src/search/lp/lp_internals.cc
```diff
@@ -118,12 +118,12 @@ unique_ptr<OsiSolverInterface> create_lp_solver(LPSolverType solver_type) {
         break;
     case LPSolverType::CPLEX:
 #ifdef COIN_HAS_CPX
-        {
-            OsiCpxSolverInterface *cpx_solver = new OsiCpxSolverInterface;
-            CPXsetintparam(cpx_solver->getEnvironmentPtr(), CPX_PARAM_THREADS, 1);
-            cpx_solver->passInMessageHandler(new ErrorCatchingCoinMessageHandler);
-            lp_solver = cpx_solver;
-        }
+    {
```
FlorianPommerening, line 121: Are these braces still necessary? I think they were a hack around the old version of uncrustify. If so, we should remove them at this point.

src/search/lp/lp_internals.cc
```diff
@@ -137,11 +137,11 @@ unique_ptr<OsiSolverInterface> create_lp_solver(LPSolverType solver_type) {
         break;
     case LPSolverType::SOPLEX:
 #ifdef COIN_HAS_SPX
-        {
-            OsiSpxSolverInterface *spx_solver = new OsiSpxSolverInterface;
-            spx_solver->getSPxOut()->setVerbosity(soplex::SPxOut::ERROR);
-            lp_solver = spx_solver;
-        }
+    {
```
FlorianPommerening, line 140: ... and here

src/search/pdbs/pattern_database_factory.cc
```diff
@@ -448,7 +448,7 @@ compute_pdb_and_plan(
     bool compute_wildcard_plan) {
     PatternDatabaseFactory pdb_factory(task_proxy, pattern, operator_costs, true, rng, compute_wildcard_plan);
     return {
-               pdb_factory.extract_pdb(), pdb_factory.extract_wildcard_plan()
+        pdb_factory.extract_pdb(), pdb_factory.extract_wildcard_plan()
```
FlorianPommerening, line 451: This should be a single line now.

FlorianPommerening (commented):
I left some comments on the code changes. Mostly places that were strange before and could now be formatted better. We should change them now because afterwards uncrustify will no longer point them out.

src/search/algorithms/segmented_vector.h
jendrikseipp: Uncrustify produces `size_t (1)`, note the space. Do you prefer this? I'm indifferent.

FlorianPommerening: That is worse, but the question remains in both cases: should a style checker introduce casts?

src/search/lp/lp_internals.cc
```diff
@@ -118,12 +118,10 @@ unique_ptr<OsiSolverInterface> create_lp_solver(LPSolverType solver_type) {
         break;
     case LPSolverType::CPLEX:
 #ifdef COIN_HAS_CPX
-        {
-            OsiCpxSolverInterface *cpx_solver = new OsiCpxSolverInterface;
-            CPXsetintparam(cpx_solver->getEnvironmentPtr(), CPX_PARAM_THREADS, 1);
-            cpx_solver->passInMessageHandler(new ErrorCatchingCoinMessageHandler);
-            lp_solver = cpx_solver;
-        }
+        OsiCpxSolverInterface *cpx_solver = new OsiCpxSolverInterface;
```
FlorianPommerening, line 121: I'm sorry, but I think I remember now why we need the braces here and it has nothing to do with style hacks or uncrustify: If only one of the solvers is installed, there was an issue with variable scopes. I don't see a problem currently, so maybe this was fixed by the `#ifdef`s. The tests compile with and without CPLEX, right? So if anything goes wrong, we'll catch it there.

src/search/algorithms/segmented_vector.h
maltehelmert: Would "1U" a better option that doesn't cause warnings or style-checker complaints here?

jendrikseipp: Yes, let's use "1U" here. To clarify, **I** introduced the static_cast, not uncrustify.

FlorianPommerening: `1U` sounds good. What did uncrustify suggest here (or did you just stumble over this dynamic cast and wanted to clean it up)?

maltehelmert: Jendrik wrote that uncrustify wanted to insert a space: `size_t (1)`.

jendrikseipp:
@FlorianPommerening The curly brackets are still required for the LP code to compile.

PR #168, opened 2023-08-02
Issue1089

ClemensBuechner:


src/search/landmarks/landmark_graph.h
```diff
@@ -25,11 +25,10 @@ enum class EdgeType {
...
     */
-    NECESSARY = 4,
-    GREEDY_NECESSARY = 3,
-    NATURAL = 2,
-    REASONABLE = 1,
-    OBEDIENT_REASONABLE = 0
+    NECESSARY = 3,
+    GREEDY_NECESSARY = 2,
+    NATURAL = 1,
+    REASONABLE = 0
```
salome-eriksson, line 31: This could theoretically cause problems if the code refers to the number instead of the name anywhere, but I certainly hope this is not the case. I quickly grepped for 4/3/2/1/0 and did not notice anything (of course quite possible that I would have missed it).

ClemensBuechner: I agree it's a bit risky but I've talked about this with Malte and we think it's best to start the counting at 0. There's certainly some clauses like `if (order_type >= EdgeType::NATURAL)` but a simple shift does clearly not affect these cases. 
Now we're at least two who tried to find further usages but weren't able to. :)

src/search/landmarks/landmark_factory_reasonable_orders_hps.h
```diff
@@ -10,13 +10,12 @@ class LandmarkFactoryReasonableOrdersHPS : public LandmarkFactory {
     virtual void generate_landmarks(const std::shared_ptr<AbstractTask> &task) override;
 
     void approximate_reasonable_orders(
-        const TaskProxy &task_proxy, bool obedient_orders);
+        const TaskProxy &task_proxy);
```
maltehelmert, line 13: Line break no longer needed.

src/search/landmarks/landmark_graph.h
maltehelmert: It's never acceptable to use numbers instead of symbolic names for constants like these, so I wouldn't worry about the possibility that we do this anywhere in the code. If you want, we can also remove the numbers entirely and rely on the automatic numbers of enum values. But then we have to reverse the order, i.e., list them in the order REASONABLE, NATURAL, ... because numbering goes from 0 onwards.

PR #172, opened 2023-08-23
issue996

ClemensBuechner:


src/search/landmarks/landmark_factory.cc
```diff
@@ -98,25 +98,14 @@ bool LandmarkFactory::is_landmark_precondition(
 
 void LandmarkFactory::edge_add(LandmarkNode &from, LandmarkNode &to,
                                EdgeType type) {
-    /* Adds an edge in the landmarks graph if there is no contradicting edge (simple measure to
-    reduce cycles. If the edge is already present, the stronger edge type wins.
-    */
+    /* Adds an edge in the landmarks graph. If an edge between the same
+       landmarks is already present, the stronger edge type wins. */
+    // TODO: Do we have a strong argument why self-loops should be prevented?
```
salome-eriksson, line 103: Remove this and next TODO as discussed offline.

salome-eriksson (approved):

src/search/landmarks/landmark_heuristic.cc
```diff
@@ -70,6 +79,42 @@ void LandmarkHeuristic::initialize(const plugins::Options &opts) {
...
+    int num_landmarks = lm_graph->get_num_landmarks();
+    vector<bool> closed(num_landmarks, false);
+    vector<bool> visited(num_landmarks, false);
+    for (auto &node : lm_graph->get_nodes()) {
+        if (depth_first_search_for_cycle_of_natural_orderings(
+                *node, closed, visited)) {
+            return true;
+        }
+    }
+    return false;
```
maltehelmert, line 92: Is this equivalent to the old code? It's wasteful because it is O(N^2) even though the problem has simple O(N) algorithms.

If the old code did this too, we can leave it as is, but I would at least add a TODO and add this as an open item in our landmark metaissue.

maltehelmert: (...or perhaps it is already O(N) because of the way closed and visited are handled by the called function as two-way arguments, which I didn't check. That's probably the case. In that case the code is perhaps not as pretty as it could be, but OK performance-wise.)

src/search/landmarks/landmark_heuristic.cc
```diff
@@ -70,6 +79,42 @@ void LandmarkHeuristic::initialize(const plugins::Options &opts) {
...
+}
+
+bool LandmarkHeuristic::depth_first_search_for_cycle_of_natural_orderings(
+    const LandmarkNode &node, vector<bool> &closed, vector<bool> &visited) {
+    int id = node.get_id();
+    if (closed[id]) {
+        return false;
+    } else if (visited[id]) {
+        // Cycle detected.
```
maltehelmert, line 101: Is this a useful comment? I think "return true" already communicates this.

src/search/landmarks/landmark_heuristic.cc
```diff
@@ -70,6 +79,42 @@ void LandmarkHeuristic::initialize(const plugins::Options &opts) {
...
+        if (child.second >= EdgeType::NATURAL) {
+            if (depth_first_search_for_cycle_of_natural_orderings(
+                    *child.first, closed, visited)) {
+                return true;
+            }
+        }
+    }
+    closed[id] = true;
+    return false;
+}
```
maltehelmert, line 116: OK, so this is all O(N). Please ignore my previous comment; apologies. And I think the code is pretty enough; the only thing I somewhat dislike is that it would be nice to have this as a generic graph algorithm and then have the landmark code use this, but that requires some design work I wouldn't bother with at this point.

src/search/landmarks/landmark_heuristic.cc
```diff
@@ -61,6 +61,15 @@ void LandmarkHeuristic::initialize(const plugins::Options &opts) {
...
         opts.get<bool>("prog_gn"), opts.get<bool>("prog_r"));
 
+    /* TODO: Maybe deal with this differently, e.g., mark initial state
+        as dead-end. (Similar use case in M&S maybe?) */
+    if (landmark_graph_has_cycle_of_natural_orderings()) {
+        log << "Found a cycle of natural (or stronger) landmark orderings. By "
+            << "the definition of natural orderings, the problem is unsolvable."
+            << endl;
+        utils::exit_with(utils::ExitCode::SEARCH_UNSOLVABLE);
+    }
```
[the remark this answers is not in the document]
ClemensBuechner: When working on the implementation draft, we asked ourselves whether we should exit immediately or do it some other way. The reason we went with this option is because we found a similar usage of `ExitCode::SEARCH_UNSOLVABLE` in `pdbs/cegar.cc`. I don't think we should add more of these if we don't want them in the long term and I see your points why we don't want them. Using a flag instead is a straightforward alternative that I can implement right away.

src/search/landmarks/landmark_heuristic.cc
ClemensBuechner: I was thinking about a more generic implementation but decided against it exactly because I think it is beyond the scope of this issue. Should I add a TODO or open a new issue right away?

maltehelmert: No, I don't think it's worth it. We can leave this as is.

src/search/landmarks/landmark_heuristic.cc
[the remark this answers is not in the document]
ClemensBuechner: Only because there is a cycle in the initial state doesn't mean the heuristic value is infinite for *every* state. I'm not sure whether this is relevant or not, but returning `DEAD_END` in `compute_heuristic(...)` for arbitrary states if there is a cycle of natural orderings in the initial state feels wrong to me. But was this your suggestion, anyway? Happy to implement it this way even if I'm not the biggest fan of the solution.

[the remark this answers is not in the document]
maltehelmert: The discussion is perhaps complex enough that I suggest we talk on Zoom. Are you available today some time between 16:45 and 17:15? In this case, can you call me on Zoom?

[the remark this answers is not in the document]
ClemensBuechner: We've had a discussion offline. I've implemented the flag-approach and added a somewhat extensive comment to reflect our discussion. Maybe it is unnecessary or too verbose, I don't know. I'm also fine with just dropping it, but I thought it might make sense to document somewhere what we discussed.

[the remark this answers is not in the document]
maltehelmert: I had a look at the comment and like it.

src/search/landmarks/landmark_heuristic.cc
```diff
@@ -152,9 +190,37 @@ void LandmarkHeuristic::generate_preferred_operators(
...
+       detect unsolvability due to cycles of natural orderings or other reasons?
+       How is this signaled to the heuristic(s)? Is adding an unsatisfiable
+       landmark to the landmark graph an option? But this requires changing the
+       landmark graph after construction which we try to avoid at the moment on
+       the implementation side.
+    */
+    if (initial_landmark_graph_has_cycle_of_natural_orderings) {
+        log << "Found a cycle of natural (or stronger) landmark orderings. By "
+            << "the definition of natural orderings, the problem is unsolvable."
+            << endl;
```
maltehelmert, line 217: I am not sure I would want to make this output here. I suggest just returning infinity, but possibly including an output along the lines of this in the code that computes the flag. I find output in a "preprocessing" stage more palatable than "inside" the heuristic computation. In general, at least in future code, I'd also always guard output like this with a verbosity test, but I didn't check how we do it with other landmark-related outputs.

src/search/landmarks/landmark_heuristic.cc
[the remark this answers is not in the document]
ClemensBuechner: Thanks! :)

src/search/landmarks/landmark_heuristic.cc
ClemensBuechner: Sure, I'll remove the comment in the heuristic. You write "including an output along the lines of this", so is there something in particular you would change about it?

maltehelmert: It could be briefer, but I don't feel strongly about this. I might just write something along the lines of "Found a cycle of natural landmark orderings." and nothing more. That the problem is unsolvable as a consequence will be printed anyway by the search algorithm, I think.

ClemensBuechner: I see, sure, makes sense.

PR #174, opened 2023-09-01
Test all portfolio configs.

jendrikseipp:


driver/tests.py
```diff
@@ -59,6 +61,42 @@ def test_portfolios():
...
+        config[index] = part
+    return config
+
+
+def test_portfolio_configs():
+    all_configs = set()
+    for portfolio in PORTFOLIOS.values():
+        configs = _get_portfolio_configs(Path(portfolio))
+        all_configs |= set(tuple(_convert_to_standalone_config(config)) for config in configs)
+    for config in set(all_configs):
```
FlorianPommerening, line 96: Do you need `set(...)` here?

jendrikseipp: Apparently yes.

FlorianPommerening: Why? Isn't `all_configs` a set already? It is not modified in the loop, so I don't see why you'd need a copy.

jendrikseipp: Ah, I thought you mean a different line. Thanks, I removed it.

PR #175, opened 2023-09-01
Add satisficing FDSS 2023 portfolio.

jendrikseipp:


driver/portfolios/seq_opt_fdss_2023.py
```diff
@@ -0,0 +1,79 @@
...
+            if line == "begin_rule":
+                has_axioms = True
+
+            if line == "begin_operator":
+                in_op = True
+            elif line == "end_operator":
+                in_op = False
+            elif in_op:
+                parts = line.split()
+                if len(parts) >= 6 and all(p.lstrip('-').isdigit() for p in parts):
```
FlorianPommerening, line 64: This seems a bit fragile to me. For example, I think `1 2 3 4 5 6` would be a legal operator name that would be considered as a conditional effect here.

jendrikseipp: True. I'm now skipping lines with operator names. Is that enough or do you have a better solution?

FlorianPommerening: I generally think it is strange to write a poor-man's SAS^+ parser inside a portfolio file. Isn't deciding whether a task has conditional effects/axioms a problem that we will face with other portfolios as well? If so, the code for it shouldn't live inside the stone soup 2023 portfolio or be replicated in all portfolios. If the format changes, we don't want to adapt multiple parsers.

Independent of where the code lives, I would prefer a full parser to something specialized that just checks for specific features. Something like
```Python
task = parse_task(sas_file)
if task.has_conditional_effects() or task.has_axioms():
    whatever()
```

The current code also hard-codes the input file as `output.sas` and the users could pick a different name for the sasfile, couldn't they?


Another option could be to go the route we did for unit cost tasks, where we have `--is-unit-cost` as a flag that is handled by the search component. With similar flags for `--has-axioms` and `--has-conditional-effects` we could leave the portfolio file static and just return a fixed configuration.

Maybe we should move this discussion to the tracker as well? Or to a Fast Downward meeting?

driver/tests.py
```diff
@@ -72,16 +83,19 @@ def _get_portfolio_configs(portfolio: Path):
...
+            for _, config in value:
+                yield config
 
 
 def _convert_to_standalone_config(config):
     replacements = [
         ("H_COST_TRANSFORM", "no_transform()"),
         ("S_COST_TYPE", "normal"),
         ("BOUND", "infinity"),
+        ("lpsolver=cplex", "lpsolver=soplex"),  # CPLEX is not available in all CI runners.
```
FlorianPommerening, line 98: Isn't this unexpected for the user? If we say "we test all parts of all portfolio configurations" and the portfolio uses CPLEX but we test SoPlex instead, the config could still fail with CPLEX.

driver/portfolios/seq_opt_fdss_2023.py
maltehelmert: I suggest going for something along the lines of the current "--if-unit-cost" option. We wanted to modernize this to an expression using the new option parser, has this already happened? If not, we can use what these portfolios needs as a new use case for designing the feature. So far we planned to keep it simple, along the lines of an "ifunitcost(config1, config2)" expression. (Perhaps with underscores: if_unit_cost.) I don't know where this discussion currently lives on the tracker, if anywhere.

maltehelmert: Perhaps we can do all this more easily once we have librarified the planner more. For the time being, one alternative could be a planner mode/option that just evaluates some tests (does the task have axioms etc.) and then returns an exit code signaling the result. Then the portfolio code could leverage the existing code in the planner, but still do its own conditionals.

driver/tests.py
jendrikseipp: True, I removed this again. Now LP configs are simply skipped on macOS.

driver/portfolios/seq_opt_fdss_2023.py
jendrikseipp: That sounds like we're resurrecting the old Rube Goldberg machine ;-) I'm reluctant about replacing a simple hack with a complicated hack, especially if a better solution (the Python interface) is on the horizon. I'd just like to have the simple version of this merged, since we already got an inquiry of someone who wants to use the FDSS portfolios within the latest codebase.

jendrikseipp: What's your stance on this, Florian and Malte?

FlorianPommerening: From my side, whatever implementation we use to get to the information shouldn't live inside the portfolio script. The hard-coded `output.sas` is an indication that this is wrong place for it.

If we generally want to support portfolios that have different configurations based on things like axioms and conditional effects, then maybe we should pass this information into the portfolio script somehow. Alternatively, we could say that a portfolio is by definition a static list of timed configurations (as the other portfolios are) and introduce a new concept above them that does the switching. For example, a "DynamicPortfolio" (or a better name) could take in a task and return a portfolio (list of timed configurations).

jendrikseipp: OK, then let's add only the satisficing portfolio. Does the rest of the patch look good to merge?

driver/tests.py
```diff
@@ -27,16 +27,27 @@ def translate():
...
     subprocess.check_call(cmd, cwd=REPO_ROOT_DIR)
 
+# We need to translate the example task when this module is imported to have the
+# SAS+ file ready for the @pytest.parametrize function below. Translating the
+# task in setup_module() does not work here, because the @pytest.parametrize
+# decorator is executed before setup_module() is called. An alternative would be
+# to use a conftest.py file and call translate() in a pytest_sessionstart()
+# function, but that adds another file and leads to dumping the translator
+# output to the terminal.
+translate()
```
FlorianPommerening, line 37: Do we need to call `translate()` at all? The function `run_driver()` will translate the task anyway, won't it? It calls `fast-downward.py` with `--keep`, which should translate the task and keep the sas file, so I assume parameters contain the PDDL files and we need the sas-file for something later on? In the previous code, this looks different: we first called `translate()` to get the sas file and then called the planner, implying that the planner is called on the sas file.

In general, what are the changes in this file about? It seems unrelated to the issue. I don't necessarily mind but I don't get what problem this is fixing.

driver/portfolios/seq_opt_fdss_2023.py
FlorianPommerening: I'm fine with adding the portfolio, I don't get the changes in the tests.

You could also add the optimal portfolios individually, e.g. add
* `seq_sat_fdss_2023.py` for the satisficing portfolio,
* `seq_opt_fdss_2023_axioms.py` for the optimal portfolio in the presence of axioms,
* `seq_opt_fdss_2023_condeffs.py` for the optimal portfolio in the presence of conditional effects, and
* `seq_opt_fdss_2023_strips.py` for the optimal portfolio for STRIPS tasks.

maltehelmert: I can have a look at this later, but are you sure you don't want to consider having the search component look at the task properties? I don't agree that this is any way like the old Rube Goldberg code. It's already the case in the portfolio code that we repeatedly call the search, "dry-run" features are not at all unusual, and we'd use the task parsing code we already have. Also, having a scriptable test of task properties like this would be something I could see other people using as well in their pipelines.

Here is an example diff for the search code that allows us to test whether the task has conditional effects with `--test-conditional-effects'):

```
 git diff | cat
diff --git a/src/search/command_line.cc b/src/search/command_line.cc
index 5b95ff424..64748f198 100644
--- a/src/search/command_line.cc
+++ b/src/search/command_line.cc
@@ -2,14 +2,18 @@
 
 #include "plan_manager.h"
 #include "search_algorithm.h"
+#include "task_proxy.h"
 
 #include "parser/lexical_analyzer.h"
 #include "parser/syntax_analyzer.h"
 #include "plugins/any.h"
 #include "plugins/doc_printer.h"
 #include "plugins/plugin.h"
+#include "task_utils/task_properties.h"
+#include "tasks/root_task.h"
 #include "utils/logging.h"
 #include "utils/strings.h"
+#include "utils/system.h"
 
 #include <algorithm>
 #include <sstream>
@@ -149,6 +153,13 @@ static shared_ptr<SearchAlgorithm> parse_cmd_line_aux(const vector<string> &args
             num_previously_generated_plans = parse_int_arg(arg, args[i]);
             if (num_previously_generated_plans < 0)
                 input_error("argument for --internal-previous-portfolio-plans must be positive");
+        } else if (arg == "--test-conditional-effects") {
+            TaskProxy task_proxy(*tasks::g_root_task);
+            if (task_properties::has_conditional_effects(task_proxy)) {
+                exit(static_cast<int>(utils::ExitCode::SEARCH_UNSUPPORTED));
+            } else {
+                exit(static_cast<int>(utils::ExitCode::SUCCESS));
+            }
         } else {
             input_error("unknown option " + arg);
         }
```
```

driver/tests.py
jendrikseipp: I made these changes to have one pytest test case per configuration instead of only a single one for all portfolios. With this change, we'll see the log output of the failing configuration directly instead of the log of all previous configurations.

driver/portfolios/seq_opt_fdss_2023.py
jendrikseipp: Thanks for the suggestions, both of you. But I think having the satisficing portfolio merged is enough for now.

Is that fine for you @maltehelmert ?

maltehelmert: Sure, works for me. Hopefully this can "soon" be done with the Python interface.

driver/tests.py
FlorianPommerening: So the goal is to change the tests from PDDL input and a single run to one translate run, followed by one planner run on the SAS+ file for every config? I don't see the change of input file in the diff, where does this happen?

Given the options in the comment, I like the `conftest.py` option more. Doing significant work when importing a module seems like a code smell to me and `pytest_sessionstart` sounds like the right place for setup like this. We recently added a `conftest.py` in `misc/tests/conftest.py`, so this would be consistent. I would add a comment at the top of the file explaining what it is, though. The first time I saw it in the pull request, I wanted to suggest to rename it to a better name, so a comment explaining that the filename is magic would be good. You could copy it from `misc/tests/conftest.py`.

jendrikseipp: The portfolio configs were always tested on SAS+ directly. But now each planner run is its own test case.

Good to merge now?

FlorianPommerening: Ahh, I see, this was in `run_driver` before. Looks good to me.

PR #178, opened 2023-09-07
Issue1108

ClemensBuechner:


src/search/landmarks/landmark_cost_assignment.cc
```diff
@@ -130,7 +128,8 @@ double LandmarkUniformSharedCostAssignment::cost_sharing_h_value(
             int num_achieved = achieved_lms_by_op[op_id];
             assert(num_achieved >= 1);
             assert(utils::in_bounds(op_id, operator_costs));
-            double shared_cost = static_cast<double>(operator_costs[op_id]) / num_achieved;
+            double shared_cost =
+                static_cast<double>(operator_costs[op_id]) / num_achieved;
```
maltehelmert, line 132: "shared_cost" is probably not what we would use these days. I'm not sure which would I would use. Something like "partitioned_cost" would be an improvement, I think.

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
```diff
@@ -18,13 +18,14 @@ namespace landmarks {
 LandmarkCostPartitioningHeuristic::LandmarkCostPartitioningHeuristic(
     const plugins::Options &opts)
     : LandmarkHeuristic(opts),
-      cost_partitioning_strategy(opts.get<CostPartitioningStrategy>("cost_partitioning")) {
+      cost_partitioning_strategy(
+          opts.get<CostPartitioningStrategy>("cost_partitioning")) {
```
jendrikseipp, line 22: Rename to `cost_partitioning_method` and `CostPartitioningMethod`? The `Algorithm` suffix would be even better, but I think we should use that name for the base class.

src/search/landmarks/landmark_cost_assignment.cc
```diff
@@ -17,12 +17,12 @@
 using namespace std;
 
 namespace landmarks {
-LandmarkCostAssignment::LandmarkCostAssignment(
+LandmarkCostPartitioningAlgorithm::LandmarkCostPartitioningAlgorithm(
```
jendrikseipp, line 20: The other classes don't have the `Landmark` prefix, so I'd drop it here as well.

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
jendrikseipp: Also, I think we should pass `cost_partitioning_method`, `alm` and `lpsolver` to the `set_cost_partitioning_algorithm` method and not store `cost_partitioning_method` as a member variable.

src/search/landmarks/landmark_cost_assignment.h
```diff
@@ -59,12 +60,11 @@ class LandmarkEfficientOptimalSharedCostAssignment : public LandmarkCostAssignme
...
-    LandmarkEfficientOptimalSharedCostAssignment(
-        const std::vector<int> &operator_costs,
-        const LandmarkGraph &graph,
-        lp::LPSolverType solver_type);
+    OptimalCostPartitioningAlgorithm(const std::vector<int> &operator_costs,
+                                     const LandmarkGraph &graph,
+                                     lp::LPSolverType solver_type);
 
-    virtual double cost_sharing_h_value(
+    virtual double compute_cost_partitioning(
```
jendrikseipp, line 67: Rename to `compute_cost_partitioned_h_value`? Not sure.

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
ClemensBuechner: I applied the changes in the first comment. Regarding the second one: I also removed the member. All options you mention are (still) passed to the function within the `opts` parameter for now. Would you also change the function header?

src/search/landmarks/landmark_cost_assignment.h
ClemensBuechner: I added the change. I'm also not sure what would be better, though.

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
jendrikseipp: I think it's bad practice to pass around the options variable, but let's not change this here.

src/search/landmarks/landmark_cost_assignment.h
maltehelmert: I'm not sure we're 100% consistent with this, but we use `compute_...` for void methods that store their results as a side effect because "compute" doesn't imply that something is passed back, only that a computation happens. This other common type of method is one with no side effects, but returning a value, for which we usually use words like `get_...`. We tend not to have methods both have a side effect and return a value, although I'm not sure in this case because the method is non-const. But there is also the distinction between having no side effects in a stronger sense or having no externally visible side effects.

Long story short, perhaps we want `get_cost_partitioned_h_value`.

maltehelmert: ...and perhaps write "heuristic", not "h".

maltehelmert (approved):

PR #180, opened 2023-09-13
Issue1075

ClemensBuechner:


src/search/landmarks/landmark_heuristic.cc
```diff
@@ -170,27 +135,11 @@ void LandmarkHeuristic::generate_preferred_operators(
...
             LandmarkNode *lm_node = lm_graph->get_node(fact_proxy.get_pair());
-            if (lm_node && landmark_is_interesting(
-                    state, past, *lm_node, all_landmarks_past)) {
-                if (lm_node->get_landmark().disjunctive) {
-                    preferred_operators_disjunctive.push_back(op_id);
-                } else {
-                    preferred_operators_simple.push_back(op_id);
-                }
+            if (lm_node && future.test(lm_node->get_id())) {
+                set_preferred(operators[op_id]);
```
salome-eriksson, line 139: change to `set_preferred(op)`

src/search/landmarks/landmark_heuristic.cc
```diff
@@ -137,30 +115,17 @@ void LandmarkHeuristic::compute_landmark_graph(const plugins::Options &opts) {
...
-    for (int i = 0; i < past.size(); ++i) {
-        if (!past.test(i)) {
-            all_landmarks_past = false;
-            break;
-        }
-    }
+    OperatorsProxy operators = task_proxy.get_operators();
 
     for (OperatorID op_id : applicable_operators) {
         OperatorProxy op = task_proxy.get_operators()[op_id];
```
salome-eriksson, line 131: Since you already defined `operators` on line 128 I would use it: `operators[op_id]`.
Alternatively, remove the definition on line 128.

src/search/landmarks/landmark_heuristic.cc
```diff
@@ -137,30 +115,16 @@ void LandmarkHeuristic::compute_landmark_graph(const plugins::Options &opts) {
...
 void LandmarkHeuristic::generate_preferred_operators(
-    const State &state, ConstBitsetView &past) {
+    const State &state, ConstBitsetView &future) {
     /*
-      Find operators that achieve landmark leaves. If a simple landmark can be
-      achieved, prefer only operators that achieve simple landmarks. Otherwise,
-      prefer operators that achieve disjunctive landmarks, or don't prefer any
-      operators if no such landmarks exist at all.
-
+      Find operators that achieve future landmarks.
```
maltehelmert, line 120: I'd like a clearer definition of how preferred operators are defined. One thing I don't understand based on the code here is how conjunctive landmarks are handled, but even without digging further I don't think they are handled correctly in the "natural" way. (We say an action achieves a landmark if it's not currently true but becomes true after action application.) The reason I think they are not handled correctly is because it's in general not possible to test whether a conjunctive landmark is achieved by iterating over operator effects one by one.

This is fine if the old code had similar difficulties, which is likely. We can then leave better handling of conjunctive landmarks to future work. But then it would be nice to keep a TODO or comment here somewhere and to perhaps add this to our landmark metaissue or another appropriate place.

If the change we make here significantly changes how conjunctive landmarks are handled, then we should perhaps run an experiment on this.

ClemensBuechner: Did you see the comment right below the line you commented on in the new code? (Lines 121 to 123) I hope this question doesn't appear rude, your comment just sounds like you didn't see that TODO.

Conjunctive landmarks are currently not considered at all when computing preferred operators. And there's already issue1072 aiming to fix that. How we do so should probably be discussed in the issue tracker.

PR #184, opened 2023-10-03
[trivial] Fix method name in LP solver interface.

jendrikseipp:


FlorianPommerening:
Oh, wow, this could have been bad. Luckily the method is not used in the code - we only modify constraint bounds in the current heuristics. Good catch!

FlorianPommerening (approved):

PR #186, opened 2023-10-04
[Trivial] Improve log message.

SimonDold:


maltehelmert (commented):
I think this would need more discussion, I don't think the new phrasing is an improvement. It looks like we lose the distinction between inherently incomplete algorithms like EHC and ones that are complete in principle, but can be made incomplete by certain parameter choices.

maltehelmert:
Also, I see that this would break some scripts, such as a merge-and-shrink parser than Silvan has been using. Not necessarily a big problem, but it means we should warn of this in the release notes. I think I've also seen others parse output for lines like these to figure out what the planner thought of the input.

maltehelmert:
I talked to some more people offline, and they agree that this will break their parsing and that they consider the current behaviour better than the proposed change, so I'm closing this. We can of course still work on this, but I guess we need some more discussion and it's not a trivial pull request (which would have to be uncontroversial).

PR #188, opened 2023-10-04
Fix CPLEX compilation on Windows for repos not named 'downward'.

jendrikseipp:


.github/workflows/windows.yml
```diff
@@ -71,12 +71,15 @@ jobs:
           echo "https://www.ibm.com/docs/en/icos/22.1.1?topic=2211-silent-installation-cplex-optimization-studio"
           curl.exe --output cplex.exe $ENV:CPLEX_URL
 
+          # This directory only already exists for repos called "downward".
+          mkdir D:\a\downward
```
FlorianPommerening, line 75: Is this necessary if we use `$ENV:GITHUB_WORKSPACE` later on?

.github/workflows/windows.yml
```diff
@@ -71,12 +71,15 @@ jobs:
...
+
           echo "Install CPLEX"
-          Start-Process -FilePath .\cplex.exe -ArgumentList "-f", "D:\a\downward\downward\.github\workflows\misc\cplex2211_windows_installer.properties" -PassThru | Wait-Process
+          Start-Process -FilePath .\cplex.exe -ArgumentList "-f", "$ENV:GITHUB_WORKSPACE\.github\workflows\misc\cplex2211_windows_installer.properties" -PassThru | Wait-Process
           del .\cplex.exe
 
           echo "Copy the relevant directory to a location which is not magically protected against cmake"
-          Xcopy /E /I ..\cplex_temp\cplex ..\cplex
+          Xcopy /E /I D:\a\downward\cplex_temp\cplex $ENV:cplex_DIR
```
FlorianPommerening, line 82: Shouldn't this be `$ENV:GITHUB_WORKSPACE` as well?

FlorianPommerening (commented):
Looks like there are two conflicting solutions: Getting rid of the hard-coded paths `D:\a\downward\` and working around them.

.github/workflows/windows.yml
jendrikseipp: Don't think so. The D:\a\downward path is hardcoded.

jendrikseipp:
> Looks like there are two conflicting solutions: Getting rid of the hard-coded paths `D:\a\downward\` and working around them.

Yes. I opted for the less intrusive one. Do you prefer the other one, i.e., using D:\a\cplex_temp and D:\a\cplex?

FlorianPommerening:
I'm fine with using the environment variable but what I meant was that there are still places with the hard-coded paths. It would make sense to use the variable path everywhere instead of using hard-coded paths in some places and variable paths in others.

FlorianPommerening:
That makes more sense now. The installation is still weird but I think all of that weirdness was there before (for example, the copy from `cplex_temp` to `cplex` because for some reason we cannot directly install to `cplex`). I'm fine with merging this.

PR #196, opened 2023-10-17
Issue1122 again

ClemensBuechner:


src/search/cmake/FindCplex.cmake
```diff
@@ -12,7 +12,11 @@
 set(IMPORTED_CONFIGURATIONS "Debug" "Release")
 set(HINT_PATHS ${cplex_DIR} $ENV{cplex_DIR})
 
-add_library(cplex::cplex IMPORTED UNKNOWN)
+if(WIN32)
+    add_library(cplex::cplex IMPORTED SHARED)
+else()
+    add_library(cplex::cplex IMPORTED UNKNOWN)
```
FlorianPommerening, line 18: Can you add a comment about why we do this?

FlorianPommerening (approved):

src/search/cmake/FindCplex.cmake
ClemensBuechner: What level of detail do you see fit? Suggestion: 
```# The Linux CPLEX installer does not provide dynamic libraries in all tested environments. If they are not available, we fall back to static ones further down, hence we mark the type unknown here.```

PR #198, opened 2023-11-24
[trivial] fix timer precision.

SimonDold:


src/search/utils/timer.cc
```diff
@@ -110,7 +111,7 @@ Duration Timer::reset() {
 }
 
 ostream &operator<<(ostream &os, const Timer &timer) {
-    os << timer();
+    os << fixed << setprecision(6) << timer();
```
jendrikseipp, line 114: I think two decimals are enough. This would also match the time precision we use in Lab.

jendrikseipp:
BTW, great that you tackle this issue! The old behavior bugged me again just a few days ago :)

src/search/utils/timer.cc
maltehelmert: I disagree that two decimals are enough. There's a difference between having two digits of precision for things like overall planner runtime in a lab report and being unable to report internal timings with a higher granularity. If we reduce the precision so far, I think this falls outside the scope of a "trivial" change. It would need discussion and mentioning in the release notes.

Can you update the pull request back to 6 digits? We can discuss a further reduction separately if you feel strongly about this, Jendrik. I personally think 6 digits (microseconds) is quite reasonable as a default.

jendrikseipp: I don't feel strongly about this. 6 digits is fine for me as well.

PR #200, opened 2023-12-19
Issue1128

ClemensBuechner:


src/search/pdbs/cegar.cc
```diff
@@ -385,7 +389,7 @@ bool CEGAR::get_flaws_for_pattern(
                 log << "plan did not lead to a goal state: ";
             }
             bool raise_goal_flaw = false;
-            for (const FactPair &goal : goals) {
+            for (const FactPair &goal : use_restricted_goal ? goals : task_properties::get_fact_pairs(task_proxy.get_goals())) {
```
FlorianPommerening, line 392: I find this a bit unintuitive. At this point `goals` is something that is passed to the class in the constructor. Instead of passing `goals` and an option that essential is `ignore_the_goal_I_gave_you_and_use_something_else`, can we handle the option earlier and just pass a modified goal vector to the class?

maltehelmert:
We are closing this pull request because we are not planning to pursue this idea further at this time.

See the tracker for more details (https://issues.fast-downward.org/issue1128).

PR #202, opened 2024-01-07
Enable more performance-related clang-tidy checks and fix code accordingly.

jendrikseipp:


misc/style/run-clang-tidy.py
```diff
@@ -47,15 +48,29 @@ def check_search_code_with_clang_tidy():
     # categories instead of deleting them to see which additional checks
     # we could activate.
     checks = [
-        # Enable with CheckTriviallyCopyableMove=0 when we require
-        # clang-tidy >= 6.0 (see issue856).
-        # "misc-move-const-arg",
-        "misc-move-constructor-init",
```
jendrikseipp, line 53: This has been renamed to "performance-move-const-arg", so I moved it down.

src/search/parser/abstract_syntax_tree.cc
```diff
@@ -404,7 +404,7 @@ const plugins::Type &ListNode::get_type(DecorateContext &context) const {
 }
 
 LiteralNode::LiteralNode(Token value)
```
FlorianPommerening, line 406: To make this more consistent with other places, I'd pass in a const ref here and copy the value in the initialization list, rather than passing in a copy and then moving the copy.

misc/style/run-clang-tidy.py
```diff
@@ -83,12 +98,17 @@ def check_search_code_with_clang_tidy():
         "readability-static-definition-in-anonymous-namespace",
         "readability-uniqueptr-delete-release",
         ]
+    config = """{CheckOptions: [\
+        {key: performance-unnecessary-value-param.AllowedTypes, value: "StateID"},\
```
FlorianPommerening, line 102: This exception is a bit hard to find in the middle of the file. Maybe make this a global variable (`LIGHTWEIGHT_TYPES = ["StateID"]` or something) at the start of the file next to `IGNORES`?

misc/style/run-clang-tidy.py
```diff
@@ -47,15 +48,29 @@ def check_search_code_with_clang_tidy():
     # categories instead of deleting them to see which additional checks
     # we could activate.
     checks = [
-        # Enable with CheckTriviallyCopyableMove=0 when we require
-        # clang-tidy >= 6.0 (see issue856).
```
FlorianPommerening, line 51: Did our required clang-tidy version change, or why is the comment no longer applicable?

silvansievers: We require `run-clang-tidy-12` since a while, so I assume the comment was outdated.

jendrikseipp: I just moved the comment down in the file.

FlorianPommerening: If it's outdated, it would be better to remove it.

jendrikseipp: It's not outdated. See the linked issue.

jendrikseipp:
Thanks for having a look! I took care of the comments.

PR #204, opened 2024-01-08
Try workaround for failing Github actions.

ClemensBuechner:


FlorianPommerening:
Looks good so far but it seems we also need the workaround for clang-tidy used in the style tests.
I'd add the step here (without the "if" statement because this action doesn't use a matrix setup):
https://github.com/aibasel/downward/blob/main/.github/workflows/style.yml#L27

ClemensBuechner:
Thanks @FlorianPommerening, now all checks passed. So when I merge this, would this be a commit with the "trivial" tag or should I open an issue for it?

silvansievers:
I'd say this goes with "trivial" - it doesn't affect the planner itself.

PR #205, opened 2024-01-09
Issue1111

ClemensBuechner:


FlorianPommerening (approved):
Looks fine to me, but I also was part of the discussion when we decided on this way of handling this, so this is a biased opinion.

PR #207, opened 2024-01-09
Backport some small changes from Scorpion.

jendrikseipp:


misc/tox.ini
```diff
@@ -70,7 +70,6 @@ commands =
 
 [testenv:clang-tidy]
 changedir = {toxinidir}/style/
-deps = PyYAML==6.0.1
```
jendrikseipp, line 73: PyYAML dependency is obsolete.

driver/portfolio_runner.py
```diff
@@ -120,7 +121,7 @@ def run_sat(configs, executable, sas_file, plan_manager, final_config,
                 configs, pos, search_cost_type, heuristic_cost_type,
                 executable, sas_file, plan_manager, timeout, memory)
             if exitcode is None:
-                return
+                continue
```
jendrikseipp, line 124: If one search has time limit 0 we want to switch to the next config, not abort the whole portfolio.

driver/portfolio_runner.py
```diff
@@ -83,9 +83,10 @@ def compute_run_time(timeout, configs, pos):
     print("remaining time: {}".format(remaining_time))
     relative_time = configs[pos][0]
     remaining_relative_time = sum(config[0] for config in configs[pos:])
-    print("config {}: relative time {}, remaining {}".format(
-          pos, relative_time, remaining_relative_time))
-    return limits.round_time_limit(remaining_time * relative_time / remaining_relative_time)
+    absolute_time_limit = limits.round_time_limit(remaining_time * relative_time / remaining_relative_time)
+    print("config {}: relative time {}, remaining time {}, absolute time {}".format(
+          pos, relative_time, remaining_relative_time, absolute_time_limit))
+    return absolute_time_limit
```
jendrikseipp, line 89: Print absolute time.

FlorianPommerening: This might break some existing parsing scripts, so I'd warn about this in the commit message.

src/search/algorithms/int_packer.cc
```diff
@@ -83,6 +84,8 @@ void IntPacker::set(Bin *buffer, int var, int value) const {
 
 void IntPacker::pack_bins(const vector<int> &ranges) {
     assert(var_infos.empty());
+    assert(all_of(ranges.begin(), ranges.end(),
+                  [](int range) {return range > 1;}));
```
FlorianPommerening, line 88: Is a range of 1 really a problem here? I can imagine this but I didn't see it, at least. It should take 1 bit to represent a variable with a single value. Variables like this are not useful but this should not be the concern of this algorithmic class.

Is the issue that `get_bit_size_for_range(1)` returns 0? In that case, I think that function should be fixed instead.

jendrikseipp: Good catch, done.

driver/portfolio_runner.py
maltehelmert: The rationale behind the original code is that each portfolio should have a non-zero relative time limit, else it shouldn't be part of the portfolio. This means that the time limit hits 0 for all configs at the same time (when the overall time limit is exhausted). So I don't see the purpose of the change. Can you say a scenario where it makes sense?

src/search/algorithms/int_packer.cc
maltehelmert: I don't agree. Wasting memory on a variable that can only take on a single value is a usage error for this class that it would be good to point out to the user. The point of these classes is to represent stuff space-efficiently. Wasting a bit on something that can only ever have the same value runs against this.

src/search/task_utils/task_properties.cc
```diff
@@ -173,9 +173,11 @@ PerTaskInformation<int_packer::IntPacker> g_state_packers(
         vector<int> variable_ranges;
         variable_ranges.reserve(variables.size());
         for (VariableProxy var : variables) {
-            variable_ranges.push_back(var.get_domain_size());
+            /* IntPacker expects all variables to have at least a domain size of
+               two. This is not the case for some domain-abstracted tasks. */
+            int domain_size = max(2, var.get_domain_size());
+            variable_ranges.push_back(domain_size);
```
maltehelmert: See comment above.

driver/portfolio_runner.py
jendrikseipp: If config5 has a low relative time limit in the portfolio file, it might get an absolute time limit of 0. But there could be a config6 with a high relative time limit that gets a non-zero absolute time limit.

src/search/algorithms/int_packer.cc
jendrikseipp: Which solution would you suggest instead?

driver/portfolio_runner.py
maltehelmert: How? I don't understand. Is there rounding involved that can round down to 0?

maltehelmert: I see now that indeed this is potentially rounded down before we get here. Then the change makes sense, sorry for the noise.

src/search/algorithms/int_packer.cc
maltehelmert: I have two possible suggestions, but I think it might be more efficient to quickly discuss this over Zoom if that works for you. I have a question or two about the use case that triggered the change. Are you available right now? If yes, let's arrange a Zoom call on Discord.

jendrikseipp: Unfortunately, I should get back to finishing ICAPS workshop papers now :-) Maybe we can chat briefly about this after one of the next reading groups or Fast Downward meetings.

maltehelmert: No worries. The short version not knowing the background for the change is: the old code was intended for the assumption that domain sizes are at least 2. If the assumption is still valid, we could add an assertion about this to make this point more explicit (assuming there wasn't already such an assertion). If we do need support for variables with domain {0} because it's not OK to require the user of the code to filter out such variables, we should store them in 0 bits, not 1. I don't think this would be difficult to do.

Can we arrange a concrete time to discuss this? I hope it will be only 5 minutes. I don't know when the workshop deadline is, but for example tomorrow I have some gaps in the schedule.

PR #208, opened 2024-01-09
Fix txt2tags markup.

jendrikseipp:


FlorianPommerening:
Looks like this removed the itemization for paper references. Was this intentional? I liked it better with the itemization, especially with multiple paper references.

PR #213, opened 2024-01-30
Issue1134

remochristen:


driver/aliases.py
```diff
@@ -141,7 +141,6 @@ def _get_lama(pref):
 ALIASES["seq-opt-lmcut"] = [
     "--search", "astar(lmcut())"]
 
-
```
FlorianPommerening, line 144: We should add the line back in just to avoid the diff in this file.

PR #216, opened 2024-01-31
Issue913 (instantiation bug with uninitialized numeric expressions)

roeger:


src/translate/normalize.py
```diff
@@ -366,10 +379,13 @@ def build_exploration_rules(task):
         proxy.build_rules(result)
     return result
 
-def condition_to_rule_body(parameters, condition):
+def condition_to_rule_body(parameters, condition, numeric_expression=None):
```
remochristen, line 382: I assume the numeric expression here is still always primitive? Maybe also rename the parameter to the more explicit `primitive_numeric_expression`. Either way I would name it the same as the variable in the calling code.

src/translate/pddl_to_prolog.py
```diff
@@ -142,6 +142,9 @@ def __str__(self):
         cond_str = ", ".join(map(str, self.conditions))
         return "%s :- %s." % (self.effect, cond_str)
 
+def get_definition_fluent(pne: pddl.PrimitiveNumericExpression):
+    return pddl.Atom(f"@def-{pne.symbol}", pne.args)
+
```
remochristen, line 147: - I think Gabi mentioned this, but I think it could indeed be nicer to move this to `normalize.py` and drop the `import pddl_to_prolog` there.
- I would make the parameter name match the one chosen in `normalize.py`.
- The name `get_definition_fluent` could maybe be less generic but I don't have a good suggestion.

src/translate/normalize.py
```diff
@@ -23,7 +24,19 @@ def delete_owner(self, task):
     def build_rules(self, rules):
         action = self.owner
         rule_head = get_action_predicate(action)
-        rule_body = condition_to_rule_body(action.parameters, self.condition)
+
+        # if the action cost is based on a primitive numeric expression,
+        # we need to require that it has a value defined in the initial state.
+        # We hand it over to condition_to_rule_body to include this in the rule
+        # body.
+        pne = None
```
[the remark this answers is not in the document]
roeger: I now went for pne everywhere. At one place, I explicitly mention primitive numeric expressions in the comment just before this piece of code. For 'conditions_to_rule_body', I added typing information to clarify what it is and added a bit more extensive comments. @remochristen maybe you want to have another look at this before merging.

src/translate/normalize.py
roeger: This is what I referred to in my other comment. Variables are now named the same and I added typing information to make things clearer.

src/translate/pddl_to_prolog.py
roeger: Parameter names now match. I moved the function to `normalize.py` and renamed it to `get_pne_definition_predicate`. I am not totally happy about the predicate part but it's the analogous name to the analogous functions `get_action_predicate` and `get_axiom_predicate` just next to it.

src/translate/normalize.py
```diff
@@ -366,10 +382,24 @@ def build_exploration_rules(task):
...
+       - all positive atoms in the condition (which must be normalized) are
+         true in the Prolog model, and
+       - the primitive numeric expression (from the action cost) has a defined
+         value (in the initial state)."""
     result = []
+    # Require parameters to be instantiated with objects of the right type.
     for par in parameters:
         result.append(par.get_atom())
+
+    # Require each positive literals in the condition to be reached and
```
remochristen, line 400:
```suggestion
    # Require each positive literal in the condition to be reached and
```
Besides that all looks good to me!

PR #218, opened 2024-02-01
[issue1082] Use Options objects only in Features

SimonDold:


src/search/evaluator.cc
```diff
@@ -9,11 +9,24 @@
 using namespace std;
 
 
+Evaluator::Evaluator(bool use_for_reporting_minima,
```
jendrikseipp, line 12: I'd add a newline after the opening bracket (here and in some other places below), but this is mainly a personal preference.

src/search/heuristic.h
```diff
@@ -73,8 +74,11 @@ class Heuristic : public Evaluator {
     State convert_ancestor_state(const State &ancestor_state) const;
 
 public:
+    explicit Heuristic(std::shared_ptr<AbstractTask> transform,
```
jendrikseipp, line 77: explicit is unneeded here.

src/search/utils/logging.cc
```diff
@@ -29,6 +29,13 @@ void add_log_options_to_feature(plugins::Feature &feature) {
         "normal");
 }
 
+LogProxy get_log_from_verbosity(const Verbosity &verbosity) {
```
jendrikseipp, line 32:
```suggestion
LogProxy get_log_for_verbosity(const Verbosity &verbosity) {
```

src/search/heuristics/max_heuristic.h
```diff
@@ -33,7 +33,10 @@ class HSPMaxHeuristic : public relaxation_heuristic::RelaxationHeuristic {
 protected:
     virtual int compute_heuristic(const State &ancestor_state) override;
 public:
-    explicit HSPMaxHeuristic(const plugins::Options &opts);
+    explicit HSPMaxHeuristic(const std::shared_ptr<AbstractTask> &transform,
```
jendrikseipp, line 36: remove `explicit`

jendrikseipp (changes requested):

src/search/evaluator.cc
```diff
@@ -97,6 +97,10 @@ void add_evaluator_options_to_feature(plugins::Feature &feature) {
     utils::add_log_options_to_feature(feature);
 }
 
+tuple<std::string, utils::Verbosity> get_evaluator_parameters_from_options(const plugins::Options &opts){
+    return utils::get_log_parameters_from_options(opts);
+}
```
maltehelmert, line 107: Our convention would be to write `string` rather than `std::string` here (for the same reason we write just `tuple`).

src/search/heuristic.cc
```diff
@@ -66,6 +66,16 @@ void Heuristic::add_options_to_feature(plugins::Feature &feature) {
...
 
+
+tuple<shared_ptr<AbstractTask>, bool, string, utils::Verbosity> Heuristic::get_heuristic_parameters_from_options(const plugins::Options &opts){
+    auto parent_parameter_tuple = get_evaluator_parameters_from_options(opts);
+    auto own_parameter_tuple = make_tuple(
+        opts.get<shared_ptr<AbstractTask>>("transform"),
+        opts.get<bool>("cache_estimates")
+    );
+    return tuple_cat(own_parameter_tuple, parent_parameter_tuple);
+}
```
maltehelmert, line 77: If we use this kind of pattern more frequently, it may be worth creating our own utility method that creates a tuple from elements and a "base" tuple in one go, along the lines of:

```
return make_flat_tuple(
    opts.get<shared_ptr<AbstractTask>>("transform"),
    opts.get<bool>("cache_estimates"),
    get_evaluator_parameters_from_options(opts));
```

If we only use this in cases where we have a feature class that is the same time a derived class and a base class (like Heuristic), it might not be worth it though, because we prefer flat inheritance hierarchies anyway. For example, the Evaluator/Heuristic distinction is something we have long wanted to get rid of, but didn't prioritize so far. (We did get rid of yet another intermediate class a long time ago though, and we reduced the differences between the two classes.)

src/search/heuristic.h
```diff
@@ -88,6 +88,8 @@ class Heuristic : public Evaluator {
     static void add_options_to_feature(plugins::Feature &feature, const std::string &name);
     static void add_options_to_feature(plugins::Feature &feature); // TODO 1082 remove this, just keep the one above
 
+    static std::tuple<std::shared_ptr<AbstractTask>, bool, std::string, utils::Verbosity> get_heuristic_parameters_from_options(const plugins::Options &opts);
```
maltehelmert, line 91: Do we generally use static methods for this kind of thing? Normally the convention we follow is to avoid access to private members where not needed, which means this could be a module-level function rather than being part of the class. This is in general the way modern C++ design leans.

src/search/heuristics/hm_heuristic.cc
```diff
@@ -12,19 +12,25 @@
...
-      has_cond_effects(task_properties::has_conditional_effects(task_proxy)),
-      goals(task_properties::get_fact_pairs(task_proxy.get_goals())) {
-    if (log.is_at_least_normal()) {
-        log << "Using h^" << m << "." << endl;
-        log << "The implementation of the h^m heuristic is preliminary." << endl
-            << "It is SLOOOOOOOOOOOW." << endl
-            << "Please do not use this for comparison!" << endl;
+
+    HMHeuristic::HMHeuristic(
+            const int m,
```
maltehelmert: Should not be `const`.

src/search/heuristics/hm_heuristic.cc
```diff
@@ -269,7 +275,7 @@ class HMHeuristicFeature : public plugins::TypedFeature<Evaluator, HMHeuristic>
         document_title("h^m heuristic");
 
         add_option<int>("m", "subset size", "2", plugins::Bounds("1", "infinity"));
-        Heuristic::add_options_to_feature(*this);
+        Heuristic::add_options_to_feature(*this, "hm");
```
maltehelmert, line 277: We had some discussions (I don't remember where and who participated) that this injection of default values into the base class will eventually become problematic when we switch to loggers. At this point, I just want to be sure that we're all aware of this and that there won't be surprises if we end up essentially undoing this change at a later point.

src/search/heuristics/hm_heuristic.cc
```diff
@@ -286,6 +292,12 @@ class HMHeuristicFeature : public plugins::TypedFeature<Evaluator, HMHeuristic>
             "yes for tasks without conditional effects or axioms");
         document_property("preferred operators", "no");
     }
+
+    virtual shared_ptr<HMHeuristic> create_component(const plugins::Options &options, const utils::Context &) const override {
+        tuple parameter_tuple = tuple_cat(make_tuple(options.get<int>("m")),
+                                          Heuristic::get_heuristic_parameters_from_options(options));
+        return plugins::make_shared_from_tuple<HMHeuristic>(parameter_tuple);
```
maltehelmert, line 298: This is really an argument tuple, not a parameter tuple. Similarly to what I wrote above, it might make sense to encapsulate this pattern once and for all because we will likely see it 100+ times.  Here the potential value is higher because we will have mostly leaf classes.

For example, the whole thing could be:
```
return plugins::make_shared_from_flattened_arguments(
    options.get<int>("m"),
    Heuristic::get_heuristic_parameters_from_options(options)
);
```

We could even go one step further, use a uniform name for `get_heuristic_parameters_from_options`-style functions and encapsulate the base class call into a template:

```
return plugins:make_shared_from_tuple_with_inheritance<HMHeuristic, Heuristic>(
    options.get<int>("m"));
```

or even encapsulate the base class in a consistent way and make this:

```
return plugins:make_shared_from_tuple_with_inheritance<HMHeuristic>(
    options.get<int>("m"));
```

Or go even further, have a `get_parameters`-style function for the leaf classes away and have the whole function go away in cases where we just do the standard thing.

src/search/heuristics/hm_heuristic.h
```diff
@@ -61,7 +61,12 @@ class HMHeuristic : public Heuristic {
     virtual int compute_heuristic(const State &ancestor_state) override;
 
 public:
-    explicit HMHeuristic(const plugins::Options &opts);
+    HMHeuristic(
+            const int m,
```
maltehelmert: Should not be `const`.

src/search/plugins/plugin.h
```diff
@@ -115,6 +115,13 @@ class TypedFeature : public FeatureAuto<Constructed> {
     }
 };
 
+// TODO issue1082 where should this live?
```
maltehelmert, line 118: To be discussed.

src/search/plugins/plugin.h
```diff
@@ -115,6 +115,13 @@ class TypedFeature : public FeatureAuto<Constructed> {
...
+// Apply the arguments extracted from *parameter_tuple* to make_shared<T>
+template<typename T, typename ...Args>
+std::shared_ptr<T> make_shared_from_tuple(std::tuple<Args...> parameter_tuple){
+       return apply([](auto... args) { return make_shared<T>(args...); }, parameter_tuple);
```
maltehelmert: Does this need perfect-forwarding references-style stuff? I'd have to read up on it.

src/search/utils/logging.cc
```diff
@@ -40,6 +40,17 @@ void add_log_options_to_feature(plugins::Feature &feature) {
         "normal");
 }
 
+
+tuple<string, Verbosity> get_log_parameters_from_options(const plugins::Options &opts){
+    auto own_tuple = make_tuple<string, Verbosity>(
+            opts.get<string>("name"), // TODO issue1082 move name into get_evaluator_parameters_from_options and rename it to *description*.
+            opts.get<Verbosity>("verbosity")
+    );
```
maltehelmert: Analogous comments to the above reply. `own_tuple` seems an awkward variable name.
I don't think there's a value in introducing the name just to return it, it doesn't add extra documentation.

src/search/heuristic.cc
```diff
@@ -66,6 +66,16 @@ void Heuristic::add_options_to_feature(plugins::Feature &feature) {
     add_evaluator_options_to_feature(feature);
 }
 
+
+tuple<shared_ptr<AbstractTask>, bool, string, utils::Verbosity> Heuristic::get_heuristic_parameters_from_options(const plugins::Options &opts){
+    auto parent_parameter_tuple = get_evaluator_parameters_from_options(opts);
```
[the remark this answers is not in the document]
SimonDold: yes, we also use it analogous in `hm_heuristc.cc` for `create_component`.
The GitHub actions did not complain about it.

src/search/plugins/plugin.h
SimonDold: `// Apply the arguments extracted from *parameter_tuple* to make_shared<T>
template<typename T, typename ... Args>
std::shared_ptr<T> make_shared_from_tuple(std::shared_ptr<std::tuple<Args...>> parameter_tuple) {
    return apply([](auto... args) {return make_shared<T>(args ...);}, *parameter_tuple);
}`
also works

src/search/heuristics/cea_heuristic.cc
```diff
@@ -409,8 +409,11 @@ int ContextEnhancedAdditiveHeuristic::compute_heuristic(
 }
 
 ContextEnhancedAdditiveHeuristic::ContextEnhancedAdditiveHeuristic(
-    const plugins::Options &opts)
-    : Heuristic(opts),
+    const shared_ptr<AbstractTask> &transform,
+    bool cache_estimates,
+    const string &name,
```
SimonDold, line 414: `name` should be `description`
also in the .h file

src/search/heuristics/cg_heuristic.cc
```diff
@@ -16,8 +16,13 @@ using namespace std;
 using namespace domain_transition_graph;
 
 namespace cg_heuristic {
-CGHeuristic::CGHeuristic(const plugins::Options &opts)
-    : Heuristic(opts),
+CGHeuristic::CGHeuristic(
+    int max_cache_size,
+    const shared_ptr<AbstractTask> &transform,
+    bool cache_estimates,
+    const string &name,
```
SimonDold, line 23: name -> description
also in .h

src/search/pdbs/pattern_collection_generator_multiple.cc
```diff
@@ -322,7 +330,40 @@ void add_multiple_options_to_feature(plugins::Feature &feature) {
         "generation is terminated already the first time stagnation_limit is "
         "hit.",
         "true");
-    add_generator_options_to_feature(feature);
     utils::add_rng_options(feature);
+    add_generator_options_to_feature(feature);
+}
+
+shared_ptr<tuple<
```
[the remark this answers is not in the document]
FlorianPommerening: And one more point: the name doesn't match what the function does. We already have `get_generator_parameters_from_options` which lives in `pattern_generator` (next to `PatternGenerator` and `PatternCollectionGenerator`). This function here is specifically for the `multiple` version of the generator and should mirror the name of `add_multiple_options_to_feature` above.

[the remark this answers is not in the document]
FlorianPommerening: Sorry, another point: `random_seed` is the argument for an RNG. It is added by `utils::add_rng_options` above, so shouldn't it have a matching function call like `get_generator_parameters_from_options`?

src/search/utils/tuples.h
```diff
@@ -0,0 +1,31 @@
+#ifndef UTILS_TUPLES_H
```
salome-eriksson, line 1: I wonder if adding some comments on how this template magic works would be beneficial. I only understood the code when Florian walked me through it. On the other hand I'm not sure if we need a detailed explanation, or are just happy that it does work ;).

salome-eriksson (commented):
This is only a partial review looking at the non-mechanical changes. The second part will follow later.

src/search/utils/logging.h
```diff
@@ -130,7 +130,13 @@ class LogProxy {
 extern LogProxy g_log;
 
 extern void add_log_options_to_feature(plugins::Feature &feature);
-extern LogProxy get_log_from_options(const plugins::Options &options);
+
+extern std::tuple<Verbosity> get_log_arguments_from_options(
+    const plugins::Options &opts);
+
```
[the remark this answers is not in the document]
SimonDold: No reason. I change it.

src/search/utils/tuples.h
SimonDold: I think this is hard to understand and I also only understood it when Florian explained it to me.

A permanent place to find an explanation of it would be nice but the comments might not be the right place for it, is it?

src/search/utils/rng_options.cc
```diff
@@ -17,9 +17,14 @@ void add_rng_options(plugins::Feature &feature) {
         plugins::Bounds("-1", "infinity"));
 }
 
-shared_ptr<RandomNumberGenerator> parse_rng_from_options(
-    const plugins::Options &options) {
-    int seed = options.get<int>("random_seed");
+tuple<int> get_rng_arguments_from_options(
+    const plugins::Options &opts) {
+    return make_tuple(
```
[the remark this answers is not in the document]
SimonDold: i agree

FlorianPommerening (commented):
Part 1 of my comments, I'll add the rest later.

salome-eriksson (commented):
Most of the comments are easy fixes.
Comments about style can also be ignored if you disagree.
One recurring comment is that the order of arguments in the constructor does not match the order of fields in the class definition. I personally find it more readable if they are the same (that is, if the arguments are used in the same order in the initialization list as in the constructor arguments, minus the ones that go to the parent of course). But I'm also ok with leaving it as it is.

FlorianPommerening:
> the order of arguments in the constructor does not match the order of fields in the class definition.
Isn't this a compiler warning that we treat as an error? I guess not, since it compiles, but I thought this would be a requirement.

maltehelmert:
> > the order of arguments in the constructor does not match the order of fields in the class definition.
> 
> Isn't this a compiler warning that we treat as an error? I guess not, since it compiles, but I thought this would be a requirement.

I think it would be good to trigger an error in the Github Actions, but not I don't think it's necessarily worth investing time to try to make that happen if it doesn't currently.

In case you don't know, unless I misremember, C++ semantics is that the initialization (including evaluation of the initializers) happens in the order in the class definition, not the order given in the constructor.

So for example, something like this:

```X::X() : foo(f()), bar(g()) {}```

would first call `g()` in order to initialize `bar` and then call `f()` in order to initalize `foo` if `bar` comes before `foo` in the class definition.

File: misc/tests/test-parameters.py [hunk 22]
```diff
@@ -0,0 +1,258 @@
...
+def matching(opening, closing):
+    return ((opening, closing) == ('(', ')') or
+        (opening, closing) == ('[', ']'))
+
+def extract_feature_parameter_list(feature_name):
+    s = str(subprocess.run(["./../../builds/release/bin/downward",
+                            "--help", "--txt2tags",
+                            "{}".format(feature_name)],
+                           stdout=subprocess.PIPE).stdout)
+    position = s.find(feature_name + "(")
```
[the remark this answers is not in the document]
salome-eriksson: Ah I see I missed the part on line 92 :). Then I don't really have an opinion, adding the ``` makes it more sturdy as you say, but also makes it more specific to how we format the output.

misc/tests/test-parameters.py
```diff
@@ -0,0 +1,258 @@
...
+    result = ''.join(result)
+    result = re.sub(r'=.*?,', ',', result + ",").split()
+    return result
+
+def extract_feature_name_and_cpp_class(cc_file, cc_files, cwd, num):
+    source_without_comments = subprocess.check_output(
+        ["gcc", "-fpreprocessed", "-dD", "-E", cc_file]).decode("utf-8")
+
+    name_pattern = r'TypedFeature\("([^"]*)"\)'
+    class_pattern = r'TypedFeature<(.*?)>'
```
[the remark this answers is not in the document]
salome-eriksson: Sounds good.

misc/tests/test-parameters.py
```diff
@@ -0,0 +1,258 @@
...
+            found_in_file, parameters = get_constructor_parameters(
+                cc_file2, class_name)
+            if found_in_file:
+                break
+
+    if found_in_file:
+        parameters = parameters.replace("\n", "") + ","
+        parameters = parameters.split()
+        parameters = [word for word in parameters if "," in word]
+        parameters = [re.sub(C_VAR_PATTERN, '', word) + ","
```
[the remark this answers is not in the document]
salome-eriksson: Yes I just meant appending the comma at the end. I think I was confused because the commas should be there already but I see now that the `C_VAR_PATTERN` strips them away again. So all good :).

src/search/cartesian_abstractions/utils_landmarks.cc
```diff
@@ -19,14 +19,10 @@ static FactPair get_fact(const Landmark &landmark) {
...
-    hm_opts.set<int>("m", 1);
-    hm_opts.set<bool>("only_causal_landmarks", false);
-    hm_opts.set<bool>("conjunctive_landmarks", false);
-    hm_opts.set<bool>("use_orders", true);
-    hm_opts.set<utils::Verbosity>("verbosity", utils::Verbosity::SILENT);
-    LandmarkFactoryHM lm_graph_factory(hm_opts);
+shared_ptr<LandmarkGraph> get_landmark_graph(
+    const shared_ptr<AbstractTask> &task) {
+    LandmarkFactoryHM lm_graph_factory(
+        1, false, true, utils::Verbosity::SILENT);
```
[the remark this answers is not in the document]
salome-eriksson: This was an oversight from us when we overhauled the parameters for the different landmark factories. Back then we removed the `only_causal_landmarks` parameters from this factory but didn't remove putting it into the option object here.

misc/tests/test-parameters.py
```diff
@@ -0,0 +1,258 @@
...
+
+HELP = """\
+Check that parameters for the command line features match the parameters
+ of the C++ constructors. Use txt2tags to compare the parameters
+ mentioned in the wiki to the parameters in the corresponding .cc file.
+"""
+
+import argparse
+import os
+import os.path
```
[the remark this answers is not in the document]
SimonDold: i agree, i switch to pathlib

src/search/cartesian_abstractions/utils.cc
```diff
@@ -15,11 +13,9 @@ using namespace std;
 namespace cartesian_abstractions {
 unique_ptr<additive_heuristic::AdditiveHeuristic> create_additive_heuristic(
```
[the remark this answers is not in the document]
SimonDold: yes, i think that would be better. I'll do this.
It is only called by subtask_generator and split_selector, so we would repeat the `false, "h^add within CEGAR abstractions", utils::Verbosity::SILENT` part only once.

src/search/merge_and_shrink/merge_scoring_function_miasm.h
```diff
@@ -12,7 +12,7 @@ namespace merge_and_shrink {
 class ShrinkStrategy;
 class MergeScoringFunctionMIASM : public MergeScoringFunction {
     const bool use_caching;
-    std::shared_ptr<ShrinkStrategy> shrink_strategy;
+    const std::shared_ptr<ShrinkStrategy> &shrink_strategy;
```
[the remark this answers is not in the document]
SimonDold: Okay I figured out why i did change something here.
clang-tidy:
```Errors and warnings:
/downward/src/search/merge_and_shrink/merge_scoring_function_miasm.cc:24:23: 
warning: parameter 'shrink_strategy' is passed by value and only copied once;
consider moving it to avoid unnecessary copies [performance-unnecessary-value-param]
```

I did undo it here but changed the constructor to actually move the shrink_strategy

src/search/pdbs/zero_one_pdbs_heuristic.h
```diff
@@ -1,10 +1,13 @@
 #ifndef PDBS_ZERO_ONE_PDBS_HEURISTIC_H
 #define PDBS_ZERO_ONE_PDBS_HEURISTIC_H
 
+#include "pattern_generator.h"
 #include "zero_one_pdbs.h"
 
 #include "../heuristic.h"
 
+#include <memory>
```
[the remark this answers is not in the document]
SimonDold: ill undo it

src/search/merge_and_shrink/merge_scoring_function_miasm.h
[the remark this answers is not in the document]
maltehelmert: For someone trying to understand the rationale of this, I think this is a good link:
 https://stackoverflow.com/questions/41871115/why-would-i-stdmove-an-stdshared-ptr

Salomé was right we definitely don't want to change the member from a pointer to a pointer reference; this is very likely to crash and burn once we actually try to access the shrink strategy later because we would reference a temporary value. More generally, I haven't looked at it much, but I don't think there should be many cases where we need to change the data members in the class definitions. The cases where we do make such changes we need to look at carefully.

src/search/cartesian_abstractions/cost_saturation.h
```diff
@@ -30,8 +30,8 @@ class CostSaturation {
     const int max_states;
     const int max_non_looping_transitions;
     const double max_time;
-    const bool use_general_costs;
     const PickSplit pick_split;
+    const bool use_general_costs;
```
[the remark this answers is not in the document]
SimonDold: To match the parameter order in the constructor of CostSaturation, there it was changed to match the constructor of AdditeiveCartesianHeuristic, there it was chosen in the way to match the wiki.

src/search/search_algorithms/plugin_eager_greedy.cc
```diff
@@ -37,38 +39,46 @@ class EagerGreedySearchFeature : public plugins::TypedFeature<SearchAlgorithm, e
...
             "```\n--search eager(single(eval1))\n```\n", true);
     }
 
-    virtual shared_ptr<eager_search::EagerSearch> create_component(const plugins::Options &options, const utils::Context &context) const override {
-        plugins::verify_list_non_empty<shared_ptr<Evaluator>>(context, options, "evals");
-        plugins::Options options_copy(options);
-        options_copy.set("open", search_common::create_greedy_open_list_factory(options_copy));
-        options_copy.set("reopen_closed", false);
-        shared_ptr<Evaluator> evaluator = nullptr;
-        options_copy.set("f_eval", evaluator);
```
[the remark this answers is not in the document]
SimonDold: We kept it if it was not trivial how to remove it.
It would be possible to get rid of more opts_copies by moving the logic into the create_component function but I think keeping it somewhat encapsulated is nice.

PR #220, opened 2024-02-08
Issue454 - move computation of negated axioms to search component

salome-eriksson:


src/search/tasks/negated_axioms_task.cc
```diff
@@ -0,0 +1,323 @@
...
+            negated_axioms.emplace_back(FactPair(var, 1), vector<FactPair>(), name);
+        } else {
+            // TODO: fix the hardcoded 1
+            add_negated_axioms(FactPair(var, 1), axioms, task_proxy);
+        }
+    }
+    cout << negated_axioms.size() << " negated axioms" << endl;
+}
+
+void NegatedAxiomsTask::find_non_dominated_hitting_sets(
```
salome-eriksson, line 148: This is a recursive function to multiply out `conditions_as_cnf`. It could also be implemented as a while loop instead, but I found this more readable.
(I would probably want to change the name though, but don't have a good idea at the moment.)

salome-eriksson (commented):
I added some comments on the code where I try to explain why I implemented certain things the way I did.

src/search/tasks/negated_axioms_task.cc
```diff
@@ -0,0 +1,323 @@
...
+        }
+    }
+
+    // If there are no derived variables that occur as their default value we don't need to do anything.
+    if (needed_negatively.empty()) {
+        return;
+    }
+
+    // Build dependency graph and collect relevant axioms for derived values occurring as their default value.
+    // TODO: The dependency information is a superset of positive_dependencies from above.
```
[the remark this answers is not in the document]
salome-eriksson: I resolved this by only using positive_dependencies, since negative dependencies cannot introduce a cycle (this would imply non-stratifiable axioms, which is already checked in the translator.)

src/search/tasks/negated_axioms_task.cc
```diff
@@ -0,0 +1,358 @@
...
+
+namespace tasks {
+NegatedAxiomsTask::NegatedAxiomsTask(const shared_ptr<AbstractTask> &parent)
+    : DelegatingTask(parent),
+      negated_axioms_start_index(parent->get_num_axioms()) {
+    TaskProxy task_proxy(*parent);
+
+    vector<vector<int>> pos_dependencies(task_proxy.get_variables().size());
+    vector<vector<int>> neg_dependencies(task_proxy.get_variables().size());
+    vector<vector<int>> axiom_ids_for_var(task_proxy.get_variables().size());
```
roeger, line 25: Maybe add a brief comment, what is stored in these vectors.

src/search/tasks/negated_axioms_task.cc
```diff
@@ -0,0 +1,358 @@
...
+       stratifiable, which is already checked in the translator.
+    */
+    vector<vector<int>> sccs =
+        sccs::compute_maximal_sccs(pos_dependencies);
+    vector<int> var_to_scc(task_proxy.get_variables().size(), -1);
+    for (int i = 0; i < (int)sccs.size(); ++i) {
+        for (int var: sccs[i]) {
+            var_to_scc[var] = i;
+        }
+    }
```
roeger, line 62: Maybe I am coming too much from the Python side here but wouldn't it be more natural to store in `var_to_scc` references to the respective SCCs rather than an index into `sccs`?

src/search/tasks/negated_axioms_task.cc
```diff
@@ -0,0 +1,358 @@
...
+                FactPair(var, default_value), vector<FactPair>());
+        } else {
+            add_negated_axioms_for_var(
+                FactPair(var, default_value), axiom_ids);
+        }
+    }
+}
+
+
+unordered_set<int> NegatedAxiomsTask::collect_needed_negatively(
```
roeger, line 94: It would be helpful to add some high-level documentation of what this is doing.

src/search/tasks/negated_axioms_task.cc
```diff
@@ -0,0 +1,358 @@
...
+    const vector<vector<int>> &positive_dependencies,
+    const vector<vector<int>> &negative_dependencies) {
+    // Stores which derived variables are needed positively or negatively.
+    set<pair<int, bool>> needed;
+
+    // TODO: Should we store the proxy in the class? Pass it through methods?
```
roeger, line 100: I would leave it as is. This is a super light-weight operation.

src/search/tasks/negated_axioms_task.cc
```diff
@@ -0,0 +1,358 @@
...
+            }
+        }
+    }
+    unordered_set<int> needed_negatively =
+        collect_needed_negatively(pos_dependencies, neg_dependencies);
+
+    /* TODO: I removed the return if needed_negatively is empty.
+       It is somewhat wasteful if we have a big graph to compute the sccs,
+       but probably should not be an issue. */
+
```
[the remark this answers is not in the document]
roeger: I am also not sure whether we want to expose it. We do this over-approximation for larger SCCs, which leads to weird behavior where you can derive from a single basic states both values for a derived variable. This is handled correctly in the axiom evaluator (ignoring axioms that set a variable to their default value) but might be unexpected to external users. In any case, it needs to be documented more clearly (also for internal use).

src/search/tasks/negated_axioms_task.cc
```diff
@@ -0,0 +1,358 @@
...
+                to_process.emplace_back(pos_dep, non_default);
+            }
+        }
+        for (int neg_dep : negative_dependencies[var]) {
+            auto insert_retval = needed.emplace(neg_dep, !non_default);
+            if (insert_retval.second) {
+                to_process.emplace_back(neg_dep, !non_default);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
roeger: Thinking more about this, I am not even convinced that it is the right thing to do. Am I right that we are anticipating here later additions of axioms? If yes, we don't consider that for large strongly connected components, we do an over-approximation that does not consider the existing rules at all.

roeger (commented):
First round of comments. I haven't looked at everything, yet.

src/search/tasks/negated_axioms_task.cc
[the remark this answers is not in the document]
salome-eriksson: No we don't anticipate the rules we will add later. The case distinction between positive and negative dependencies is to figure out whether we are really ever interested in the negative (=default) value. For example, if I have derived variables 'a` and `b`, with rule `a <- ¬b`, `a` only occurs as `¬a` in the goal and `b` does not occur in any goal or operators, then `b` is only needed positively.

src/search/tasks/negated_axioms_task.cc
```diff
@@ -0,0 +1,323 @@
...
+int NegatedAxiomsTask::convert_operator_index_to_parent(int index) const {
+    return index;
+}
+
+int NegatedAxiomsTask::get_num_axioms() const {
+    return parent->get_num_axioms() + negated_axioms.size();
+}
+
+
+// TODO: should we even offer this as a Feature? Can we just leave this out if we only want it used internally?
```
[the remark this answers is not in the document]
salome-eriksson: I currently have a comment where the negated axioms are computed, but I guess you mean something more high-level and easily visible (above the class definition probably?), which makes it clear what the use case for this transformation is. Will do.

src/search/tasks/negated_axioms_task.cc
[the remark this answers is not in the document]
roeger: I am not sure I can follow. Why is in your example b needed positively? I mean, why at all?

src/search/tasks/negated_axioms_task.h
```diff
@@ -0,0 +1,62 @@
...
+namespace plugins {
+class Options;
+}
+
+namespace tasks {
+struct NegatedAxiom {
+    FactPair head;
+    std::vector<FactPair> condition;
+
+    // TODO: clang suggested moving the second argument, is this correct? (clang doesn't complain at least :))
```
roeger, line 21: I did not see anything that would render this wrong.

src/search/tasks/negated_axioms_task.h
```diff
@@ -0,0 +1,62 @@
...
+    virtual FactPair get_operator_precondition(
+        int op_index, int fact_index, bool is_axiom) const override;
+    virtual int get_num_operator_effects(int op_index, bool is_axiom) const override;
+    virtual int get_num_operator_effect_conditions(
+        int op_index, int eff_index, bool is_axiom) const override;
+    virtual FactPair get_operator_effect_condition(
+        int op_index, int eff_index, int cond_index, bool is_axiom) const override;
+    virtual FactPair get_operator_effect(
+        int op_index, int eff_index, bool is_axiom) const override;
+    // TODO: Is this only called for actual operators or also axioms? I assume so because there is no bool is_axiom...
```
roeger, line 55: I think it can only be called for actual operators (otherwise there is no way to determine whether it means the corresponding operator or axiom).

src/search/tasks/negated_axioms_task.cc
```diff
@@ -0,0 +1,358 @@
...
+    if (!is_axiom || op_index < negated_axioms_start_index) {
+        return parent->get_operator_effect(op_index, eff_index, is_axiom);
+    }
+
+    assert(eff_index == 0);
+    return negated_axioms[op_index - negated_axioms_start_index].head;
+}
+
+int NegatedAxiomsTask::convert_operator_index_to_parent(int index) const {
+    return index;
```
roeger, line 329: I don't think this should be overwritten at all.

src/search/tasks/negated_axioms_task.cc
```diff
@@ -0,0 +1,358 @@
...
+    if (!is_axiom || op_index < negated_axioms_start_index) {
+        return parent->get_num_operator_effect_conditions(op_index, eff_index, is_axiom);
+    }
+
+    return 0;
+}
+
+FactPair NegatedAxiomsTask::get_operator_effect_condition(
+    int op_index, int eff_index, int cond_index, bool is_axiom) const {
+    assert(!is_axiom || op_index < negated_axioms_start_index);
```
roeger, line 314: The handling of NegatedAxioms and Axioms is not consistent in this transformation. If I am not mistaken, the body of a normal axiom is treated as an effect condition, whereas you treat the ones of a negated axiom as precondition. Right?

roeger (commented):
Hi @salome-eriksson,

I added another round of comments.

For the record, we talked about two more general aspects in person:
- it is a bit confusing that parts of the implementation have a perspective of a fact that can be true or false (e.g. "needed negatively") and other parts use the perspective of a binary finite-domain variable that has a default value and the other value.
- @speckdavid suggested to add an option to the affected heuristics to use the overapproximation for all derived variables that are needed negatively (weakening the heuristic but saving the time for the transformation).

Best,
Gabi

src/search/heuristics/cea_heuristic.cc
```diff
@@ -412,6 +413,12 @@ ContextEnhancedAdditiveHeuristic::ContextEnhancedAdditiveHeuristic(
     const plugins::Options &opts)
     : Heuristic(opts),
       min_action_cost(task_properties::get_min_operator_cost(task_proxy)) {
+
+    if (task_properties::has_axioms(task_proxy)) {
+        task = make_shared<tasks::NegatedAxiomsTask>(tasks::NegatedAxiomsTask(task));
+        task_proxy = TaskProxy(*task);
+    }
```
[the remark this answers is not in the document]
salome-eriksson: I locally tried Christian's tasks with the cea heuristic (so a setting where we only do the transformation once) and that got stuck for at least 10 minutes, so these tasks are already unmanageable when only computing the transformation once. But yes, it would be even worse with lama-first :).

src/search/tasks/negated_axioms_task.cc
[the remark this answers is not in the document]
salome-eriksson: Done

src/search/tasks/negated_axioms_task.cc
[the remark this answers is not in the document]
salome-eriksson: Due to other changes (discussed in person with @roeger) we moved the computations of sccs up, making returning earlier pointless.

src/search/tasks/negated_axioms_task.cc
salome-eriksson: References in vectors are very cumbersome, but i used pointers now.

src/search/tasks/negated_axioms_task.cc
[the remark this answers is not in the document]
salome-eriksson: Done.

src/search/tasks/negated_axioms_task.cc
```diff
@@ -0,0 +1,358 @@
...
+        for (const FactPair &f : hitting_set) {
+            if (f.var == elem.var) {
+                same_var_occurs_already = true;
+                break;
+            }
+        }
+        if (same_var_occurs_already) {
+            continue;
+        }
+
```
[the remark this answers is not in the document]
salome-eriksson: I added the set `hitting_set_vars` and now just check if elem.var is in this set; is that what you meant?
At the same time I change `hitting_set` to also just add and remove elements instead of copying it all the time.

src/search/tasks/negated_axioms_task.cc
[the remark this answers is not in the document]
salome-eriksson: I removed the Feature and added a high-level comment to the header file.

src/search/landmarks/landmark_heuristic.cc
```diff
@@ -26,8 +32,10 @@ void LandmarkHeuristic::initialize(const plugins::Options &opts) {
       CostAdaptedTask *of the root task*, but there is currently no good
       way to do this, so we use this incomplete, slightly less safe test.
     */
+    // TODO: update comment and cerr message
```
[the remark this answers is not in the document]
salome-eriksson: Done

src/search/tasks/negated_axioms_task.cc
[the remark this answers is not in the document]
salome-eriksson: Done.

src/search/tasks/negated_axioms_task.h
```diff
@@ -0,0 +1,62 @@
+#ifndef TASKS_NEGATED_AXIOMS_TASK_H
+#define TASKS_NEGATED_AXIOMS_TASK_H
+
+#include "delegating_task.h"
+
+#include "../task_proxy.h" // TODO: can we get rid of this include?
+
+#include "../algorithms/sccs.h"
```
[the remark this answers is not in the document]
salome-eriksson: No, I cleaned up the includes in general a bit (also in the cc file).

src/search/tasks/negated_axioms_task.cc
```diff
@@ -0,0 +1,358 @@
...
+        }
+        if (not_uniquely_used.empty()) {
+            results.insert(hitting_set);
+        }
+        return;
+    }
+
+    const set<FactPair> &conditions = conditions_as_cnf[index];
+    for (const FactPair &fact_pair : conditions) {
+        // The current set is covered with the current hitting set elements.
```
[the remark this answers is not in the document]
salome-eriksson: Done. I also replaced set with condition (I found set and hitting set a bit confusing).

src/search/tasks/negated_axioms_task.cc
```diff
@@ -0,0 +1,358 @@
...
+    }
+}
+
+
+void NegatedAxiomsTask::collect_non_dominated_hitting_sets_recursively(
+    const vector<set<FactPair>> &conditions_as_cnf, size_t index,
+    set<FactPair> &hitting_set, set<set<FactPair>> &results) {
+    if (index == conditions_as_cnf.size()) {
+        /*
+           Check whether the axiom body denoted in body is dominated.
```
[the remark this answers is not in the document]
salome-eriksson: I guess I must have written the comment when the parameters were called different. I adjusted the names and also rephrased the comment in general.

src/search/tasks/negated_axioms_task.cc
```diff
@@ -0,0 +1,358 @@
...
+        }
+    }
+}
+
+
+unordered_set<int> NegatedAxiomsTask::collect_needed_negatively(
+    const vector<vector<int>> &positive_dependencies,
+    const vector<vector<int>> &negative_dependencies) {
+    // Stores which derived variables are needed positively or negatively.
+    set<pair<int, bool>> needed;
```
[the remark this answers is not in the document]
salome-eriksson: I think the reason was that pairs are not easily hashable (at least to me) and I figured the set wouldn't get too big anyways. But thanks to a recent discussion on Discord I learned that we have our own HashSet implementation :). I'm using this now.

src/search/tasks/negated_axioms_task.cc
```diff
@@ -0,0 +1,358 @@
...
+
+    return 0;
+}
+
+string NegatedAxiomsTask::get_operator_name(int index, bool is_axiom) const {
+    if (!is_axiom || index < negated_axioms_start_index) {
+        return parent->get_operator_name(index, is_axiom);
+    }
+
+    return "<axiom>";
```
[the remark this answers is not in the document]
salome-eriksson: Done.

src/search/tasks/negated_axioms_task.h
salome-eriksson: Removed the comment, thanks!

src/search/tasks/negated_axioms_task.h
salome-eriksson: I removed the method.

src/search/tasks/negated_axioms_task.cc
salome-eriksson: I removed the method.

src/search/tasks/negated_axioms_task.cc
salome-eriksson: I adjusted the functions accordingly. As far as I can tell, axioms still have one precondition, namely the value of the derived variable before the axiom is applied. I specified this as `1 - head.value`, (where `head.value` is the value the derived variable will change to). This only works for binary derived variables, but we have this assumption pretty much everywhere in the code anyways.

src/search/CMakeLists.txt
```diff
@@ -682,6 +682,7 @@ create_fast_downward_library(
     SOURCES
         tasks/cost_adapted_task
         tasks/delegating_task
+        tasks/negated_axioms_task
```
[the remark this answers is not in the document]
salome-eriksson: I changed the structure such that the task is its own library and is added as dependency-only to the relevant libraries.

src/search/heuristic.h
```diff
@@ -55,7 +55,8 @@ class Heuristic : public Evaluator {
     bool cache_evaluator_values;
 
     // Hold a reference to the task implementation and pass it to objects that need it.
-    const std::shared_ptr<AbstractTask> task;
+    // TODO: reinstate const if possible
+    std::shared_ptr<AbstractTask> task;
```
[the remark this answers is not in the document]
salome-eriksson: const is reinstated :)

src/search/axioms.cc
```diff
@@ -22,35 +22,28 @@ AxiomEvaluator::AxiomEvaluator(const TaskProxy &task_proxy) {
...
             EffectProxy cond_effect = axiom.get_effects()[0];
             FactPair effect = cond_effect.get_fact().get_pair();
             int num_conditions = cond_effect.get_conditions().size();
-            // Ignore axioms which set the variable to its default value.
-            if (effect.value != variables[effect.var].get_default_axiom_value()) {
-                AxiomLiteral *eff_literal = &axiom_literals[effect.var][effect.value];
-                axiom_id_to_position[axiom.get_id()] = rules.size();
-                rules.emplace_back(
+
+            // We don't allow axioms that set the variable to its default value.
```
FlorianPommerening, line 31: Should this maybe be an `if (...) { exit_with(INPUT_ERROR) }` instead of an assertion?

src/search/heuristics/cea_heuristic.cc
```diff
@@ -450,6 +454,12 @@ class ContextEnhancedAdditiveHeuristicFeature
     ContextEnhancedAdditiveHeuristicFeature() : TypedFeature("cea") {
         document_title("Context-enhanced additive heuristic");
 
+        add_option<bool>(
```
FlorianPommerening, line 457: I saw this same text in multiple places. For other cases like this, we introduced two static methods to add an option and to parse the arguments from an options object. This way, you only have the documentation in one place,

For example for log objects:
```c++
extern void add_log_options_to_feature(plugins::Feature &feature);
extern std::tuple<Verbosity> get_log_arguments_from_options(
    const plugins::Options &opts);
```

src/search/landmarks/landmark_heuristic.cc
```diff
@@ -27,14 +31,17 @@ void LandmarkHeuristic::initialize(
...
     */
     if (task != tasks::g_root_task
-        && dynamic_cast<tasks::CostAdaptedTask *>(task.get()) == nullptr) {
+        && dynamic_cast<tasks::CostAdaptedTask *>(task.get()) == nullptr
+        && dynamic_cast<tasks::DefaultValueAxiomsTask *>(task.get()) == nullptr) {
         cerr << "The landmark heuristics currently only support "
-             << "task transformations that modify the operator costs. "
-             << "See issues 845 and 686 for details." << endl;
+             << "task transformations that modify the operator costs "
+             << "or add negated axioms. See issues 845 and 686 "
```
FlorianPommerening, line 43: Add issue 454 here?

src/search/heuristics/cea_heuristic.cc
FlorianPommerening: You might also be able to move the logic of `get_default_value_axioms_task_if_needed` there to avoid more code duplication.

FlorianPommerening (commented):
I did a light review on the auxiliary parts of the code (CMake, Options, etc.) I didn't look at the actual implementation in `src/search/tasks` and `src/translate`.

What I didn't quite get from the documentation is what happens if I don't use the option `simple_default_value_axioms`. Will axioms just behave "correctly" (whatever this means) without the option and once I set this option, I approximate this behavior in a way that can only weaken heuristics?

src/search/CMakeLists.txt
```diff
@@ -700,6 +704,18 @@ create_fast_downward_library(
     DEPENDENCY_ONLY
 )
 
+create_fast_downward_library(
+    NAME default_value_axioms_task
+    HELP "Task transformation adding axioms for the default value of derived variables"
+    SOURCES
+        tasks/default_value_axioms_task
+    DEPENDS
```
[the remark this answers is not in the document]
salome-eriksson: I forgot that we had this test; I ran  it now and all libraries passed the dependency test.

src/search/axioms.cc
salome-eriksson: I briefly discussed this with Malte and Gabi, they prefer an assertion here and would rather add checks that terminate the execution in an earlier and more principled way in the future.

src/search/heuristics/cea_heuristic.cc
salome-eriksson: I added the two methods to the `default_value_axioms_task.{cc,h}` files and call it on the relevant places. At the same time I changed the option from a bool to an enum after today's discussion with Gabi and Malte.

I don't understand how to move the logic of `get_default_value_axioms_task_if_needed`. I would imagine that this needs to be a separate method since we need to pass its return value to the constructor of `Heuristic`, but maybe I'm overlooking something?

src/search/landmarks/landmark_heuristic.cc
salome-eriksson: Done

salome-eriksson:
> I did a light review on the auxiliary parts of the code (CMake, Options, etc.) I didn't look at the actual implementation in `src/search/tasks` and `src/translate`.
> 
> What I didn't quite get from the documentation is what happens if I don't use the option `simple_default_value_axioms`. Will axioms just behave "correctly" (whatever this means) without the option and once I set this option, I approximate this behavior in a way that can only weaken heuristics?

The negated axioms are always "correct" in the sense that the heuristic will be safe and stay admissible if it should be (for example hmax). But sometimes the conditions under which a derived variable achieves its default value are just (over)approximated, making the heuristic weaker. If the option is set, then *all* negated axioms are just overapproximated, which is much faster to compute, but can lose a lot of information. If the option is not set, then the overapproximation is only done for some axioms (those whose derived variables have cyclic dependencies, see issue453 for details).

src/search/axioms.cc
```diff
@@ -22,35 +22,28 @@ AxiomEvaluator::AxiomEvaluator(const TaskProxy &task_proxy) {
...
-                    int val = condition.get_value();
-                    AxiomRule *rule = &rules[position];
-                    axiom_literals[var_id][val].condition_of.push_back(rule);
-                }
+            int id = axiom.get_id();
+            EffectProxy effect = axiom.get_effects()[0];
+            for (FactProxy condition : effect.get_conditions()) {
+                int var_id = condition.get_variable().get_id();
+                int val = condition.get_value();
+                AxiomRule *rule = &rules[id];
```
roeger, line 45: Why is this correct? Is there a guarantee that the id of an axiom corresponds to the position in which we get it if we iterated over the axioms via the axioms proxy?

src/search/CMakeLists.txt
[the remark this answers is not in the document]
FlorianPommerening: Sounds good but did you test whether the line break in the string works? I think this way it counts as two arguments to `HELP` and CMake might concatenate them with an extra space (or maybe ignore the second one).

src/search/heuristics/cea_heuristic.cc
FlorianPommerening: We talked about this offline and decided there is nothing to do.

src/search/axioms.cc
salome-eriksson: As I understand it our current code should guarantee it, since ProxyIterators are based on position and the [] operator. But it would certainly be good if someone can give a more confident answer :)

salome-eriksson: Florian answered this on discord, yes the code does guarantee it.

src/search/tasks/default_value_axioms_task.cc
```diff
@@ -0,0 +1,426 @@
...
+     setting var=nondefault, then var'=!val is needed, where
+       - !val=nondefault if val=default
+       - !val=default if val=nondefault
+  (var and var' are always derived variables.)
+
+  If var=default is needed but we already know that the axioms we will
+  introduce for var=default are going to have an empty body, then we don't
+  apply a)/b) (because the axiom for var=default will not depend on anything).
+*/
+unordered_set<int> DefaultValueAxiomsTask::get_default_value_needed(
```
roeger, line 127: I am not totally convinced by the name of the method but struggle to come up with something better.
`determine_variables_with_relevant_default_value` is maybe a bit long.

roeger (approved):
Hi @salome-eriksson,

I had another look at the changes and only left a few comments on cosmetics. From my side,  the code in principle is ready to merge.

Best,
Gabi

src/search/tasks/default_value_axioms_task.cc
```diff
@@ -0,0 +1,426 @@
...
+}
+
+void add_axioms_option_to_feature(plugins::Feature &feature) {
+    feature.add_option<AxiomHandlingType>(
+        "axioms",
+        "How to compute axioms that describe how the negated "
+        "(=default) value of a derived variable can be achieved.",
+        "approximate_negative_cycles");
+}
+tuple<AxiomHandlingType> get_axioms_arguments_from_options(
```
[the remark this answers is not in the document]
salome-eriksson: Agreed (and the style checker seems to not complain about it either ;)).

src/search/tasks/default_value_axioms_task.cc
```diff
@@ -0,0 +1,426 @@
...
+    AxiomHandlingType axioms)
+    : DelegatingTask(parent),
+      axioms(axioms),
+      default_value_axioms_start_index(parent->get_num_axioms()) {
+    TaskProxy task_proxy(*parent);
+
+    /*
+      (non)default_dependencies store for each derived variable v which
+      derived variables appear with their (non)default value in an axiom
+      setting v to their nondefault value.
```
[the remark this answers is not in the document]
salome-eriksson: Sounds good, I changed it.

src/search/tasks/default_value_axioms_task.cc
salome-eriksson: I used the slightly shorter `get_vars_with_relevant_default_value`.

roeger: Sounds good!

PR #221, opened 2024-02-08
Issue879 (deterministic translator output)

roeger:


src/translate/invariants.py
```diff
@@ -324,10 +318,24 @@ def _get_cover_equivalence_conjunction(self, literal):
...
+
+        actions = list(actions_to_check.keys())
+        while actions:
+            # For a better expected perfomance, we want to randomize the order
+            # in which actions are checked. Since candidates are often already
+            # discarded by an early check, we do not want to shuffle the order
+            # but instead always draw the next action randomly from those we
+            # did not yet consider.
+            pos = balance_checker.random.randrange(len(actions))
+            actions[pos], actions[-1] = actions[-1], actions[pos]
```
maltehelmert, line 337: I would only do the one assignment: `actions[pos] = actions[-1]` and not assign the element that we next discard. But if you think it looks clearer this way, no need to change.

maltehelmert: Ah, never mind, we also do this so that we can assign in the pop. Makes sense, ignore the comment.

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -611,15 +610,18 @@ def parse_init(context, alist):
                 if not isinstance(fact, list) or not fact:
                     context.error("Invalid negated fact.", syntax=SYNTAX_LITERAL_NEGATED)
                 atom = pddl.Atom(fact[0], fact[1:])
-                check_atom_consistency(context, atom, initial_false, initial_true, False)
-                initial_false.add(atom)
+                check_atom_consistency(context, atom,
+                                       initial_proposition_values, False)
+                initial_proposition_values[atom] = False
             else:
                 if len(fact) < 1:
```
maltehelmert, line 617: This is a weird test. First, `len(fact) < 1` is a strange way to test for an empty collection. But more importantly, if `fact` can be empty, we will already have crashes in the earlier `if`s that reference `fact[0]`. This is obviously unrelated to the issue, but perhaps it's worth adding a `TODO` here for the future.

maltehelmert: Looking a bit more, there is already the test `if not fact` earlier that invokes `context.error` so this test already happens earlier and we are safe. So rather than a TODO, I suggest we remove this `if` block as a drive-by cleanup, unless I'm reading the code wrong.

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -611,15 +610,18 @@ def parse_init(context, alist):
...
                     context.error(f"Expecting {SYNTAX_LITERAL} for atoms.")
                 atom = pddl.Atom(fact[0], fact[1:])
-                check_atom_consistency(context, atom, initial_true, initial_false)
-                initial_true.add(atom)
-    initial.extend(initial_true)
+                check_atom_consistency(context, atom,
+                                       initial_proposition_values, True)
+                initial_proposition_values[atom] = True
+    initial.extend(atom for atom, val in initial_proposition_values.items()
+                   if val is True)
```
maltehelmert, line 624: It seems we set `initial` to the empty list, never modify it, then extend it here once. So there seems no reason to introduce `initial` before, and we can instead assign here rather than extend. But it's unrelated to the issue, so feel free to do it or leave it.

src/translate/pddl_parser/parsing_functions.py
roeger: This simply makes no sense. Removed.

src/translate/pddl_parser/parsing_functions.py
roeger: In line 605 we insert the assignments to the numeric fluents to initial, so the extend  makes sense here.

PR #222, opened 2024-05-08
add --approximate-negative-always option to translator

ott2:
This option is intended for situations where computation of negative axioms causes timeouts before search even begins. This implements a suggestion of Salomé Eriksson in response to a Discord question by Joan Espasa on 18 March 2024, and wraps it in a new commandline option for the translator. Malte Helmert has suggested a similar approach in the past, commenting out the computation of negative axioms, but it is not clear to us precisely when that is sound so this implements the approximation approach instead.

We use FD with this option for our models of gravity-based tile matching puzzles to avoid unnecessary timeouts. Stock FD seldom gets past the negative axiom generation phase within the timeout, even for small instances. In contrast, with this option FD with a blind heuristic and a fine-grained PDDL model is often our best solver, usually beating human experts: see our [ModRef 2023 paper](https://modref.github.io/papers/ModRef2023_TowardsAModelOfPuzznic.pdf).

### DIFF
--- a/src/translate/axiom_rules.py
+++ b/src/translate/axiom_rules.py
@@ -62,6 +62,9 @@ def handle_axioms(operators, axioms, goals, layer_strategy):
     # the future. Similarly, it would be a good idea to remove the notion of
     # axiom layers and derived variable default values from the output.
     # (All derived variables should be binary and default to false.)
+# Malte Helmert suggested commenting out the following 2 lines
+# in conjunction with the blind heuristic this is sound
+# we instead introduce a commandline option changing compute_negative_axioms()
     with timers.timing("Computing negative axioms"):
         compute_negative_axioms(clusters)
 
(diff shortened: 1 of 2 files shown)


ott2:
Adding @JoanEspasa.

roeger:
Sorry that nobody reacted here. This is probably superseded by pull request #220, where we move the computation of negated axioms from the translator to the search component and only perform the computation if the heuristic requires it.

maltehelmert:
@ott2 or @JoanEspasa Can you point us to some PDDL files you are interested in that we could use as a performance test for this?

ott2:
@maltehelmert I will investigate PR #220 to see whether @salome-eriksson's new approach addresses our performance bottleneck and report back. We have a few small PDDL files which show the dramatic difference in case Salome's approach isn't enough.

ott2:
The following PDDL domain/problem pair illustrates our motivation.

[domain-pddl.txt](https://github.com/user-attachments/files/16184192/domain-pddl.txt)
[5x7-pddl.txt](https://github.com/user-attachments/files/16184191/5x7-pddl.txt)

Salome's #220 modified FD performs well, possibly a tiny bit faster than the brute force approach in this PR:
```
salome-downward/fast-downward.py --search-time-limit 3600 domain-pddl.txt 5x7-pddl.txt --search 'astar(blind())'
```

The code in this PR is run as:
```
stacs-downward/fast-downward.py --search-time-limit 3600 domain-pddl.txt 5x7-pddl.txt --search 'astar(blind())' --translate-options --approximate-negative-always
```

In contrast the original code as of hash 61646d7e7e70bcd4319190f2175578d7d86b6b8f times out for me during generation of negative axioms.

PR #226, opened 2024-07-17
Issue730

FlorianPommerening:


src/search/command_line.cc
```diff
@@ -184,6 +185,147 @@ shared_ptr<SearchAlgorithm> parse_cmd_line(
...
+    } else if (last_arg == "--internal-plan-file") {
+        // Suggest filename starting with current_word
+        complete_filename(prefix, suggestions);
+    } else if (last_arg == "--internal-previous-portfolio-plans") {
+        // no suggestions, integer expected
+    } else if (last_arg == "--search") {
+        // suggestions in search string based on current_word
+        plugins::Registry registry = plugins::RawRegistry::instance()->construct_registry();
+        for (const shared_ptr<const plugins::Feature> &feature : registry.get_features()) {
+            suggestions.push_back(feature->get_key() + "(");
```
salome-eriksson, line 212: For me this shows on the command line as "adapt_costs\(", and if I use the autocomplete I also get the backslash.
Furthermore, this only works for the "outermost" part, i.e. if I already entered "astar(" then it won't suggest anything anymore (but I presmue you know this).

TABCOMPLETION.md
```diff
@@ -0,0 +1,34 @@
...
+or pip
+
+```bash
+pip install argcomplete
+```
+
+We recommand to globally enable tab completion for argcomplete as follows. This will enable argcomplete for all python programs which use argcomplete. You can find local alternatives in the [argcomplete documentation](https://pypi.org/project/argcomplete/#activating-global-completion).
+
+```bash
+activate-global-python-argcomplete
```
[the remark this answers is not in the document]
speckdavid: @FlorianPommerening was it the same for you with the `3` in `activate-global-python-argcomplete`?

We may want to recommend `activate-global-python-argcomplete3 --user`. This should avoid the need of `sudo`.

src/search/command_line.h
```diff
@@ -3,11 +3,13 @@
 
 #include <memory>
 #include <string>
+#include <vector>
```
SimonDold, line 6: double include in .cc and .h

TABCOMPLETION.md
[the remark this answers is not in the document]
FlorianPommerening: I think I also had `activate-global-python-argcomplete` but it looks like this changed:


22.04: https://packages.ubuntu.com/jammy/all/python3-argcomplete/filelist
24.04: https://packages.ubuntu.com/noble/all/python3-argcomplete/filelist

I would handle the user-level installation as we do the other "local" case: say that the option `--user` exists for a user-level installation but don't go into detail.

driver/tab_completion.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def complete_build_arg(prefix, parsed_args, **kwargs):
+    if parsed_args.debug:
+        argcomplete.warn("The option --debug is an alias for --build=debug. Do no specify both --debug and --build.")
+        exit(1)
+
+    builds_folder = Path(util.REPO_ROOT_DIR) / "builds"
+    if not builds_folder.exists():
+        argcomplete.warn("No build exists.")
```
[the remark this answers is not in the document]
FlorianPommerening: We don't know what to check for at this point: we might just need the translator and it would be fine if there is no search binary (or the other way around). We later check for the tool we need and complain if it is not there.

driver/tab_completion.py
```diff
@@ -0,0 +1,122 @@
...
+    num_planner_args = len(planner_args)
+    mode_switches = ["--translate-options", "--search-options"]
+    last_option_was_mode_switch = planner_args and (planner_args[-1] in mode_switches)
+    double_dash_in_options = "--" in planner_args
+
+    current_mode = util.split_planner_args(parsed_args)
+    num_filenames = len(parsed_args.filenames)
+    has_only_filename_options = (num_filenames == num_planner_args)
+    has_only_filename_or_double_dash_options = (num_filenames + int(double_dash_in_options) == num_planner_args)
+    can_use_double_dash = (1 <= num_planner_args <= 2) and has_only_filename_or_double_dash_options
```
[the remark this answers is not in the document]
FlorianPommerening: It's ... complicated. But I guess we should document this better. 

The double dash can be used to separate filename options from the other options. Say your domain file is called `--debug` for some reason and your problem file is obviously called `--`. To call the planner correctly, you'd use `./fast-downward.py --debug -- -- --search "astar(lmcut)"` The last double dash is removed and everything before it is treated as a filename. This gets complicated because our partial parse might only have the first double dash that is meant to be a filename but since it is currently the last double dash, it is considered a separator. So even though we already have a separator, we still have to suggest one  because there is an interpretation of the command line where we could still need one. Together with the fact that we can take either one or two input files, this creates this mess.

driver/tab_completion.py
```diff
@@ -0,0 +1,122 @@
...
+    comp_line = " ".join(simulated_commandline)
+    comp_point = str(len(comp_line))
+    comp_cword = str(len(simulated_commandline) - 1)
+    cmd = [str(downward), "--bash-complete",
+           comp_point, comp_line, comp_cword] + simulated_commandline
+    output = subprocess.check_output(cmd, text=True)
+    return output.split()
+
+
+def get_completions_from_translator(translator, options, prefix):
```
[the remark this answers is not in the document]
FlorianPommerening: We are trying to simulate what would happen if we had pressed tab in the `simulated_commandline` at the position `comp_point`. For that we get the parser by importing the translator (lines 100-104) and using its code to set up its parser (lines 108-109). The we run a `CompletionFinder` which is what `argcomplete.autocomplete()` does, but we patch it to return its completions rather than printing them and exiting the program.

driver/tab_completion.py
```diff
@@ -0,0 +1,122 @@
...
+        bin_dir = Path(util.REPO_ROOT_DIR) / "builds" / build / "bin"
+        if current_mode == "search":
+            if not last_option_was_mode_switch:
+                completions.append("--translate-options")
+
+            downward = bin_dir / "downward"
+            if downward.exists():
+                completions += get_completions_from_downward(
+                    downward, parsed_args.search_options, prefix)
+        else:
```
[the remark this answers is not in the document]
FlorianPommerening: Yes, that should work. Can you check what versions of Python and argcomplete you have? Some combinations are incompatible.

src/search/command_line.cc
FlorianPommerening: This is a very naive way of adding completions. For now, we will leave the `--search` argument mostly as it is. It will be more complicated to get this working.

driver/tab_completion.py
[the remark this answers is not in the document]
maltehelmert: In case it's not generally known: "--" acting as a separator between options (which always start with "-") and other arguments (which only rarely start with "--", but are in principle unrestricted) is a general though not universal Unix command-line parsing feature. Tools like argparse handle "--" in this fashion automatically, and because this is expected behavior for all "modern" programs, it's not usually documented to the user.

In the driver we have the special situation that we have to infer *two* separators in the command-line options (between driver options and filenames and between filenames and component options), which is why we extend this convention further.

build.py
```diff
@@ -114,25 +142,28 @@ def build(config_name, configure_parameters, build_parameters):
...
-        build(config_name, CONFIGS[config_name], build_parameters)
+            args.build_parameters.append(arg)
+
+    if args.debug:
+        args.config_names.append(DEBUG_CONFIG_NAME)
+    if args.all:
+        args.config_names.extend(sorted(CONFIGS.keys()))
+    if not args.config_names:
+        args.config_names.append(DEFAULT_CONFIG_NAME)
+
```
salome-eriksson, line 158: It might make sense to eliminate duplicates in `args.config_names`. If I call `./build.py debug debug`, it will build the debug config two times. (Of course, the second time it will be very fast since everything is built already.)

build.py
```diff
@@ -34,37 +62,37 @@ def print_usage():
...
-  ./{script_name} -j4                 # build {default_config_name} in 4 threads
-  ./{script_name} debug               # build debug
-  ./{script_name} --debug             # build {debug_config_name}
-  ./{script_name} release debug       # build release and debug configs
-  ./{script_name} --all VERBOSE=true  # build all build configs with detailed logs
-""")
-
+example usage:
+  {script_name}                     # build {DEFAULT_CONFIG_NAME} with #cores parallel processes
+  {script_name} -j4                 # build {DEFAULT_CONFIG_NAME} with 4 parallel processes
```
salome-eriksson, line 70: Trying this command leads to "build.py: error: unrecognized arguments: -j4" for me; but `build.py release -j4` for example works.

maltehelmert: The problem is likely that build.argparse tries to recognize `-j4` as options rather than positional arguments (cf. the `--` discussion we recently had; I expect `./build.py -- -j4` would also work, but of course we want to make it easier than that). Using `argparse.parse_known_args` may or may not be good enough to address this.

build.py
maltehelmert: Trying to build it twice is exactly what the user asks for with this call. I think it's not a good feature to try to "optimize" this.

TABCOMPLETION.md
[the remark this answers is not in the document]
speckdavid: We changed the instructions according to a discussion. We now recommend local activation of argcomplete.

driver/tab_completion.py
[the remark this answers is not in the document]
salome-eriksson: I just meant if no folders (like "release" or "debug") exist within the "builds" folder. But it's fine as it is.

src/search/command_line.cc
speckdavid: We completely disabled this now since it is not really helpful.

driver/tab_completion.py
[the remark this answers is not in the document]
salome-eriksson: Python 3.10.12 and  argcomplete 1.8.1. (That is my work computer where I installed it today and it doesn't work either; I can't check on my home computer at the moment). I do want to note however that tab completion works on everything else I tested.

build.py
salome-eriksson: `./build.py -- -j4` doesn't give me an error but It also does not do anything (i.e. no output whatsoever).

speckdavid: We fixed it using`argparse.parse_known_args` as suggested.

driver/tab_completion.py
```diff
@@ -0,0 +1,122 @@
...
+    has_only_filename_or_double_dash_options = (num_filenames + int(double_dash_in_options) == num_planner_args)
+    can_use_double_dash = (1 <= num_planner_args <= 2) and has_only_filename_or_double_dash_options
+
+    completions = []
+
+    if can_use_double_dash:
+        completions.append("--")
+
+    if parsed_args.filenames or double_dash_in_options:
+        bin_dir = Path(util.REPO_ROOT_DIR) / "builds" / build / "bin"
```
maltehelmert, line 59: It's not good that information about where certain files are located is now repeated in many places in the code. We already have a central place for this and shouldn't reimplement this logic.

src/search/command_line.cc
```diff
@@ -184,6 +184,118 @@ shared_ptr<SearchAlgorithm> parse_cmd_line(
...
+    }
+    int cursor_pos = parse_int_arg("COMP_POINT", static_cast<string>(argv[2]));
+    const string command_line = static_cast<string>(argv[3]);
+    int cursor_word_index = parse_int_arg("COMP_CWORD", static_cast<string>(argv[4]));
+    vector<string> words;
+    words.reserve(argc - 4);
+    for (int i = 5; i < argc; ++i) {
+        words.emplace_back(argv[i]);
+    }
+    // Sentinal for cases where the cursor is after the last word.
```
maltehelmert, line 259: Typo: "Sentinel".

TABCOMPLETION.md
```diff
@@ -0,0 +1,30 @@
...
+
+function _downward_complete() {
+    local IFS=$'\n'
+    COMPREPLY=( $( "$1" --bash-complete \
+                   "$COMP_POINT" "$COMP_LINE" "$COMP_CWORD" ${COMP_WORDS[@]}))
+}
+complete -o nosort -o default -o bashdefault -F _downward_complete downward
+```
+
+Restart your shell afterwards.
```
maltehelmert, line 24: Do we need to say if/how this affects people that also have other versions of "fast-downward.py" etc. somewhere? Will this break (tab completion for) other programs with the same name? I can imagine that other people have scripts called "build.py", for example, perhaps even "translate.py". If there is a possible interaction, I think it's enough to warn about it and then leave it for people to decide if they care. I don't think we need to be more elaborate (e.g. propose workarounds/solutions).

build.py
```diff
@@ -24,7 +32,27 @@
     # Number of available CPUs as a fall-back (may be None)
     NUM_CPUS = os.cpu_count()
 
-def print_usage():
+
+class RawHelpFormatter(argparse.HelpFormatter):
+    """Preserve newlines and spacing."""
+    def _fill_text(self, text, width, indent):
+        return ''.join([indent + line for line in text.splitlines(True)])
```
maltehelmert, line 39: We used to generally use `""` rather than `''`, although I'm not sure if this ship has sailed and we're now mixing these freely. (I think in the past I tried to be consistent about this at least in the driver and translator, but not for example in the tests. But we may have given up on consistency for this already.)

build.py
```diff
@@ -34,36 +62,43 @@ def print_usage():
...
+                        help="alias to build all build configurations")
+    remaining_args = parser.add_argument("arguments", nargs=argparse.REMAINDER,
+                        help="build configurations (see below) or build options")
+    remaining_args.completer = complete_arguments
+
+    if HAS_ARGCOMPLETE:
+        argcomplete.autocomplete(parser)
+    args, unparsed_args = parser.parse_known_args()
+
+    if hasattr(args, "remaining_args"):
```
maltehelmert, line 91: Can this case ever arise? How?

build.py
```diff
@@ -34,36 +62,43 @@ def print_usage():
...
+        args.remaining_args += unparsed_args
+    else:
+        args.remaining_args = unparsed_args
+    return args
+
+
+def complete_arguments(prefix, parsed_args, **kwargs):
+    split_args(parsed_args)
```
maltehelmert, line 99: This looks like it modifies parsed_args in place. Should `complete_arguments` do this as a side effect? If yes, should we document this somewhere?

maltehelmert:
A few general points that we discussed live:

- This changes the translator to no longer be fully black box. There are good reasons for and against this, and perhaps we want to discuss this more. If we keep it this way, I think the parts of the translator that are meant to be used as a library from the outside should say this clearly in the py file (if that's not already the case) so that in future changes where we touch these files, we're aware that they are used outside of the regular use of the translator.

- We are generally happy to convert (const) char * to string implicitly when calling a function that takes a string, so using static_cast<std::string>(foo) is a bit of a stylistic digression. If you prefer to be explicit, personally I'd be much happier to write this using constructor call syntax "string(foo)" because casts always read a bit like a warning flag, but no warning flag is justified in the relevant places.

- A few new functions look quite long to me; I wonder if it would make sense to break them up.

- In the driver, I think it would be good to keep the information about where the various executables are located in one place. The new code replicates this in a few places. (I think we already have a canonical place for this information, but I'd have to search what it is.)

driver/tab_completion.py
```diff
@@ -0,0 +1,122 @@
...
+    import argcomplete
+    HAS_ARGCOMPLETE = True
+except ImportError:
+    HAS_ARGCOMPLETE = False
+
+
+def complete_build_arg(prefix, parsed_args, **kwargs):
+    if parsed_args.debug:
+        argcomplete.warn("The option --debug is an alias for --build=debug. Do no specify both --debug and --build.")
+        exit(1)
```
maltehelmert, line 19: The "argcomplete.warn + exit" combo occurs at least three times, so I think we should encapsulate it in a function. The exit code shouldn't be 1; see "https://www.fast-downward.org/ExitCodes". The closest match is perhaps DRIVER_INPUT_ERRROR, which should already be defined as a symbolic constant somewhere.

driver/tab_completion.py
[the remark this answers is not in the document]
maltehelmert: "Check for existence, then used" (look before you leap, LBYL)  is considered an anti-pattern in Python (and not just in Python) because the folder might exist now but be gone by the time we try to access it. The preferred style is to attempt the access and handle an exception if it doesn't work. See for example here:

https://realpython.com/python-lbyl-vs-eafp/#working-with-files-and-directories

driver/tab_completion.py
```diff
@@ -0,0 +1,122 @@
...
+
+def get_completions_from_downward(downward, options, prefix):
+    simulated_commandline = [str(downward)] + options + [prefix]
+    comp_line = " ".join(simulated_commandline)
+    comp_point = str(len(comp_line))
+    comp_cword = str(len(simulated_commandline) - 1)
+    cmd = [str(downward), "--bash-complete",
+           comp_point, comp_line, comp_cword] + simulated_commandline
+    output = subprocess.check_output(cmd, text=True)
+    return output.split()
```
maltehelmert, line 92: Will the completions never contain spaces? For example if the user included a space in the argument to be completed?

In shell scripts, one would normally try to deal with this by using newlines for splitting (you often see IFS set to newline for this reason), which is also potentially problematic because arguments can use newlines, but is an improvement over splitting on all whitespace.

src/search/command_line.cc
```diff
@@ -184,6 +184,118 @@ shared_ptr<SearchAlgorithm> parse_cmd_line(
...
+            "$COMP_POINT \"$COMP_LINE\" $COMP_CWORD ${COMP_WORDS[@]}\n"
+            "where the environment variables have their usual meaning for bash completion:\n"
+            "$COMP_POINT is the position of the cursor in the command line.\n"
+            "$COMP_LINE is the current command line.\n"
+            "$COMP_CWORD is an index into ${COMP_WORDS} of the word under the cursor.\n"
+            "$COMP_WORDS is the current command line split into words.\n"
+            );
+    }
+    int cursor_pos = parse_int_arg("COMP_POINT", static_cast<string>(argv[2]));
+    const string command_line = static_cast<string>(argv[3]);
```
maltehelmert, line 252: We generally don't use `const` for local variables like these (e.g. we don't use `const int` in the line above).

src/search/command_line.cc
```diff
@@ -184,6 +184,118 @@ shared_ptr<SearchAlgorithm> parse_cmd_line(
...
+    int cursor_word_index = parse_int_arg("COMP_CWORD", static_cast<string>(argv[4]));
+    vector<string> words;
+    words.reserve(argc - 4);
+    for (int i = 5; i < argc; ++i) {
+        words.emplace_back(argv[i]);
+    }
```
maltehelmert, line 258: If you like, you could replace the loop with a single assignment:
`words.assign(argv + 5, argv + argc);`
or if you prefer:
`words.assign(&argv[5], &argv[argc]);`.
I would even consider replacing all of lines 254-258 like this:
`vector<string> words(argv + 5, argv + argc);`.
This loses the reserve so can lead to one extra allocation for the sentinel ("assign"/the constructor will take care not to do unnecessary reallocations for the parts it assigns), but this is not really an issue here.

src/search/command_line.cc
```diff
@@ -184,6 +184,118 @@ shared_ptr<SearchAlgorithm> parse_cmd_line(
...
+        pos += word_len;
+        cursor_pos -= word_len;
+        parsed_args.push_back(current_word);
+
+        // Skip whitespace between words.
+        while (pos < end && isspace(command_line[pos])) {
+            ++pos;
+            --cursor_pos;
+        }
+    }
```
maltehelmert, line 291: I don't fully understand the code, but is it OK that the test of the consistent cursor_pos only happens when we break out of the loop, and there is no test if the loop completes without breaking out?

maltehelmert:
Like we discussed, I added comments for the points I noticed, but this is not a code review because I didn't check most of the larger cases of moved code, I skipped over parts I couldn't easily follow, and I didn't think much about the bigger picture, only/mostly comments that were very local to the code I looked at, such as style things.

build.py
FlorianPommerening: There are a couple of `''` in the driver, mostly in the portfolio definitions. We also use them for output like `print("Expected ')'")` but this is a different situation. Overall, it is pretty consistently `""` but there are a couple of cases that should be changed besides this one (this particular use is copied from existing code in the driver). I'll only change the parts that the issue touches anyway to avoid increasing the diff.

build.py
FlorianPommerening: This happens with the command that Salome posted above: `build.py release -j4`. Our fix was to use `parse_known_args` instead which means that `remaining_args` remains non-existent and `-j4` ends up in `unparsed_args`. I'll add a comment

FlorianPommerening: Ahh, I see you meant the other case. You're right, this should not happen because everything except the known options (`--all`, `--debug`) would end up in `unparsed_args`. In that case, having the option with `nargs=argparse.REMAINDER` is redundant. We could remove it but it would affect the generated help text: the description of the positional arguments was bad anyway and we could improve it this way, but the "using" statement would lose the meta vars `[BUILD ...] [CMAKE_OPTION ...]` in the end. Is that OK?

FlorianPommerening: I guess we have to keep the argument because this is where our argument completer attaches to. Removing it would not be an option.

FlorianPommerening: I keep finding problems with this code: `remaining_args` is the name of the argparse representation of the argument, not the name of the parsed values. Argparse always creates the actual value, so the whole check isn't necessary, sorry for the noise. Here is the updated version:

```Python
    # With calls like 'build.py release -j4' the parser would try to interpret
    # '-j4' as an option and fail to parse. We use parse_known_args to parse
    # everything we recognize and add everything else to the list of arguments. 
    args, unparsed_args = parser.parse_known_args()
    args.arguments += unparsed_args
```

build.py
FlorianPommerening: It is fine to do this because the process will stop after generating suggestions for the tab completion. I'll add a comment.

build.py
maltehelmert: That's the direction I meant, yes. But the example is wrong, unless I'm now really confused. `build.py release -j4` would work with regular argparse parsing. After the first positional argument, argument parsing no longer looks for options, so the example would be fine because seeing `release` means that we're now past the options and everything that follows is a positional argument. The failing cases would be those without explicit builds listed but with CMake options, for example `./build.py -j4`.

Note that if we switch to `parse_known_args`, there is a difference of behaviour to the old code: in the old code "./build.py -j4 --all" would be illegal because we expect the CMake options at the end. With `parse_known_args`, they can occur anywhere, which isn't necessarily a problem for build.py, but something to be aware of. (This may be the reason why we avoided argparse for build.py before this pull request.)

In the translator one reason we decided to go for the syntax "driver_options filenames component_options" is because the filenames always put a buffer in between the two types of options, so we can use the regular argparse behaviour of assuming the parsed options end the first time we encounter an argument not starting with "-". (Of course excluding arguments that are parameters of options, like the 400s in `--overall-time-limit 400s`.) Part of the rationale is that if someone writes a typo as in `--overall-timelimit 400s`, we don't want this to be silently converted to a search component option, which would run the translator and then abort search with a confusing error message.

I don't mind mixing the different kinds of arguments in build.py, as it has a simpler syntax than fast-downward.py, but I think the proposed code has another, more subtle bug. Normally options are parsed from left to right and can override each other, so if I call "make -j4 -j8", this means using 8 jobs (not tested, but this is the general convention because it makes it possible to do stuff like `alias make=make -j4` to set a default while still allowing this default to be overridden). With the proposed code, if I write `./build.py -j4 --debug -j8`, we would set `args.arguments` to `["-j8", "-j4"]`, inverting the order. The values we get from `unparsed_args` will always appear in the command line before the values we get from the positional `args.arguments`, so I would replace the `+=` with something like `args.arguments = unparsed_args + args.arguments` if we want to recycle `args.arguments` in the first place and not assign the concatenation to a fresh variable.

(This could be done in-place as `args.arguments[0:0] = unparsed_args`, but this looks too clever.)

maltehelmert: Actually, I'm not 100% sure if it's still the case that the first positional arguments finishes the search for options if `parse_known_args` is used. So not sure if it is indeed the case that we can know in which order the arguments occurred. (I think I know how the "usual" arg parsing mode works, but I'm a bit less sure about `parse_known_args`.)

maltehelmert: OK, I tested it: indeed also with known arg parsing, once something looks like a positional argument, the rest will show up in the remainder.

Test code:
```
#! /usr/bin/env python3
# -*- coding: utf-8 -*-

import argparse

parser = argparse.ArgumentParser()
parser.add_argument("--debug", action="store_true")
parser.add_argument("--all", action="store_true")
parser.add_argument("arguments", nargs=argparse.REMAINDER)
args, unparsed_args = parser.parse_known_args()

print(f"args.arguments: {args.arguments}")
print(f"unparsed_args: {unparsed_args}")
```

Example run:
```
$ ./foo.py -j4 --debug foo --all bar
args.arguments: ['foo', '--all', 'bar']
unparsed_args: ['-j4']
```
So `foo` causes the rest to appear in the remainder `args.arguments`, even though a legal option (`--all`) appears later. This means that indeed all arguments in `unparsed_args` must come before all arguments in `args.arguments`.

FlorianPommerening: Indeed `./build.py -j4 --all` was legal before argparse, so I think it is good that it is legal again with using `parse_known_args`. I went with `args.arguments = unparsed_args + args.arguments`

TABCOMPLETION.md
speckdavid: Indeed, the local installation messes with old version or other unrelated files with the same names. The global installation check for a magic comment `# PYTHON_ARGCOMPLETE_OK` and only executes the file if this is found. Unfortunately, the local version does not and simply executes the file when pressing tab. With older versions of build.py it triggers a build. We have the following options:

1. Leave as is and warn about it in the description.
2. Recommend a global activation. This is more intrusive and can cause bugs due to incompatible versions such that sometimes file names are recommended additionally to valid options.
3. Write our own local register function which additionally checks for the magic comment `# PYTHON_ARGCOMPLETE_OK`.

Any thoughts?

driver/tab_completion.py
speckdavid: The surrounding code also uses newlines (\n) as separators. We changed from `split` to `splitlines`. So far, the search component does not return any newlines in the suggestions. Once we implement the suggestions for `--search`, it could be part of the prefix and we will deal with it.

src/search/command_line.cc
speckdavid: With the new code, the test is now always performed.

driver/tab_completion.py
speckdavid: Yes, but it probably makes sense to unify it after merging the pull request #228 which adds `pathlib` support.

TABCOMPLETION.md
FlorianPommerening: We discussed this offline and will go with option 1.

driver/tab_completion.py
FlorianPommerening: This place now uses the existing function. Malte mentioned "many places". Are there other places than this that we should fix?

maltehelmert: In the current version, I see only one other somewhat related place, although it's not quite the same thing. I'll add a comment there.

driver/tab_completion.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def _call_argcomplete(python_file, comp_line, comp_point):
+    with tempfile.NamedTemporaryFile(mode="r") as f:
+        env = os.environ.copy()
+        env["COMP_LINE"] = comp_line
+        env["COMP_POINT"] = str(comp_point)
+        env["_ARGCOMPLETE"] = "1"
+        env["_ARGCOMPLETE_STDOUT_FILENAME"] = f.name
+        subprocess.check_call([sys.executable, python_file], env=env)
```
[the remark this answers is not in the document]
FlorianPommerening: The only other place where the translator is called that I could find was in `run_components.py` where it is also called with `sys.executable` (https://github.com/aibasel/downward/blob/main/driver/run_components.py#L65). We cannot simply reuse this code, because it does a lot of additional stuff: setting up resource limits, handling how output is printed, etc. It also takes an `args` argument that is assumed to have arguments for the call in place. Internally, it uses `get_error_output_and_returncode` which is meant for user-facing execution (e.g., it prints the call string to the log which would be a problem for us here because it would be interpreted as a completion suggestion). I don't think the overlap is big enough that we have to extract some common functionality here.

As to the point about using the shebang: I thought the shebang is only considered if you use subprocess with `shell=True` which I thought to be a security red flag. I see the point of overwriting the shebang this way, but I don't think this is necessarily a problem. Say I have a Python venv and call the driver with the Python version from that venv (without activating it). Then the translator keeps using the venv, which I wouldn't find surprising as a user who doesn't know about the internals of Fast Downward. I think what I'm trying to argue is that if `sys.executable` and the executable on the PATH are not the same, then the user already specified a preference for one over the other in the call to the driver so it is OK to keep using it.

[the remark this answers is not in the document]
maltehelmert: Thanks! I looked one level less deeply into the code; not into the code that calls the translator but the code that finds the translator. Because that function was called "get_executable", I assumed that this is then used as an executable, but actually it isn't any more. I see now that this was changed as part of issue733 in order to be able to do experiments that compare the translator on different Python versions by calling the driver with different Python versions.

No, the shebang and shell=True are not related. (And indeed shell=True needs a lot of care to avoid security issues.) Linux has no notion of "file extensions" inside, a filename is just a string. So if I run a file as an executable, Linux needs some way to figure out how to run it. So it looks inside the file. If the file has a recognized binary format (usually ELF), then the kernel knows how to run ELF binaries. Otherwise it looks at the first line to find out which other executable to use in order to invoke the program. It's an OS mechanism, not a shell mechanism. (This is different from Windows, which registers specific file extensions with specific programs and has no real notion of an executable script.)

I'm OK with continuing to use get_executable, though I do find it surprising behaviour. (I understand the use case from issue733, although it could also have been achieved by setting PATH.) I don't know other programs that behave in this way when calling another Python program that is intended as a separate executable. Because for me it's a very unintuitive thing to do, I don't like encoding this logic twice.

I think as a user I would be surprised if a venv takes precedence over the installed version if I don't activate it. Perhaps this depends on whether you think of the driver and translator as independent programs.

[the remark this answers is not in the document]
maltehelmert: (BTW, looks like Linux now also has a registration mechanism that *can* use file extensions as part of the `binfmt_misc` infrastructure. Not relevant to what we're doing here, but it means that "Linux has no notion of file extensions" is not true; binfmt_misc does support dispatch based on file extensions, although I haven't seen an example of this in use.)

driver/tab_completion.py
```diff
@@ -0,0 +1,120 @@
...
+    entry_separator, help_separator = _get_field_separators(os.environ)
+    help_separator = help_separator or ":"
+    simulated_commandline = [str(downward)] + options + [prefix]
+    comp_line = " ".join(simulated_commandline)
+    comp_point = str(len(comp_line))
+    comp_cword = str(len(simulated_commandline) - 1)
+    cmd = [str(downward), "--bash-complete", entry_separator, help_separator,
+           comp_point, comp_line, comp_cword] + simulated_commandline
+    output = subprocess.check_output(cmd, text=True)
+    return _split_argcomplete_ouput(output, entry_separator, help_separator)
```
maltehelmert, line 108: typo `ouput`

driver/tab_completion.py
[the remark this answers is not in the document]
maltehelmert: In any case, I see why you find this difficult to untangle. In my view, part of the reason for this is that the {_,}get_..._executable functions don't have a very good design with the way they handle errors in two different ways depending on an argument passed in. I think in a cleaner design they would not have this argument and should instead raise an exception in cases where they fail. These exceptions could then be handled in different ways by the different callers, although and the `if downward` etc. tests avoided. Then if we want to call the translator as `[sys.executable, translator_python_file]`, I'd have this list be what `get_translator_executable` (perhaps with a different name; it's not a good name currently as described, but also not a good name if it returns this list) returns. And rather than a method for calling generic Python code, I wouldn't parameterize it with the translator source file, but always keep the information on how to call the translator as a unit.

driver/tab_completion.py
```diff
@@ -0,0 +1,120 @@
...
+
+    if can_use_double_dash:
+        completions["--"] = ""
+
+    if parsed_args.filenames or double_dash_in_options:
+        if current_mode == "search":
+            completions["--translate-options"] = ""
+
+            downward = get_search_executable(parsed_args.build, exit_on_failure=False)
+            if downward and downward.exists():
```
maltehelmert, line 48: Isn't the same `exists` test already done by `get_search_executable`?

driver/tab_completion.py
```diff
@@ -0,0 +1,120 @@
...
+                completions.update(_get_completions_from_downward(
+                    downward, parsed_args.search_options, prefix))
+        else:
+            completions["--search-options"] = ""
+
+            translator = get_translate_executable(parsed_args.build, exit_on_failure=False)
+            if translator and translator.exists():
```
maltehelmert, line 55: Isn't the same `exists` test already done by `get_translate_executable`?

driver/tab_completion.py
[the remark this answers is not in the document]
FlorianPommerening: We tried to write another version that uses `get_*_command()` instead of `get_*_executable()` and trows exceptions there. This now has the bundling of `[sys.executable, translate.py]` at only one place, but now we repeat some error handling code instead. We also moved the lookup of the executable further inside to `_get_completions_from_translator`. We didn't move it here (`_call_argcomplete`), because the surrounding function `_get_completions_from_translator`also needs it to compute the line for which we want to get completions.

TABCOMPLETION.md
FlorianPommerening: Now that the argcomplete version that apt installs on Ubuntu 24.04 is compatible with their Python version, we decided to describe both local and global activation. We discuss side effects and potential problems with both variants.

driver/tab_completion.py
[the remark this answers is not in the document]
FlorianPommerening: The commit that addresses these comments is https://github.com/aibasel/downward/pull/226/commits/a239859f653d779ca6b9b1b534c4217c5dd094c2. Does it resolve this comment?

PR #227, opened 2024-07-18
[issue1146] Validate SAS file

tanjaschindler:


src/search/tasks/root_task.cc
```diff
@@ -179,84 +178,104 @@ ExplicitEffect::ExplicitEffect(
...
+        in.error("expected boolean");
+    }
+    in.read_magic_line("end_metric");
     return use_metric;
 }
 
-static vector<ExplicitVariable> read_variables(istream &in) {
-    int count;
-    in >> count;
+static vector<ExplicitVariable> read_variables(utils::InputFileParser &file_parser) {
```
FlorianPommerening, line 258: The parser is sometimes called `in` and sometimes `file_parser`. I'm fine with both but would prefer it to be consistent.

src/search/tasks/root_task.cc
```diff
@@ -179,84 +178,104 @@ ExplicitEffect::ExplicitEffect(
...
+    file_parser.set_context("variable_section");
+    int count = file_parser.read_line_int("variable count");
+    if (count < 1) {
+        file_parser.error("Number of variables is less than 1, should be at least 1.");
```
FlorianPommerening, line 262: I'd prefer something like "Number of variables should be at least 1 but is 0." (i.e., a full sentence including the actual value).

src/search/tasks/root_task.cc
```diff
@@ -266,18 +285,24 @@ static vector<vector<set<FactPair>>> read_mutexes(istream &in, const vector<Expl
       aware of.
     */
     for (int i = 0; i < num_mutex_groups; ++i) {
-        check_magic(in, "begin_mutex_group");
-        int num_facts;
-        in >> num_facts;
+        file_parser.read_magic_line("begin_mutex_group");
+        int num_facts = file_parser.read_line_int("number of facts in mutex group");
+        if (num_facts < 1) {
+            file_parser.error("Number of facts in mutex group is less than 1, should be at least 1.");
```
FlorianPommerening, line 291: See above.

src/search/tasks/root_task.cc
```diff
@@ -266,18 +285,24 @@ static vector<vector<set<FactPair>>> read_mutexes(istream &in, const vector<Expl
...
+        }
         vector<FactPair> invariant_group;
         invariant_group.reserve(num_facts);
         for (int j = 0; j < num_facts; ++j) {
```
FlorianPommerening, line 295: Don't we already have a function that reads a list of facts? And doesn't it also need the check that there are no duplicate facts?

src/search/tasks/root_task.cc
```diff
@@ -311,10 +337,10 @@ static vector<FactPair> read_goal(istream &in) {
 }
 
 static vector<ExplicitOperator> read_actions(
-    istream &in, bool is_axiom, bool use_metric,
+    utils::InputFileParser &in, bool is_axiom, bool use_metric,
     const vector<ExplicitVariable> &variables) {
-    int count;
-    in >> count;
+    in.set_context(is_axiom ? "axiom section" : "operator section");
```
FlorianPommerening, line 342: You could use a single `if ... else ...` here instead of two `?`-operators.

src/search/tasks/root_task.cc
```diff
@@ -325,28 +351,38 @@ static vector<ExplicitOperator> read_actions(
 }
 
 RootTask::RootTask(istream &in) {
-    read_and_verify_version(in);
-    bool use_metric = read_metric(in);
-    variables = read_variables(in);
+    utils::InputFileParser file(in);
```
FlorianPommerening, line 354: This doesn't have to be a file. How about `parser`?

src/search/tasks/root_task.cc
```diff
@@ -325,28 +351,38 @@ static vector<ExplicitOperator> read_actions(
...
 
-    mutexes = read_mutexes(in, variables);
+    mutexes = read_mutexes(file, variables);
+    for (size_t i = 0; i < mutexes.size(); ++i) {
+        for (size_t j = 0; j < mutexes[i].size(); ++j) {
+            check_facts(mutexes[i][j], variables);
+        }
+    }
 
+    // TODO: Maybe we could move this into a separate function as well
```
FlorianPommerening, line 367: sounds good

src/search/tasks/root_task.cc
```diff
@@ -325,28 +351,38 @@ static vector<ExplicitOperator> read_actions(
...
 
-    goals = read_goal(in);
+    goals = read_goal(file);
     check_facts(goals, variables);
-    operators = read_actions(in, false, use_metric, variables);
-    axioms = read_actions(in, true, use_metric, variables);
+    operators = read_actions(file, false, use_metric, variables);
+    axioms = read_actions(file, true, use_metric, variables);
+    file.confirm_end_of_file();
     /* TODO: We should be stricter here and verify that we
```
FlorianPommerening, line 386: The TODO is fixed now.

FlorianPommerening (commented):
I had a look at the code. I don't know the list of things this should check so I didn't verify completeness.

SimonDold (commented):
LGTM, just added renaming suggestions but i am happy as is.

src/search/utils/task_parser.h
```diff
@@ -1,28 +1,33 @@
...
 #include <regex>
 
 namespace utils {
-class InputFileLine;
-class InputFileLineParser;
-class InputFileToken;
 
-class InputFileParser {
+/*
+  Parse a task file.
```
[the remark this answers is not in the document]
FlorianPommerening: Looking at the interface of this class, it doesn't really parse a task (neither a file nor an encoding). It mostly does the scanning/lexing, i.e., breaking the input stream into tokens. The actual parsing then happens in `root_task.cc` in methods like `read_goal`, `read_actions` and so on. Maybe we should either move the parsing functionality here, or rename this class and file.

src/search/utils/input_file_parser.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include <memory>
+#include <regex>
+
+namespace utils {
+class InputFileLine;
+class InputFileLineParser;
+class InputFileToken;
+
+class InputFileParser {
```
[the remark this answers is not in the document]
tanjaschindler: We decided to keep two independent implementations.

src/search/tasks/root_task.cc
salome-eriksson: I was wondering the same thing here.

src/search/utils/task_parser.cc
```diff
@@ -114,15 +115,15 @@ void InputFileParser::confirm_end_of_line() {
...
 }
 
-void InputFileParser::confirm_end_of_file() {
-    string next_line = find_next_line(false);
-    if(next_line != "") {
-        error("Expected end of file, found non-empty line " + next_line);
+void TaskParser::confirm_end_of_input() {
+    find_next_line(false);
+    if(line != "") {
+        error("Expected end of task, found non-empty line " + line);
```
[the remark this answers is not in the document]
tanjaschindler: We keep the name.

src/search/utils/task_parser.h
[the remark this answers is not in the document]
tanjaschindler: obsolete due to major changes in the code

src/search/tasks/root_task.cc
tanjaschindler: We add a TODO to the code.

src/search/tasks/root_task.cc
```diff
@@ -108,155 +121,392 @@ class RootTask : public AbstractTask {
...
     }
+    return number;
 }
 
-static void check_facts(const ExplicitOperator &action, const vector<ExplicitVariable> &variables) {
-    check_facts(action.preconditions, variables);
-    for (const ExplicitEffect &eff : action.effects) {
-        check_fact(eff.fact, variables);
-        check_facts(eff.conditions, variables);
+void TaskParser::check_nat(const string &value_name, int value) {
```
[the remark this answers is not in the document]
tanjaschindler: We add a TODO for this to the code.

src/search/tasks/root_task.cc
```diff
@@ -494,7 +810,9 @@ void RootTask::convert_ancestor_state_values(
 
 void read_root_task(istream &in) {
     assert(!g_root_task);
-    g_root_task = make_shared<RootTask>(in);
+    utils::TaskLexer lexer(in);
+    TaskParser parser(move(lexer));
```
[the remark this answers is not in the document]
tanjaschindler: We add a TODO to the code.

src/search/tasks/root_task.cc
```diff
@@ -108,155 +121,392 @@ class RootTask : public AbstractTask {
...
+        context.error("Expected magic line '" + magic + "', got '" + line + "'.");
     }
 }
 
-static vector<FactPair> read_facts(istream &in) {
-    int count;
-    in >> count;
+vector<FactPair> TaskParser::read_facts(
+    bool read_from_single_line, const vector<ExplicitVariable> &variables) {
+    utils::TraceBlock block(context, "parsing conditions");
```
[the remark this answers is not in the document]
tanjaschindler: We add this as a TODO to the code.

PR #232, opened 2024-09-16
[trivial] Update pdb synopsis.

SimonDold:


src/search/pdbs/pdb_heuristic.cc
```diff
@@ -33,13 +33,27 @@ int PDBHeuristic::compute_heuristic(const State &ancestor_state) {
     return h;
 }
 
+static basic_string<char> paper_references() {
+    return utils::format_conference_reference(
+        {"Stefan Edelkamp"},
+        "Planning with Pattern Databases",
+        "https://ecp01.icaps-conference.org/papers/A-longpapers/2/edelkamp.pdf",
+        "Proceedings of the Sixth European Conference on Planning (ECP 2001)",
+        "84--90",
```
maltehelmert, line 43: The tools we use don't use LaTeX syntax, so we use a single dash for page ranges, not two dashes.
The linked version of the paper is not the one described by the reference. (The linked version has 12 pages, not 7; its page numbers would be 13-24.) I think the linked version is the pre-proceedings version (in Springer style) from 2001, not the actual "published" version produced in AAAI style much later. Perhaps that's a good compromise if we can't link the actual published paper, but it is confusing.

src/search/pdbs/pdb_heuristic.cc
```diff
@@ -33,13 +33,27 @@ int PDBHeuristic::compute_heuristic(const State &ancestor_state) {
...
 public:
     PDBHeuristicFeature() : TypedFeature("pdb") {
         document_subcategory("heuristics_pdb");
         document_title("Pattern database heuristic");
-        document_synopsis("TODO");
+        document_synopsis(
+            "Computes dictionaries for state-to-goal distances in"
+            "state space abstractions based on projections."
+            "First used in domain idependent planning by:"
+            + paper_references());
```
maltehelmert, line 57: There are spaces missing, so I think this will generated "instate" and "projections.First".
Typo "idependent", and I think "domain-independent" should be written with a hyphen, not as two words.

I don't think people describe PDBs as "dictionaries" (in particular because they don't store keys, only values). I think it's not necessary to get into such details in the first place, this could just be changed to "Computes goal distance in state space abstractions based on projections" etc.).

The particular PDB implementation in Fast Downward has its own paper (by Ortlieb et al.), which I think is worth referencing. It is already referenced from another PDB-related plug-in, see https://www.fast-downward.org/Doc/PatternCollectionGenerator#Hill_climbing

maltehelmert: I mean the paragraph "For implementation notes, see:" followed by the Ortlieb et al. reference. Most of that paper is actually about the basic PDB implementation, not the hill-climbing pattern collection paper. (But the paper discusses that too, so it's not wrong that the paper is also referenced from that plugin.)

PR #233, opened 2024-09-17
Update axiom documentation

salome-eriksson:
Update documentation for axiom support and safety properties for heuristics that have the new axioms option.
Furthermore, removed the axioms option from landmark_cost_partitioning_heuristic since it does not support axioms.

### DIFF
--- a/src/search/landmarks/landmark_sum_heuristic.h
+++ b/src/search/landmarks/landmark_sum_heuristic.h
@@ -17,23 +17,14 @@ class LandmarkSumHeuristic : public LandmarkHeuristic {
     int get_heuristic_value(const State &ancestor_state) override;
 public:
     LandmarkSumHeuristic(
-        const std::shared_ptr<LandmarkFactory> &lm_factory,
-        tasks::AxiomHandlingType axioms, bool pref,
+        const std::shared_ptr<LandmarkFactory> &lm_factory, bool pref,
         bool prog_goal, bool prog_gn, bool prog_r,
         const std::shared_ptr<AbstractTask> &transform,
         bool cache_estimates, const std::string &description,
-        utils::Verbosity verbosity);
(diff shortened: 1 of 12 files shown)


src/search/landmarks/landmark_sum_heuristic.h
```diff
@@ -17,9 +17,9 @@ class LandmarkSumHeuristic : public LandmarkHeuristic {
     int get_heuristic_value(const State &ancestor_state) override;
 public:
     LandmarkSumHeuristic(
+        tasks::AxiomHandlingType axioms,
```
salome-eriksson, line 20: I'm not happy about the order here because landmark heuristics always had the factory as first option which has no default value. Now we have axioms as first option here with a default value, meaning we cannot write `landmark_sum(lm_rhw())` anymore and instead need to write `landmark_sum(lm_factory=lm_rhw())`

SimonDold: I agree that `landmark_sum(lm_rhw())` is more user friendly than `landmark_sum(lm_factory=lm_rhw())`.

One issue is the parameter test that checks if the parameters in the cc constructor match the parameters in the documentation of the feature. 

If we do `tasks::add_axioms_option_to_feature(*this);
        add_landmark_heuristic_options_to_feature(
            *this, "landmark_sum_heuristic");` then the test expects the AxiomHandlingType first. If we do it the other way around it expects the AxiomHandlingType in the very end.
            
A third option would be to not call `add_landmark_heuristic_options_to_feature(
            *this, "landmark_sum_heuristic");`
            but to implement `add_landmark_sum_heuristic_options_to_feature(
            *this);`
            That does adds the options (that differ from Heuristic) one by one, as `add_landmark_heuristic_options_to_feature` does but in the order we prefer and with a call to `tasks::add_axioms_option_to_feature` at the right moment.

SimonDold:
Looks got to me

PR #239, opened 2025-01-13
Bitsets / Dynamic Bitset enhancements

guicho271828:
ported from the private codebase.


salome-eriksson:
Thanks for the PR! We would like to pursue it. As a next step, we will do a detailed code review; while I can't guarantee a specific date we will try to do this in the near future.

PR #240, opened 2025-01-14
More quality-of-life improvements: Exp/Eval/Gen limits, and soft limits

guicho271828:
This commit adds separate limits for expansion / evaluaiton / generations, and soft limits.

Soft limit addresses the limitation of hard limits (e.g. max_time) that one limit interferes another. For example, when max_eval = 100k and max_time = 5 sec, majority of runs are cut off before reaching 100k evaluations, and the result is not an appropriate representation of running 100k evaluations. To measure the evaluations and time separately, therefore one would run two separate experiments, which is a waste of compute for instances solved by both. Instead of running them separately, this option allows one to run the union of the two configurations which can be later filtered to produce distinct tables/figures. 

(Tbh I was surprized that FD does not have this feature yet.)


guicho271828:
If you are eager to add a class for representing the limits, I can do that too.

salome-eriksson:
Thanks for the PR! While we're in general open to the idea of adding more limits, we would prefer to not add them as simple parameters to each search algorithm since it blows up the constructor and adds redundancy (also for example in the wiki documentation. Instead we try to move the code in a more modular direction where things like this are done with wrappers: To implement limits, one would define a new search algorithm that takes an existing search algorithm as argument and adds the check for limits around it. We for example do something like this already for landmark graph generation; where we outsourced several post processing steps that previously were options for each of the generation methods to new wrapper landmark graph factories.
Does this direction sound good to you? We quite possibly would need to make some changes to the search algorithm code first to easily allow for such wrappers, but we would be happy to work towards that direction.

PR #241, opened 2025-01-15
Kuroiwa's Biased Exploration code base ported to modern fast downward

guicho271828:
This PR adds three OPEN list implementations (Kuroiwa and Beck, ICAPS2022, https://github.com/Kurorororo/biased-exploration) ported for the current fast downward. I am not the original author of the source code. I reproduced the results for Softmin (the most performant) and it is quite impressive.

* `softmin` : Open list that randomly selects h with a probability that is a softmax over different h values. Works with a single heuristic function.
* `softmin_heap` : same, but uses a heap.
* `softmin_type_based` :  Similar to `type_based`, it works with multiple evaluators. Instead of selecting buckets uniform randomly, uses the softmin criteria.
* `linear_weighted`, `linear_weighted_heap`, `linear_weighted_type_based` : Open list that randomly selects h with a probability that is linearly interpolated between max/min h, and its variants.
* `nth`, `nth_type_based` : Open list that randomly selects from the n th best element.

I am open to adding `softmin` and `softmin_type_based` variants only, removing the less-performant variants. `softmin_heap` may be useful depending on the action cost. 

Another opportunity might be to remove existing type-based open list, given that `softmin_type_based` is better theoretically & empirically.


salome-eriksson:
Thanks for the PR! Unfortunately we need to defer a decision here for a while, since we don't yet have a clear strategy on how to handle the integration of a larger code basis. For example there are open questions such as "Who owns the code?" and "Who maintains the code?". We are currently planning a meeting to discuss this and will get back to you once we have a clearer picture.

PR #242, opened 2025-01-24
Issue1165

ClemensBuechner:


src/search/cartesian_abstractions/utils_landmarks.cc
```diff
@@ -38,10 +38,23 @@ vector<FactPair> get_fact_landmarks(const LandmarkGraph &graph) {
     return facts;
 }
 
-VarToValues get_prev_landmarks(const LandmarkGraph &graph, const FactPair &fact) {
+utils::HashMap<FactPair, LandmarkNode *> get_facts_to_landmarks_map(
```
salome-eriksson, line 41: I would rename this to `get_fact_to_landmark_map` to emphasize that the map takes a single fact and maps it to a single landmark.

src/search/cartesian_abstractions/utils_landmarks.h
```diff
@@ -20,14 +23,17 @@ extern std::shared_ptr<landmarks::LandmarkGraph> get_landmark_graph(
...
+extern utils::HashMap<FactPair, landmarks::LandmarkNode *> get_facts_to_landmarks_map(
+    const std::shared_ptr<landmarks::LandmarkGraph> &graph);
+
 /*
   Do a breadth-first search through the landmark graph ignoring
-  duplicates. Start at the node for the given fact and collect for each
-  variable the facts that have to be made true before the given fact
-  can be true for the first time.
+  duplicates. Start at the given node and collect for each variable the
+  facts that have to be made true before the given fact can be true for
```
salome-eriksson, line 32: the given node?

src/search/landmarks/landmark_heuristic.h
```diff
@@ -29,6 +29,7 @@ class LandmarkHeuristic : public Heuristic {
 protected:
     std::shared_ptr<LandmarkGraph> lm_graph;
     const bool use_preferred_operators;
+    utils::HashMap<FactPair, std::unordered_set<int>> landmarks_achieved_by_fact;
```
salome-eriksson, line 32: Why don't we need to include utils here to get the HashMap? (In `cartesian_abstractions/utils_landmarks.h an include was added)

salome-eriksson: Also, I would add a comment here mentioning that this variable is only filled if preferred operators are computed. Just to avoid that in a couple of years someone needs a similar functionality, sees this variable and thinks that it always contains the information. Or alternatively change the name to make it clear that this has to do with preferred operators, but I don't have a good idea for such a name.

src/search/landmarks/landmark_heuristic.cc
```diff
@@ -119,29 +120,57 @@ void LandmarkHeuristic::compute_landmark_graph(
     }
 }
 
+void LandmarkHeuristic::precompute_for_preferred_operators() {
```
salome-eriksson, line 123: Maybe rename to `compute_landmarks_achieved_by_fact` or similar?

src/search/landmarks/landmark_heuristic.h
ClemensBuechner: > Why don't we need to include utils here to get the HashMap? (In `cartesian_abstractions/utils_landmarks.h an include was added)

The file we we actually need to include is `../utils/hash.h`. In `cartesian_abstractions/utils_landmarks.h`, this comes through including `utils.h` which is just a not so nice way to do this. I'll change that right away. In `landmarks/landmark_heuristic.h`, it comes through including `../heuristic.h` which in turn includes `operator_id.h` which includes `utils/hash.h`. I'm not sure what the best-practice C++ way to do this is. Should I also explicitly include `../utils/hash.h` in `landmarks/landmark_heuristic.h`?

FlorianPommerening: As far as I know, there are different philosophies about what to include where. One approach is "include what you use" (there is a tool of the same name) where each file has all the includes needed for the things that are used in that file, even if they are also brought in indirectly (usually with exceptions for some low level system libraries). We tried the tool once but I think gave up on it because it generated a lot of what we considered false positives. For example, we wouldn't necessarily include headers that are only needed by the base class of a derived class because we know that the base class must be known. 
Outdated wiki page about the tool: https://www.fast-downward.org/ForDevelopers/Includes

ClemensBuechner: Thanks for the pointer. Do we follow this policy despite not using the tool? I know without automatic testing the code probably has many places where this is violated, but now that I touch it anyway, I could at least try to do it right here.

FlorianPommerening: I don't think we currently follow any specific standard about this consistently or have a particular one in mind that we work towards. Generally, if something is indirectly included, I'd say this is fine if the indirect include is for something where we know the indirect include has to be there (e.g., a type used in a base class). It is not fine and should be made an explicit include in cases where the indirect include is "accidental", so for example if the reason it is included along the indirect path is unrelated to the reason it should be included here.

For example, I'd include `<vector>` if there is a local variable that is a vector, I'd not necessarily include it, if the only variable of type vector is a method parameter from a method defined in a base class, because since we have to know the base class, we have to have an include of the base class header and in that header `<vector>` has to be included.

ClemensBuechner: That makes sense, thanks for the elaboration. I have added another include just now.  I believe we now follow this policy regarding the `HashMap` within this changeset.

PR #245, opened 2025-02-03
Issue1081

SimonDold:
https://issues.fast-downward.org/issue1081


src/search/utils/component_errors.h
```diff
@@ -0,0 +1,28 @@
+#ifndef UTILS_COMPONENT_ERRORS_H
+#define UTILS_COMPONENT_ERRORS_H
+
+#include "exceptions.h"
```
jendrikseipp, line 4: Is this include needed?

src/search/evaluators/combining_evaluator.cc
```diff
@@ -4,6 +4,7 @@
 #include "../evaluation_result.h"
 
 #include "../plugins/plugin.h"
+#include "../utils/component_errors.h"
```
jendrikseipp, line 7: This include is present in some files that use verification functions and missing from others. I suggest removing it from all files that include `plugin.h`, because `plugin.h` includes `component_errors.h` anyway.

src/search/utils/component_errors.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include <vector>
+
+namespace utils {
+class ComponentArgumentError : public Exception {
```
jendrikseipp, line 9: The name ComponentArgumentError sounds like this is inherently tied to plugins. But the functionality can now be used by any class or function. So I think it makes sense to give it a more generic name. `ValueError` comes to mind since it's the equivalent from Python.

If you only want "plugin classes" to use the functions below, the ComponentArgumentError should live in plugin.h, I think.

jendrikseipp (changes requested):

src/search/merge_and_shrink/merge_and_shrink_algorithm.cc
```diff
@@ -55,8 +56,65 @@ MergeAndShrinkAlgorithm::MergeAndShrinkAlgorithm(
...
+                << "correcting threshold." << endl;
+        }
+    }
+
+    // TODO are these asserts needed?
     assert(max_states_before_merge > 0);
     assert(max_states >= max_states_before_merge);
+    // TODO why is this assert comparing threshold to max_states_before_merge
+    // while the default handling checks the comparison of threshold to max_states
     assert(shrink_threshold_before_merge <= max_states_before_merge);
```
FlorianPommerening, line 118: It looks like this could be violated.

src/search/merge_and_shrink/merge_and_shrink_algorithm.cc
```diff
@@ -55,8 +56,65 @@ MergeAndShrinkAlgorithm::MergeAndShrinkAlgorithm(
...
+        shrink_threshold_before_merge = max_states;
+        if (log.is_warning()) {
+            log << "WARNING: "
+                << "threshold exceeds max_states, "
+                << "correcting threshold." << endl;
+        }
+    }
+
+    // TODO are these asserts needed?
     assert(max_states_before_merge > 0);
```
FlorianPommerening, line 114: It is strange to write it as `> 0` here and `>= 1` in the test above. For the same reason, I would also write teh assumptions in lines 114, 115, and 118 in the same direction, rather than using `a < b` once and `c > b` another time.

Also, the assert just after the test seems redundant. Maybe the asserts should be in the constructor to show readers at one glance what properties we guarantee after construction.

src/search/merge_and_shrink/merge_and_shrink_algorithm.h
```diff
@@ -32,12 +32,12 @@ class MergeAndShrinkAlgorithm {
...
     // Hard limit: the maximum size of a transition system before being merged.
-    const int max_states_before_merge;
+    int max_states_before_merge;
     /* A soft limit for triggering shrinking even if the hard limits
        max_states and max_states_before_merge are not violated. */
-    const int shrink_threshold_before_merge;
+    int shrink_threshold_before_merge;
 
     // Options for pruning
     const bool prune_unreachable_states;
```
FlorianPommerening, line 43: Since you are removing the `const`, I'd remove this one as well.

src/search/plugins/plugin.h
```diff
@@ -111,7 +112,12 @@ class TypedFeature : public FeatureAuto<Constructed> {
     }
 
     Any construct(const Options &options, const utils::Context &context) const override {
-        std::shared_ptr<Base> ptr = this->create_component(options, context);
+        std::shared_ptr<Base> ptr;
+        try {
+            ptr = this->create_component(options, context);
```
FlorianPommerening, line 117: (Why) do we need `this` here?

src/search/utils/component_errors.h
FlorianPommerening: Looks like it is used in line 9 for the base class.

src/search/utils/component_errors.h
FlorianPommerening: It should only be used by "components" and we want to treat our plugin code as optional for components. For example, the Python interface could create components without using plugins. "Plugins" are a feature related to the command line parsing, e.g., where you transform a string like `astar(lmcut())` into components. In one of the draft PRs dealing with the component interaction problem, we introduce a base class for all components. Once we have that class, I think it would make sense for `ComponentArgumentError` to live next to it.

SimonDold: Then Jendriks review-comment to `combining_evaluator.cc` https://github.com/aibasel/downward/pull/245#discussion_r1950664272
would be handled as the include of the components father class would subsume the current include of `components_errors.h`.

FlorianPommerening: Yes, once we have the parent class, an implicit include would be fine. In the current code, I'd leave the include explicit, though.

src/search/plugins/plugin.h
SimonDold: It is needed 
https://isocpp.org/wiki/faq/templates#nondependent-name-lookup-members

FlorianPommerening: I like that the description starts with "This might hurt your head; better if you sit down."

src/search/merge_and_shrink/merge_and_shrink_algorithm.h
SimonDold: Is it just a concern of consistency? Isn't it better to have values `const` if possible?

The value `shrink_threshold_before_merge` is changed inside the constructor and therefore non-const.
However, `prune_unreachable_states` is never changed and therefore const.

src/search/utils/component_errors.h
SimonDold: @jendrikseipp About the `ValueError` suggestion: I think we do not want to have it this generic. At the moment we only use it for components (as Florian said) and there only in the constructor.
I don't think it would be good to catch a `ValueError` when trying to create a component as it could also be a value error in an internal function call (assuming they were used at different places too) that is not caused by the user putting weird argument values from the CLI. Such errors should either crash and burn or be caught at places that expect them.
If/When we decide to use `ValueError`s the `ComponentArgumentError` should probably inherit from them.

src/search/merge_and_shrink/label_reduction.cc
```diff
@@ -34,6 +35,10 @@ LabelReduction::LabelReduction(
       lr_method(method),
       lr_system_order(system_order),
       rng(utils::get_rng(random_seed)) {
+    utils::verify_comparison(
```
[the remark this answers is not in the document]
SimonDold: I like the suggestion and implemented it.

src/search/merge_and_shrink/merge_and_shrink_algorithm.cc
SimonDold: Yes, with this call you can trigger the violation: `"astar(merge_and_shrink(shrink_strategy=shrink_bisimulation(greedy=false),merge_strategy=merge_sccs(order_of_sccs=topological,merge_selector=score_based_filtering(scoring_functions=[goal_relevance(),dfp(),total_order()])),label_reduction=exact(before_shrinking=true,before_merging=false),max_states=3k,threshold_before_merge=2k,max_states_before_merge=1k))"`

It is unclear to me why the assertion is there...

src/search/merge_and_shrink/merge_and_shrink_algorithm.cc
```diff
@@ -55,9 +56,61 @@ MergeAndShrinkAlgorithm::MergeAndShrinkAlgorithm(
       log(utils::get_log_for_verbosity(verbosity)),
       main_loop_max_time(main_loop_max_time),
       starting_peak_memory(0) {
-    assert(max_states_before_merge > 0);
-    assert(max_states >= max_states_before_merge);
-    assert(shrink_threshold_before_merge <= max_states_before_merge);
+    handle_shrink_limit_defaults();
+
+    assert(this->max_states_before_merge > 0);
```
FlorianPommerening, line 61: Do we need `this` here? It looks like a different setting than before because there are no templates involved here.

SimonDold: Without `this` it would mean the parameter (which is not changed by `handle_shrink_limit_defaults`), with `this` it means the field (which happens to have the same name).

FlorianPommerening: I see. This is a bit unintuitive. Maybe add a short comment explaining this.

src/search/utils/component_errors.h
maltehelmert: I agree that `ComponentArgumentError` is very specific, and that this is our intention, so this is as it should be, I would say. This is a very specific barrier in the code (where the "user interface", whether it's CLI or library borders the internal code) and that deserves its own class. If and when we have a richer external API than the builders/constructors, perhaps something more generic makes sense. But it's usually not a good idea to try to define a general abstraction without knowing more about the contexts in which it will be used.

Regarding deriving from `ValueError` or not in the Python interface, the purpose of such inheritance is to be able to catch related things together, so `except ValueError` would also catch `ComponentArgumentError`. Whether to do this or not is more a question of Python best practices and not something we need to decide today.

src/search/merge_and_shrink/merge_and_shrink_algorithm.cc
SimonDold: We could also have a `_` prefix for the parameters.

FlorianPommerening: We decided against this, the last time we discussed this issue. I'd go with a comment even if we changed the name.

SimonDold: I added a comment

PR #247, opened 2025-02-05
[issue220] Improve PDDL error detection and handling

grucla:


grucla:
Salomé, Malte, Travis, and I discussed the points that Salomé listed above and came to the following conclusions:

1. When something expects a string, this string should not start with a number, i. e., especially numbers should not be treated as strings.
2. The changed meaning of error code TRANSLATE_INPUT_ERROR was an aftermath of merging pull request #155 but is not what we want. We would like to introduce a new error code for input files that are not valid PDDL. But we might not address it in this pull request because the error codes could use some more cleanup which can be bundled in a separate issue.
3. We did not reach a final decision because we did not know the "official" PDDL semantics regarding this. Fox and Long 2003 (which is a rather canonical PDDL specification) allows empty conditions. I quickly tried to find a definition in the PDDL book (Haslum et al. 2019) but only found a rather informal one for STRIPS that allows single facts or conjunctions of facts as conditions (I do not know if this includes empty conjunctions, or not).
4. We want to return to an old treatment where the parser prints a warning if a predicate is defined in the predicates-block that has the same name as a type (because some benchmarks allow this although not all people would agree to allowing this), and where the parser accepts using a type as a unary predicate.
5. We are still not completely clear on this point but think that it is already addressed. Furthermore, the word "string" should be avoided in error messages for the user, a better word might be "word" as it cannot be confused with the string-type from Python or C++.
6. We assume that "numeric" in the summary of [issue220](https://issues.fast-downward.org/issue220) should be "number" as mentioned in msg1249 of the same issue and want it to be reserved as a type. To implement this we could just disallow defining a new type called "number", i. e. check for this word when going through the types-block of a domain file.

If I forgot something from our meeting, feel free to correct me.

src/translate/pddl_parser/warning.py
```diff
@@ -0,0 +1,10 @@
+import sys
+
+
+PRINTED_WARNINGS = set()
```
maltehelmert, line 4: I wouldn't make this all-caps any more. I think this is now a more "normal" variable than it used to be.

src/translate/pddl_parser/warning.py
```diff
@@ -0,0 +1,10 @@
...
+
+def print_warning(msg):
+    global PRINTED_WARNINGS
```
maltehelmert, line 7: We don't need global any more. The access here is not different from how we access other names from the global namespace, such as "sys" and "print". The old code needed "global" because it needed to rebind the global variable (use "="), but now we just call methods.

salome-eriksson: Ah I didn't realize that assignment in particular is the problem. I changed it (and also changed the all-caps to no caps)

PR #250, opened 2025-02-11
Issue1074

ClemensBuechner:


src/search/landmarks/exploration.cc
```diff
@@ -83,8 +83,8 @@ void Exploration::build_unary_operators(const OperatorProxy &op) {
             precondition.push_back(&propositions[precondition_fact.var]
                                    [precondition_fact.value]);
 
-        FactProxy effect_fact = effect.get_fact();
-        Proposition *effect_proposition = &propositions[effect_fact.get_variable().get_id()][effect_fact.get_value()];
+        const auto &[var, value] = effect.get_fact().get_pair();
```
salome-eriksson, line 86: I find it weird to use const references here, since we usually just copy ints; but I'm also not familiar with the new syntax. Is there a specific reason? (The comment applies to several other places as well.)

src/search/landmarks/exploration.cc
```diff
@@ -120,11 +120,19 @@ void Exploration::setup_exploration_queue(
...
     for (FactProxy fact : state) {
         Proposition *init_prop =
             &propositions[fact.get_variable().get_id()][fact.get_value()];
-        enqueue_if_necessary(init_prop);
+        /*
+          The following condition is a workaround to avoid compiling the
+          initial state away in a task transformation in case of unary
+          relaxation. Otherwise, every initial proposition is a landmark
+          and should be enqueued.
+        */
```
salome-eriksson, line 132: While I understand the comment at the moment, I'm not sure we will still do so after a while. In particular, we might not remember the connection to compiling the initial state away.
Suggestion:
In the unary relaxation, atoms true in the current state are only landmarks if they are required in some operator along the (relaxed) solution. The conceptual way to simulate this would be to compile the initial state away and introduce unary operators that achieve each atom. We can however also just not enqueue excluded atoms that are currently true; that way they do not become reached and can thus not be used by any operator.

src/search/landmarks/exploration.cc
```diff
@@ -140,15 +148,16 @@ void Exploration::setup_exploration_queue(
...
-                                      excluded_op_ids.end());
-    for (OperatorProxy op : task_proxy.get_operators()) {
-        for (EffectProxy effect : op.get_effects()) {
-            if (effect.get_conditions().empty()
-                && propositions[effect.get_fact().get_variable().get_id()]
-                [effect.get_fact().get_value()].excluded) {
-                op_ids_to_mark.insert(op.get_id());
-                break;
+    unordered_set<int> op_ids_to_mark;
+    if (!use_unary_relaxation && !excluded_props.empty()) {
```
salome-eriksson, line 152: Do we still need the check for !excluded_props.empty()? The advantage of checking it is that we don't do the loop if we have no excluded propositions; but does this happen in practice now that we only have excluded props (i.e. no excluded ops)?

src/search/landmarks/landmark_factory_rpg_exhaust.cc
```diff
@@ -35,26 +35,23 @@ void LandmarkFactoryRpgExhaust::generate_relaxed_landmarks(
...
     }
-
-    if (only_causal_landmarks) {
-        discard_noncausal_landmarks(task_proxy, exploration);
-    }
 }
 
 bool LandmarkFactoryRpgExhaust::supports_conditional_effects() const {
+    /* TODO: I would say it does support conditional effects if unary relaxation
+        is used, but not sure. */
```
salome-eriksson, line 54: I'm not sure enough either to justify changing this :).

src/search/landmarks/exploration.cc
salome-eriksson: I'm actually wondering if the behaviour for the "normal" exploration was correct before. If we exclude an atom that is currently true, should nothing be reachable then? Because we cannot reach anything without reaching the excluded atom.
At least for exhaust it did not matter before because we did not call an exploration for atoms that are currently true, but now we do. I assume exhaust with use_unary_relaxation=false would currently not recognize any initial atom as landmark, correct?

ClemensBuechner: Indeed, the comment may be more confusing than helpful, and also it is not correct. This is because in the unary relaxation, every atom that holds in the initial state is also a landmark. The reason why we wanted to compile the initial state away is to achieve that the unary relaxation landmarks are equivalent to the causal landmarks.

I've talked about this with Malte offline. We discussed that what we would like is that `lm_exhaust` with unary relaxation behaves the same as `lm_hm` with `m=1`. This indeed requires not enqueuing propositions that hold initially here, so I think the implementation can stay but the comment should be changed. I suggest to simply remove it. In line 123 just slightly above, it already says that we only mark non-excluded propositions as reached, I think that's enough information and we don't need to confuse people with the fact that this has some relation to compiling the initial state away.

We have also talked about how the code should behave if the option `use_unary_relaxation` is set to `false`. We think it would be best if it behaves as similarly as possible to the case where it is set to `true`. So I'm going to remove the condition on `use_unary_relaxation` in the `if` condition. This means that the new implementation no longer does exactly the same thing as before working on the issue. In particular, the new version will find fewer landmarks with the `lm_exhaust` factory but we don't think this is a problem because they all hold in the initial state anyway and would be set to `past` from the start in the initial progression step. Since `lm_exhaust` also does not compute any orderings, it should make no difference except saving some memory.

src/search/landmarks/landmark_factory_rpg_exhaust.cc
ClemensBuechner: Although I'm proud of my noble intention above, I would prefer to make this ready to merge rather than answering this theoretical question. I've taken note of it on my todo list for after the sprint but will just remove the comment for now.

PR #252, opened 2025-02-13
Use views.

ClemensBuechner:
This is just to check if this change passes all tests in the Github actions.

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -21,7 +21,10 @@ jobs:
       matrix:
         version:
           - {ubuntu: 'ubuntu-22.04', python: '3.10', cc: gcc-12, cxx: g++-12, run_tox_tests: true}
-          - {ubuntu: 'ubuntu-22.04', python: '3.10', cc: clang-15, cxx: clang++-15, run_tox_tests: false}
+          # NOTE: Temporarily removing support for the second compiler of the
+          # older LTS Ubuntu because clang-15 does not support the `ranges`
+          # and `views` libraries.
+          # - {ubuntu: 'ubuntu-22.04', python: '3.10', cc: clang-15, cxx: clang++-15, run_tox_tests: false}
           - {ubuntu: 'ubuntu-24.04', python: '3.10', cc: gcc-14, cxx: g++-14, run_tox_tests: true}
           - {ubuntu: 'ubuntu-24.04', python: '3.10', cc: clang-18, cxx: clang++-18, run_tox_tests: false}
     env:
@@ -137,7 +140,6 @@ jobs:
           pip3 install tox
           sudo apt-get -y install zlib1g-dev libgmp3-dev gcc flex bison
 
-        # NOTE: VAL does not compile with clang-12.
       - name: Install VAL
         run: |
           git clone https://github.com/KCL-Planning/VAL.git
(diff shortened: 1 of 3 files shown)


FlorianPommerening:
I fixed the detection of cases in which we want to add the flag `--fexperimental-library`. Previously it was only added with clang 15.0.0, now it is added with clang 15.*.*. Unfortunately, the error remains. Google keeps bringing me back here: https://github.com/llvm/llvm-project/issues/52696 and there it sounds like  libstdc++ and clang each blaming each other.

maltehelmert:
I've also unsuccessfully tried to get this to work for an hour or so.
I think it would be fine to drop support for clang 15. People that want to use Ubuntu 22.04 can use gcc.

ClemensBuechner:
> I think it would be fine to drop support for clang 15. People that want to use Ubuntu 22.04 can use gcc.

Are there objections to this? If not, what are the required steps to do this? Simply remove the corresponding cases from the Github actions and adapt the documentation?

FlorianPommerening:
No objections from my side, but I would treat this as an exception, i.e., I would say that once we switch to supporting Ubuntu 24.04 and 26.04 and drop support for 22.04, we would then go back to the model of supporting the clang version available on those LTS versions.

One alternative to dropping clang 15 support would be to replace it with clang 16 support. We already have documentation on how to get it to run on Ubuntu 22.04 and use it for our style checks (https://www.fast-downward.org/ForDevelopers/DevelopmentSetup). I checked in my test container and it seems to build this pull request without problems.

Either way, we'd have to document the changes in these places:
* https://www.fast-downward.org/ForDevelopers/Requirements?highlight=%28clang%29
* https://www.fast-downward.org/ForDevelopers/CodingConventions?highlight=%28clang%29 (out of date anyway)
* https://github.com/aibasel/downward/blob/main/.github/workflows/ubuntu.yml

maltehelmert:
> No objections from my side, but I would treat this as an exception, i.e., I would say that once we switch to supporting Ubuntu 24.04 and 26.04 and drop support for 22.04, we would then go back to the model of supporting the clang version available on those LTS versions.
> 
> One alternative to dropping clang 15 support would be to replace it with clang 16 support. We already have documentation on how to get it to run on Ubuntu 22.04 and use it for our style checks (https://www.fast-downward.org/ForDevelopers/DevelopmentSetup). I checked in my test container and it seems to build this pull request without problems.
> 
> Either way, we'd have to document the changes in these places:
> 
>     * https://www.fast-downward.org/ForDevelopers/Requirements?highlight=%28clang%29
> 
>     * https://www.fast-downward.org/ForDevelopers/CodingConventions?highlight=%28clang%29 (out of date anyway)
> 
>     * https://github.com/aibasel/downward/blob/main/.github/workflows/ubuntu.yml

I think the demand for clang-16 support  will be modest and I wouldn't want to make the Github Actions more complex just for this reason. But if you want to add it to the tests, I have no objections.

One of the reasons why we decided in the past that we support both compilers on an old Ubuntu LTS was that it didn't really cost us anything because the lowest common denominator was determined by the Windows compilers for a long time anyway. If this support comes at a higher cost, I think it's perfectly fine to reconsider this. Supporting two compilers on the same OS is a luxury.

In any case, this is the kind of decision that I would want to bring up with everyone on the Discord server (internal channel). Do you want me to bring it up?

FlorianPommerening:
I'm also fine with dropping support now and reevaluating the situation once we change supported Ubuntu versions (in about a year when 26.04 is out). But we can discuss this on Discord. I people prefer keeping support for clang 16 on Ubuntu 22.04, I would be willing to update the Github actions. I don't think it would be too big of a change.

FlorianPommerening:
In case you plan to merge this, remember to undo the change to the CMake files before merging. They are not required now that we do not support clang 15.

ClemensBuechner:
> In case you plan to merge this, remember to undo the change to the CMake files before merging. They are not required now that we do not support clang 15.

Yes, thank you! I just wanted to make sure we're back on the green before doing anything else.

PR #253, opened 2025-02-15
Fix if/else logic for computing cache_misses in cg heuristic

pac48:
There seems to be a bug in the computation of `cache_misses` in the cg heuristic. It looks like it should only be increased if `use_the_cache` is true and `cached_val != CGCache::NOT_COMPUTED`.


roeger:
The proposed change looks good to me.

PR #255, opened 2025-03-06
[issue1174] Polish markdown documentation

roeger:
This is a collection of small fixes and improvements I made while checking that we moved all content from the old wiki.


roeger:
@FlorianPommerening had a look and agreed with the change. Merging it...

PR #259, opened 2025-04-01
[trivial] Add pull request template.

ClemensBuechner:


.github/pull_request_template.md
```diff
@@ -0,0 +1,33 @@
+In the `Add a title` field, write a meaningful short commit message in imperative mood. Start the title with [issueX] where X is the issue number on https://issues.fast-downward.org.
+
+---
+
+In this field (titeled `Add a description`), write a more extensive summary of the changes made in this pull request. This description will be suggested as the commit message when squashing and merging the pull request, so it should be refined along the development process. The final commit message should describe how the changes affect planner behavior:
```
FlorianPommerening, line 5:
```suggestion
In this field (titled `Add a description`), write a more extensive summary of the changes made in this pull request. This description will be suggested as the commit message when squashing and merging the pull request, so it should be refined along the development process. The final commit message should describe how the changes affect planner behavior:
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,33 @@
...
+
+Prior to this change, the search component was implemented in C++.
+We have changed the planner to run in C#, which is an industry
+standard language. Expect a performance increase of at least -37%
+in common cases. The planner is now able to solve large tasks
+in the Gripper and Spanner domains.
+```
+
+Within the issue branch, you are free to write commit messages however you like, since these commits will be squashed. We still recommend the following guidelines as it will make doing reviews and writing the final commit message easier.
+ - Prepend "[<branch>]" to all commits of the branch <branch>. You can copy misc/hooks/commit-msg to .git/hooks/commit-msg to enable a hook which automatically prepends the right string to each commit message.
```
FlorianPommerening, line 31: Do we still want this?

ClemensBuechner: Probably not. I don't do this anymore, so I wouldn't recommend doing it either. But then we should probably also change it on our Wiki because this part is really just copied from there.

PR #261, opened 2025-05-06
[issue220] Addendum: Update PDDL Support web page.

grucla:
We addressed multiple points related to invalid PDDL input files (issue220) in
an earlier change. This change updates the Fast Downward website to reflect
those changes. (No code changes.)

### DIFF
--- a/docs/pddl-support.md
+++ b/docs/pddl-support.md
@@ -52,19 +52,44 @@ supported beyond basic STRIPS, with some limitations mentioned below:
 These limitations are somewhat likely to be lifted in the future, but progress
 is slow.
 
+## Notes on syntax checks
+
+There exist multiple PDDL definitions and they differ in some aspects, this
+section describes how Fast Downward handles such cases.
+
+- **empty conditions:** Like Fox and Long ("PDDL2.1: An Extension to PDDL for
+  Expressing Temporal Planning Domains" by Maria Fox and Derek Long, JAIR
+  20:61-124, 2003), Fast Downward allows empty conditions (i. e., preconditions
+  and goals).
+- **keyword "either":** As mentioned above, Fast Downward does not support
+  `(either ...)` types, however it does not raise an error when `(either ...)`
+  is used in the `:predicates` section. For example, defining `(somepredicate ?x -
+  (either type1 type2))` is allowed (but ignored). Furthermore, type validity
+  is not checked for the `:predicates` section.
+- **empty list before type specification:** While most (if not all) PDDL
+  specifications require at least one token before a type specification (e. g.
+  `(someobject1 someobject2 - sometype)`), Fast Downward only prints a warning,
+  instead of raising an error when nothing is given before a type
+  specification (e. g. `(- sometype)`). The reason is that the woodworking
+  domain from IPC 2008 and 2011 contain problems where this oddity occurs.
+- **same name for predicate and type:** According to McDermott et al. ("PDDL -
+  The Planning Domain Definition Language - Version 1.2" by Drew McDermott,
+  Malik Ghallab, Adele Howe, Craig Knoblock, Ashwin Ram, Manuela Veloso, Daniel
+  Weld, and David Wilkins, Yale Center for Computational Vision and Control,
+  Technical Report CVC TR-98-003/DCS TR-1165, 1998) a type is essentially just
+  a unary predicate. Thus, defining a new predicate with the same name as a
+  type might lead to unexpected behaviour. Because of this, Fast Downward
+  prints a warning if a predicate (of any arity) uses the same name as a type.
+- **predefined type "number":** Fast Downward does not allow defining a type
+  called "number" because this keyword is reserved.
+
 ## Bugs
 
 Some features are supported in theory, but currently affected by bugs.
 In addition to the issues with conditional effects mentioned above which
-are somewhere between a bug and a missing feature, we are currently
-aware of the following bugs:
+are somewhere between a bug and a missing feature, we keep track of known bugs
+in our [issue tracker](http://issues.fast-downward.org/).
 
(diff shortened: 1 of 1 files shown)


docs/pddl-support.md
```diff
@@ -52,19 +52,44 @@ supported beyond basic STRIPS, with some limitations mentioned below:
...
 
+## Notes on syntax checks
+
+There exist multiple PDDL definitions and they differ in some aspects, this
+section describes how Fast Downward handles such cases.
+
+- **empty conditions:** Like Fox and Long ("PDDL2.1: An Extension to PDDL for
+  Expressing Temporal Planning Domains" by Maria Fox and Derek Long, JAIR
+  20:61-124, 2003), Fast Downward allows empty conditions (i. e., preconditions
+  and goals).
```
maltehelmert: I would clarify here what you mean by "empty conditions". There are at least three ways to say "no preconditions" that I've seen:

- no `:precondition` block at all
- empty conjunction: `:precondition (and)`
- list with no arguments: `:precondition ()`

I think we're talking about #3 here.

docs/pddl-support.md
```diff
@@ -52,19 +52,44 @@ supported beyond basic STRIPS, with some limitations mentioned below:
...
+- **keyword "either":** As mentioned above, Fast Downward does not support
+  `(either ...)` types, however it does not raise an error when `(either ...)`
+  is used in the `:predicates` section. For example, defining `(somepredicate ?x -
+  (either type1 type2))` is allowed (but ignored). Furthermore, type validity
+  is not checked for the `:predicates` section.
+- **empty list before type specification:** While most (if not all) PDDL
```
maltehelmert: What do you mean by "if not all" here?
"If not all" means "most of them do, and maybe all of them do, but we don't know this for sure". Do you mean "but not all" instead? Perhaps there parenthetical part can be dropped here entirely.

docs/pddl-support.md
```diff
@@ -52,19 +52,44 @@ supported beyond basic STRIPS, with some limitations mentioned below:
...
+  specifications require at least one token before a type specification (e. g.
+  `(someobject1 someobject2 - sometype)`), Fast Downward only prints a warning,
```
maltehelmert: Should have no comma at the end of the line.

docs/pddl-support.md
```diff
@@ -52,19 +52,44 @@ supported beyond basic STRIPS, with some limitations mentioned below:
...
+  instead of raising an error when nothing is given before a type
+  specification (e. g. `(- sometype)`). The reason is that the woodworking
+  domain from IPC 2008 and 2011 contain problems where this oddity occurs.
+- **same name for predicate and type:** According to McDermott et al. ("PDDL -
+  The Planning Domain Definition Language - Version 1.2" by Drew McDermott,
+  Malik Ghallab, Adele Howe, Craig Knoblock, Ashwin Ram, Manuela Veloso, Daniel
+  Weld, and David Wilkins, Yale Center for Computational Vision and Control,
+  Technical Report CVC TR-98-003/DCS TR-1165, 1998) a type is essentially just
+  a unary predicate. Thus, defining a new predicate with the same name as a
```
maltehelmert, line 80: Instead of saying "is essentially just", I would say "can be used as".

docs/pddl-support.md
```diff
@@ -52,19 +52,44 @@ supported beyond basic STRIPS, with some limitations mentioned below:
 These limitations are somewhat likely to be lifted in the future, but progress
 is slow.
 
+## Notes on syntax checks
+
+There exist multiple PDDL definitions and they differ in some aspects, this
+section describes how Fast Downward handles such cases.
```
maltehelmert, line 58: This is a run-on sentence. Should have a period instead of a comma.

docs/pddl-support.md
maltehelmert: Also "i.e." doesn't work here. Conditions are not just used in preconditions and goals. Other places where conditions occur are axiom bodies and effect conditions in conditional effects. The phrase "i.e." requires that you define what a condition is here rather than just give examples. Perhaps use "e.g.,". (BTW, for "i.e." and "e.g.", it's customary not to have a space in the middle.)

docs/pddl-support.md
```diff
@@ -52,19 +52,44 @@ supported beyond basic STRIPS, with some limitations mentioned below:
 These limitations are somewhat likely to be lifted in the future, but progress
 is slow.
 
+## Notes on syntax checks
```
maltehelmert: Syntax *checks* is very technical about implementation, I would rather use "syntax". But also I don't think all of these are about syntax. For example, using types as predicates is not really about syntax because types and predicates have the same syntax. From the section intro, this seems to be about ambiguities in PDDL, so perhaps something like "## PDDL ambiguities" or "## PDDL corner cases" or something like that.

docs/pddl-support.md
grucla: Like you guessed, we indeed mean the third point. (Though the first one is also allowed and I think the second one as well.) Also, I agree that it should be "e.g." instead of "i.e.". I'll change this and give an example for an empty precondition for clarification.

docs/pddl-support.md
grucla: We mean "most of them do, and maybe all of them do, but we don't know this for sure". I think without the part in parentheses people would wonder for which PDDL specifications this does not hold. How about changing "if not all" to "or maybe all"?

docs/pddl-support.md
grucla: I agree and will change it to "PDDL ambiguities".

docs/pddl-support.md
maltehelmert: Perhaps write "While all PDDL specifications we know"?

grucla: Sounds good to me, I just pushed the change.

PR #263, opened 2025-05-10
[issue1184] Adding the option to keep noop operators.

haz:
> New option to avoid filtering noop operators

The command line now has a translator option to `--keep-noop-operators`, which is used in several places to make sure that operators with no effects are _not_ pruned. Default is false.

This is akin to `--keep-unimportant-variables`, and is essential for some applications (e.g., FOND encodings, as per the discussion [here](https://github.com/AI-Planning/fond-utils/issues/16)).


src/translate/instantiate.py
```diff
@@ -53,7 +53,7 @@ def instantiate_goal(goal, init_facts, fluent_facts):
 
 # The input task must have been normalized
 # The model has been computed by build_model.compute_model
-def instantiate(task: pddl.Task, model: Any) -> Tuple[
+def instantiate(task: pddl.Task, model: Any, filter_noop_operators: bool) -> Tuple[
```
maltehelmert: At the moment it seems that the prevailing way we use the `options` module is by accessing the options directly where needed, as in importing `options` and using `options.filter_noop_operators` like a global variable. It's a questionable design choice regarding encapsulation, and there are one or two places where we handle things a bit differently, but I think at this point it's better to keep the code consistent in that style. So I would get rid of changes like this one that are about passing the option through. For example, in this file, this would mean undoing all changes.

src/translate/options.py
```diff
@@ -49,6 +49,10 @@ def parse_args():
         "--keep-unimportant-variables",
         dest="filter_unimportant_vars", action="store_false",
         help="keep variables that do not influence the goal in the causal graph")
+    argparser.add_argument(
+        "--keep-noop-operators",
+        dest="filter_noop_operators", action="store_false",
+        help="keep noop operators in the output")
```
maltehelmert: Despite the previous comment on consistency, I suggest we do not follow the precedent set by `filter_unimportant_vars` in two ways. Firstly, I think it's clearer if the option sets a flag to true rather than to false. Secondly, "filter" is a poor word because it can either mean "keep" (as in Python's "filter" function) or "remove" (as in filter out, like here). I also prefer "no-op" over "noop" in user-facing text, and for consistency would then perhaps also use "no_op" for variable names.

Suggestion: rename option to "keep-no-ops" (no-op operators feels a bit redundant), with destination "keep_no_ops" using "store_true", flipping the meaning of the switch. Help text changed to "keep operators without effects in the output".

src/translate/pddl/actions.py
```diff
@@ -71,7 +71,7 @@ def untyped(self):
         return result
 
     def instantiate(self, var_mapping, init_facts, init_assignments,
-                    fluent_facts, objects_by_type, metric):
+                    fluent_facts, objects_by_type, metric, filter_noop_operators):
```
maltehelmert: remove parameter

src/translate/pddl/actions.py
```diff
@@ -93,7 +93,7 @@ def instantiate(self, var_mapping, init_facts, init_assignments,
         for eff in self.effects:
             eff.instantiate(var_mapping, init_facts, fluent_facts,
                             objects_by_type, effects)
-        if effects:
+        if effects or not filter_noop_operators:
```
maltehelmert: `if effects or options.keep_no_ops` (need import options if not already there)

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -466,7 +466,7 @@ def parse_assignment(context, alist):
                           f" Use '=' or 'increase'.")
 
 
-def parse_action(context, alist, type_dict, predicate_dict, constant_names):
+def parse_action(context, alist, type_dict, predicate_dict, constant_names, filter_noop_operators):
```
maltehelmert: as above

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -515,7 +515,7 @@ def parse_action(context, alist, type_dict, predicate_dict, constant_names):
             context.error(f"Missing fields. Expecting {SYNTAX_ACTION}.")
         for _ in iterator:
             context.error(f"Too many fields. Expecting {SYNTAX_ACTION}")
-    if eff:
+    if eff or not filter_noop_operators:
```
maltehelmert: as above

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -541,7 +541,7 @@ def parse_axiom(context, alist, type_dict, predicate_dict, constant_names):
                           len(predicate.arguments), condition)
 
 
-def parse_axioms_and_actions(context, entries, type_dict, predicate_dict, constant_names):
+def parse_axioms_and_actions(context, entries, type_dict, predicate_dict, constant_names, filter_noop_operators):
```
maltehelmert: as above, also for remaining changes in this file

src/translate/pddl_parser/pddl_file.py
```diff
@@ -22,7 +22,7 @@ def parse_pddl_file(type, filename):
                          (type, filename, e))
 
 
-def open(domain_filename=None, task_filename=None):
+def open(domain_filename=None, task_filename=None, filter_noop_operators=True):
```
maltehelmert: undo changes in this file

src/translate/simplify.py
```diff
@@ -223,14 +223,14 @@ def register_variable(self, old_domain_size, init_value, new_domain):
             self.new_sizes.append(new_size)
             self.new_var_count += 1
 
-    def apply_to_task(self, task):
+    def apply_to_task(self, task, filter_noop_operators):
```
maltehelmert: undo changes in this file that pass the parameter through

src/translate/simplify.py
```diff
@@ -369,7 +369,7 @@ def translate_operator(self, op):
                 new_var = new_entry[0]
                 new_prevail_vars.discard(new_var)
 
-        if not new_pre_post:
+        if not new_pre_post and filter_noop_operators:
```
maltehelmert: ...`and not options.keep_no_ops`

src/translate/translate.py
```diff
@@ -332,7 +332,8 @@ def build_sas_operator(name, condition, effects_by_variable, cost, ranges,
             # precondition, so we remove it from the prevail condition
             condition.pop(var, -1)
     if not pre_post:  # operator is noop
-        return None
+        if options.filter_noop_operators:
+            return None
```
maltehelmert: Combine into a single `if` as in the other places: `if not pre_post and not options.keep_no_ops:`

src/translate/translate.py
```diff
@@ -544,7 +545,7 @@ def unsolvable_sas_task(msg):
 def pddl_to_sas(task):
     with timers.timing("Instantiating", block=True):
         (relaxed_reachable, atoms, actions, goal_list, axioms,
-         reachable_action_params) = instantiate.explore(task)
+         reachable_action_params) = instantiate.explore(task, options.filter_noop_operators)
```
maltehelmert: Get rid of changes in this file that just pass variables through.

src/translate/translate.py
```diff
@@ -614,7 +615,8 @@ def pddl_to_sas(task):
         with timers.timing("Reordering and filtering variables", block=True):
             variable_order.find_and_apply_variable_order(
                 sas_task, options.reorder_variables,
-                options.filter_unimportant_vars)
+                options.filter_unimportant_vars,
```
maltehelmert: I am aware this existing line is inconsistent with the recommendations I'm making here, but it's the exception rather than the rule in the current code. (I grepped for all uses of `options`.)

src/translate/variable_order.py
```diff
@@ -190,12 +190,12 @@ def __init__(self, ordering):
         self.ordering = ordering
         self.new_var = {v: i for i, v in enumerate(ordering)}
 
-    def apply_to_task(self, sas_task):
+    def apply_to_task(self, sas_task, filter_noop_operators):
```
maltehelmert: same in this file

src/translate/variable_order.py
```diff
@@ -243,7 +243,7 @@ def _apply_to_operators(self, operators):
                                     if var in self.new_var)
                     pre_post.append(
                         (self.new_var[eff_var], pre, post, new_cond))
-            if pre_post:
+            if pre_post or not filter_noop_operators:
```
maltehelmert: `or options.keep_no_ops`

maltehelmert (changes requested):
Also see issue tracker.

src/translate/options.py
```diff
@@ -49,6 +49,10 @@ def parse_args():
         "--keep-unimportant-variables",
         dest="filter_unimportant_vars", action="store_false",
         help="keep variables that do not influence the goal in the causal graph")
+    argparser.add_argument(
+        "--keep-no-ops",
+        dest="keep_no_ops", action="store_true",
+        help="keep operators without effects in the output")
```
maltehelmert, line 54: Minor polishing, but why not: I think we can now remove the "dest" argument because "keep_no_ops" is the default if the option is called "--keep-no-ops". It's a bit magic, but it looks like we exploit the default in the other cases where it works (e.g. "--add-implied-preconditions", "--dump-task".)

haz: Makes sense, all set!

PR #266, opened 2025-07-23
[issue1186] Refactor EagerSearch::step() into smaller helper functions

speckdavid:
This is a decomposition of the step() function of EagerSearch into multiple smaller subfunctions, which makes the function more accessible.


src/search/search_algorithms/eager_search.cc
```diff
@@ -168,48 +193,45 @@ SearchStatus EagerSearch::step() {
             }
         }
 
-        node->close();
-        assert(!node->is_dead_end());
+        node.close();
+        assert(!node.is_dead_end());
         update_f_value_statistics(eval_context);
         statistics.inc_expanded();
```
SimonDold: I would have expected the `statistics.inc_expanded()` call in the function named `expand`.

src/search/search_algorithms/eager_search.cc
```diff
@@ -168,48 +193,45 @@ SearchStatus EagerSearch::step() {
...
-        if ((node->get_real_g() + op.get_cost()) >= bound)
+        if ((node.get_real_g() + op.get_cost()) >= bound)
             continue;
 
-        State succ_state = state_registry.get_successor_state(s, op);
+        State succ_state = state_registry.get_successor_state(node.get_state(), op);
         statistics.inc_generated();
         bool is_preferred = preferred_operators.contains(op_id);
 
         SearchNode succ_node = search_space.get_node(succ_state);
```
SimonDold, line 234: I think it would be fitting to the story to have a function call like `SearchNode succ_node = generate_succ_node(node,op);`  

I think this would naturally change the position of the `bool is_preferred = preferred_operators.contais(op_id);` ... i do not see why this is checked here already(?) and not right before the "big if" in line 238.

src/search/search_algorithms/eager_search.cc
speckdavid: I think this will change the number of expanded nodes by 1. Currently, we increment the number of expanded nodes immediately after popping from the open list, and then perform the goal check. So, the goal check is considered part of the expansion.

One option would be to rename the expand function to something like `generate_successors`. Alternatively, we could change how we count expansions:  not counting nodes that are popped and immediately recognized as goal states without generating successors. Lastly, we could live with it as is. What do you think @SimonDold ?

src/search/search_algorithms/eager_search.cc
speckdavid: Moving down `bool is_preferred = preferred_operators.contais(op_id);` makes sense to me. I will do so.

speckdavid: Regarding function `generate_succ_node(const Node& node, const OperatorProxy& op)`:  I think this is a more general function that could be used for other search algorithms besides `EagerSearch`. Should we implement it as a function in `SearchAlgorithm`? However, I would like to avoid modifying `LazySearch` and `EHC` in this PR to use it directly.

src/search/search_algorithms/eager_search.cc
SimonDold: I agree that the statistic output would change this way.

I think it would be weird to keep it as is. If there is statistic that tells me the 'expansions' and a function named 'expand', then I would expect them to match in how often this function is called.

An additional alternative (that would not change the statistic) would be to add the generation of applicable operators to the `expand` function.
That would be line 126-141.

This would reduce the `step` function even more to
```
SearchStatus EagerSearch::step() {

    optional<SearchNode> candidate_node = get_next_node_to_expand();
    if (!candidate_node.has_value()) {
        assert(open_list->empty());
        log << "Completely explored state space -- no solution!" << endl;
        return FAILED;
    }

    SearchNode node = candidate_node.value();
    
    expand(node);
    return IN_PROGRESS;
}
```

src/search/search_algorithms/eager_search.cc
SimonDold: I agree that this could end up as a function in the base class. However, doing this would be out of scope for the story.
In my opinion this (that sibling classes could benefit from it, too) should not stop us from adding a `generate_succ_node` function to eager search if we think it is usefull for this class.

src/search/search_algorithms/eager_search.cc
maltehelmert: There is no general consensus in the literature what counts as an expansion, for example in the case where the goal state is "ready for expansion". Changing the definition is not really a refactoring but a semantic change that needs a good argument in the release notes to explain why it improves the planner. If we do it, the name of an internal function is not a good reason.

speckdavid: We kept the functionality as it was and encapsulated the expansion counter increment, the goal check and the successor generation within one expand function.

src/search/search_algorithms/eager_search.cc
speckdavid: I moved the line `bool is_preferred = preferred_operators.contains(op_id);`.

I also took a look at the suggested `generate_succ_node` function: we need both the `succ_state` and the `succ_node` variables in the subsequent code. So, I'm not sure if such a function will improve the quality of the code, as we would need to unpack the succ_state if we introduced such a function. For now, I would argue for leaving it as it is.

SimonDold:
LGTM

PR #270, opened 2025-07-25
[issue1189] Check style with clang-format instead of uncrustify

remochristen:
For now this pull request just exists so that we can see the diff.


src/search/cartesian_abstractions/cegar.cc
```diff
@@ -20,7 +20,8 @@
 using namespace std;
 
 namespace cartesian_abstractions {
-// Create the Cartesian set that corresponds to the given preconditions or goals.
+// Create the Cartesian set that corresponds to the given preconditions or
+// goals.
```
ClemensBuechner, line 24: Our coding conventions specify that multiline comments should use `/* ... */`. Not sure such a replacement can be automated using clang-format, so I'm not expecting any changes, but just wanted to point out the issue.

src/search/cartesian_abstractions/cegar.cc
```diff
@@ -99,14 +95,15 @@ CEGAR::CEGAR(
     if (log.is_at_least_normal()) {
         log << "Start building abstraction." << endl;
         log << "Maximum number of states: " << max_states << endl;
-        log << "Maximum number of transitions: "
-            << max_non_looping_transitions << endl;
+        log << "Maximum number of transitions: " << max_non_looping_transitions
```
ClemensBuechner, line 98: In our coding conventions, we say lines should preferrably be 72 charactars with some leeway up to 80 characters. This line originally obeyed to the 72 limit but was corrected by clang-format to run over into the leeway. I assume this happens in many other places as well. Is there any way to avoid that? Otherwise, we might want to change our guidelines.

src/search/cartesian_abstractions/cost_saturation.h
```diff
@@ -54,14 +54,11 @@ class CostSaturation {
...
-        const std::vector<std::shared_ptr<SubtaskGenerator>> &subtask_generators,
-        int max_states,
-        int max_non_looping_transitions,
-        double max_time,
-        PickSplit pick_split,
-        bool use_general_costs,
-        utils::RandomNumberGenerator &rng,
-        utils::LogProxy &log);
+        const std::vector<std::shared_ptr<SubtaskGenerator>>
+            &subtask_generators,
```
ClemensBuechner, line 58: This looks ugly in my opinion. I would prefer if the name would stay on the line of the type in this case, even if it runs over the line length.

src/search/cartesian_abstractions/utils_landmarks.cc
```diff
@@ -43,8 +44,8 @@ utils::HashMap<FactPair, LandmarkNode *> get_atom_to_landmark_map(
     const shared_ptr<LandmarkGraph> &graph) {
     // All landmarks are atomic, i.e., each has exactly one atom.
     assert(all_of(graph->begin(), graph->end(), [](auto &node) {
-                      return node->get_landmark().atoms.size() == 1;
-                  }));
+        return node->get_landmark().atoms.size() == 1;
+    }));
```
ClemensBuechner, line 48: Oh I like this change. :-)

src/search/cartesian_abstractions/utils_landmarks.h
```diff
@@ -23,7 +23,8 @@ extern std::shared_ptr<landmarks::LandmarkGraph> get_landmark_graph(
 extern std::vector<FactPair> get_atom_landmarks(
     const landmarks::LandmarkGraph &graph);
 
-extern utils::HashMap<FactPair, landmarks::LandmarkNode *> get_atom_to_landmark_map(
+extern utils::HashMap<FactPair, landmarks::LandmarkNode *>
+get_atom_to_landmark_map(
```
ClemensBuechner, line 27: Not a fan but I can live with it.

src/search/evaluators/combining_evaluator.h
```diff
@@ -46,10 +46,10 @@ class CombiningEvaluator : public Evaluator {
...
 extern void add_combining_evaluator_options_to_feature(
     plugins::Feature &feature, const std::string &description);
-extern std::tuple<std::vector<std::shared_ptr<Evaluator>>,
-                  const std::string, utils::Verbosity>
-get_combining_evaluator_arguments_from_options(
-    const plugins::Options &opts);
+extern std::tuple<
+    std::vector<std::shared_ptr<Evaluator>>, const std::string,
+    utils::Verbosity>
+get_combining_evaluator_arguments_from_options(const plugins::Options &opts);
```
ClemensBuechner, line 52: This style confuses me quite a bit. Not saying it was better before, though...

src/search/heuristics/cea_heuristic.cc
```diff
@@ -151,9 +146,9 @@ LocalProblem *ContextEnhancedAdditiveHeuristic::build_problem_for_variable(
             int target_value = dtg_trans.target->value;
             LocalProblemNode &target = problem->nodes[target_value];
             for (const ValueTransitionLabel &label : dtg_trans.labels) {
-                OperatorProxy op = label.is_axiom ?
-                    task_proxy.get_axioms()[label.op_id] :
-                    task_proxy.get_operators()[label.op_id];
+                OperatorProxy op =
+                    label.is_axiom ? task_proxy.get_axioms()[label.op_id]
+                                   : task_proxy.get_operators()[label.op_id];
```
ClemensBuechner, line 156: That's definitely clearer. :+1:

src/search/heuristics/cea_heuristic.cc
```diff
@@ -359,14 +354,16 @@ void ContextEnhancedAdditiveHeuristic::mark_helpful_transitions(
...
         if (first_on_path->target_cost == first_on_path->action_cost) {
             // Transition possibly applicable.
             const ValueTransitionLabel &label = *first_on_path->label;
-            OperatorProxy op = label.is_axiom ?
-                task_proxy.get_axioms()[label.op_id] :
-                task_proxy.get_operators()[label.op_id];
-            if (min_action_cost != 0 || task_properties::is_applicable(op, state)) {
+            OperatorProxy op = label.is_axiom
+                                   ? task_proxy.get_axioms()[label.op_id]
+                                   : task_proxy.get_operators()[label.op_id];
```
ClemensBuechner, line 369: Do you have any clue why this is formatted differently than the exact same statement in lines 149-151 of this file?

src/search/heuristics/cg_heuristic.cc
```diff
@@ -150,13 +150,15 @@ int CGHeuristic::get_transition_cost(const State &state,
 
                 // Scan labels of the transition.
                 for (ValueTransitionLabel &label : transition.labels) {
-                    OperatorProxy op = label.is_axiom ?
-                        task_proxy.get_axioms()[label.op_id] :
-                        task_proxy.get_operators()[label.op_id];
+                    OperatorProxy op =
+                        label.is_axiom
+                            ? task_proxy.get_axioms()[label.op_id]
+                            : task_proxy.get_operators()[label.op_id];
```
ClemensBuechner, line 156: And again another formatting for the same thing... This time it's probably due to line lengths with this much indent.

src/search/heuristics/max_heuristic.cc
```diff
@@ -121,11 +120,11 @@ class HSPMaxHeuristicFeature
...
 
-    virtual shared_ptr<HSPMaxHeuristic>
-    create_component(const plugins::Options &opts) const override {
+    virtual shared_ptr<HSPMaxHeuristic> create_component(
+        const plugins::Options &opts) const override {
         return plugins::make_shared_from_arg_tuples<HSPMaxHeuristic>(
-            relaxation_heuristic::get_relaxation_heuristic_arguments_from_options(opts)
-            );
+            relaxation_heuristic::
+                get_relaxation_heuristic_arguments_from_options(opts));
```
ClemensBuechner, line 127: Not a fan, but I can live with this.

src/search/landmarks/landmark_factory_hm.cc
```diff
@@ -14,6 +12,7 @@
 #include "../utils/markup.h"
 #include "../utils/system.h"
 
+#include <numeric>
```
ClemensBuechner, line 15: Nice that it detects missing includes.

src/search/landmarks/landmark_factory_hm.cc
```diff
@@ -38,10 +37,12 @@ static void set_difference(vector<int> &set1, const vector<int> &set2) {
...
-static bool are_mutex(const VariablesProxy &variables,
-                      const FactPair &atom1, const FactPair &atom2) {
-    return variables[atom1.var].get_fact(atom1.value).is_mutex(
-        variables[atom2.var].get_fact(atom2.value));
+static bool are_mutex(
+    const VariablesProxy &variables, const FactPair &atom1,
+    const FactPair &atom2) {
+    return variables[atom1.var]
+        .get_fact(atom1.value)
+        .is_mutex(variables[atom2.var].get_fact(atom2.value));
```
ClemensBuechner, line 45: I don't like this at all... I guess an easy solution to not get stuff like this would be to introduce variables for some required intermediate part before the return, so not necessarily an issue with the style.

src/search/landmarks/landmark_factory_merged.h
```diff
@@ -9,14 +9,17 @@ namespace landmarks {
...
     std::vector<std::shared_ptr<LandmarkFactory>> landmark_factories;
 
-    std::vector<std::shared_ptr<LandmarkGraph>> generate_landmark_graphs_of_subfactories(
+    std::vector<std::shared_ptr<LandmarkGraph>>
+    generate_landmark_graphs_of_subfactories(
         const std::shared_ptr<AbstractTask> &task);
-    void add_atomic_landmarks(
-        const std::vector<std::shared_ptr<LandmarkGraph>> &landmark_graphs) const;
+    void add_atomic_landmarks(const std::vector<std::shared_ptr<LandmarkGraph>>
+                                  &landmark_graphs) const;
```
ClemensBuechner, line 16: I would prefer the way it was before in this case.

src/search/landmarks/util.cc
```diff
@@ -12,7 +13,8 @@ namespace landmarks {
 static bool condition_is_reachable(
     const ConditionsProxy &conditions, const vector<vector<bool>> &reached) {
     for (FactProxy condition : conditions) {
-        if (!reached[condition.get_variable().get_id()][condition.get_value()]) {
+        if (!reached[condition.get_variable().get_id()]
+                    [condition.get_value()]) {
```
ClemensBuechner, line 17: Not sure how I feel about this. I struggled to accept this at first but after a bit of time I think it's fine and actually I wouldn't know a better way (except maybe introducing variable names for intermediate parts).

ClemensBuechner (commented):
I went through quite a few files, but not the entire codebase. I like most of the changes, and in particular that some things are enforced that previously weren't (e.g., line lengths, order of includes, double empty lines).

One thing I'm not particularly fond of is the bin-packing of function parameters. In some cases, I see value in having each argument on a new line, or grouping arguments with similarities on the same line. I'm not objecting to enforcing this, though, as long as others don't mind.
Similarly, I think it's fine if the initializer list of constructors has multiple arguments on the same line sometimes, which is now forbidden. In this case, however, the enforced rule puts everything on a line of it's own which i prefer over enforced bin-packing.

One thing that came to mind while we're at it: What are your opinions on enforcing braces around single-line bodies of if-statements etc.? I personally always put them because it makes breaking the code (e.g. by commenting out / removing a line) harder.

src/search/cartesian_abstractions/cegar.cc
remochristen: I don't see a way to make clang-format automatically adapt the comment style in case it gets too long. We could disallow reflowing comments entirely, but given that clang-format correctly maintains our multiline style while making comments fit into the 80 characters in cases where the comment style is already correct, I find it useful to have feature enabled even if it does the wrong thing here.

src/search/cartesian_abstractions/cegar.cc
remochristen: In our discussion we figured that it would be easiest to simplify the guidelines and only ask for a maximum of 80 characters, it doesn't seem feasible to have clang-format enforce the more nuanced style we had before.

src/search/cartesian_abstractions/cost_saturation.h
remochristen: I don't see a way to penalize this specific case, so we would have to relax the column limit globally, which I assume would make things unpredictable and inconsistent.

src/search/evaluators/combining_evaluator.h
remochristen: Yeah, breaking at the pointy brackets looks weird to me too, but they are treated the same way as the opening parentheses of functions for example so we cannot adapt this case independently.

src/search/cartesian_abstractions/cegar.cc
ClemensBuechner: I agree, it's better to stay within the margin. If we see the change after calling clang-format, we can still fix it by hand.

FlorianPommerening: I think it's fine to leave this up to people writing the code to make this change. As far as I know, we already have some violations of the guideline in the code base (i.e., some multi-line comments using `//`). Our policy so far was to clean this up as we spot them, so if this change breaks some existing comments, we'll find them eventually and clean them up then.

src/search/heuristics/cg_heuristic.cc
remochristen: Indeed it adapts the style based on the space available.
```c++
void f() {
    OperatorProxy op = label.is_axiom ? task_proxy.get_axioms()[label.op_id]
                                      : task_proxy.get_operators()[label.op_id];
    void g() {
        OperatorProxy op = label.is_axiom
                               ? task_proxy.get_axioms()[label.op_id]
                               : task_proxy.get_operators()[label.op_id];
        void h() {
            OperatorProxy op = label.is_axiom
                                   ? task_proxy.get_axioms()[label.op_id]
                                   : task_proxy.get_operators()[label.op_id];
            void i() {
                OperatorProxy op =
                    label.is_axiom ? task_proxy.get_axioms()[label.op_id]
                                   : task_proxy.get_operators()[label.op_id];
                void j() {
                    OperatorProxy op =
                        label.is_axiom
                            ? task_proxy.get_axioms()[label.op_id]
                            : task_proxy.get_operators()[label.op_id];
                }
            }
        }
    }
}
```

src/search/landmarks/landmark_factory_hm.cc
FlorianPommerening: I also stumbled over this and decided that it is fine in the end. I think this pattern is meant for code that uses a specific pattern where you have methods that modify an object and return it (The builder pattern is an example as far as I know). This is common in some code bases and then can look very clean. In this case I probably also would suggest to rewrite the code with local variables.

src/search/landmarks/landmark_factory_hm.cc
remochristen: Didn't realize that it did that!

src/search/landmarks/landmark_factory_merged.h
remochristen: Agreed, an unfortunate victim of being strict about the column limit.

src/search/cartesian_abstractions/cegar.cc
SimonDold: I think before merging would be a good situation to fix all occurrences of this violation and replace them with `/* ... */` comments.

remochristen:
Thanks for your comments Clemens!

clang-format is rather strict and often forces you to choose one style while disallowing the alternatives. Function parameters (and arguments) are such a case: Either it bin-packs everywhere, or it puts them on one line each (or a single line if they fit). So it's for example not possible to manually group parameters. Personally I prefer the bin-packed style.

It's a good point that the initializer list not being bin-packed is not consistent with the function style. I think the main reason why the initializer is not also bin-packed is that the line break before the colon becomes mandatory if bin-packing is enabled. With our current config the initializer list stays on a single line if it fits.

In terms of clang-format, I think it's not possible to force the braces, you can just control whether the statement is allowed to be on a single line or not, which is currently never allowed. Personally I don't mind the easy-to-break two line statements.

src/search/cartesian_abstractions/cost_saturation.h
maltehelmert: For cases like these, one option is always to introduce type aliases in the class with `using`, either for the vector of for the pointer.

src/search/landmarks/landmark_factory_merged.h
ClemensBuechner: So is there no way to prefer breaking after the opening parenthesis (just as if the function had multiple arguments, I think) rather than after the type?

remochristen: It would break after the opening parenthesis if everything after it fit on a single line, but with the `const` and all it ends up 2 characters too long in this case. With a slightly shorter variable name it would look like this:
```c++
    void add_atomic_landmarks(
        const std::vector<std::shared_ptr<LandmarkGraph>> &landmark_grap) const;
```
Breaking at the `const` would help too, but it doesn't do so and I can't find a way to enable something like that.

ClemensBuechner: Ah, I didn't consider that the line ends up being too long when formatted as in your example. Now I understand why it resorts to some other rule that adheres to the character limit, I can live with that.

ClemensBuechner: And I don't think that I would like it better if it broke at either one of the `const`s.

remochristen:
We merged the changes proposed in this pull request in #273 because this pull request became incompatible with the main branch. This pull request only exists to preserve the discussion but does not need to be merged anymore, so I'm closing it.

PR #271, opened 2025-07-29
[issue1185] Make the translator a python package.

roeger:
The translator is now a python package. This does not affect how Fast Downward is called,
but if you want to call the translator directly, please use 'python3 -m translate' instead of 
'python3 translate/main.py' (which replaced the previous 'translate.py').

### DIFF
--- a/src/translate/options.py
+++ b/src/translate/options.py
@@ -87,6 +87,10 @@ def parse_args(args=None):
 
 
 def get_options():
+    if options is None:
+        msg = ("No options provided (via options.set_options(...)). For example"
+               " 'options.set_options([<domain file>, <problem_file>])'.")
+        raise RuntimeError(msg)
     return options
 
 


roeger:
This is not ready to be merged, yet, because of an ugly output from argparse. This will only require a local change, so you can nevertheless already have a look at the diff.

driver/call.py
```diff
@@ -50,6 +50,11 @@ def fail(exception, exitcode):
     else:
         return set_limits
 
+def _get_env(prepend_python_path=None):
```
FlorianPommerening: This could use a doc string or comment to explain why we need it. In general adding something to the (Python)path feels a bit hacky. As far as I understand, we only do this because the translator is not installed in the way that other Python packages are. In the long run, we might want to change that and publish the translator as its own package. Then, ti would be installed as a dependency and messing with the path would not be necessary. I suggest to mention this here in a comment.

jendrikseipp:
```suggestion
def _get_env(prepend_to_python_path=None):
```
The current version sounds to me like it's a Boolean option.

driver/run_components.py
```diff
@@ -71,15 +70,20 @@ def run_translate(args):
...
         args.translate_memory_limit, args.overall_memory_limit)
-    translate = get_executable(args.build, REL_TRANSLATE_PATH)
+
+    # Check existence of translate in build.
+    get_executable(args.build, REL_TRANSLATE_PATH)
+
     assert sys.executable, "Path to interpreter could not be found"
-    cmd = [sys.executable] + [translate] + args.translate_inputs + args.translate_options
+    cmd = [sys.executable] + ["-m", "translate"] + args.translate_inputs + args.translate_options
+    path = str(util.BUILDS_DIR / args.build / "bin")
```
FlorianPommerening: I think a nicer way to do this would be to use the result of `get_executable` from above because that is where we checked for the existence of the translator:
```Python
translator = get_executable(args.build, REL_TRANSLATE_PATH)
# ...
prepend_python_path=translator.parent
```

src/translate/build_model.py
```diff
@@ -4,8 +4,8 @@
 import sys
 import itertools
 
-import pddl
-import timers
+from translate import pddl
+from translate import timers
 from functools import reduce
```
FlorianPommerening: should this be in the block of system includes above?

src/translate/pddl/pddl_types.py
```diff
@@ -65,6 +65,6 @@ def uniquify_name(self, type_map, renamings):
 
     def get_atom(self):
         # TODO: Resolve cyclic import differently.
-        from . import conditions
+        from translate.pddl import conditions
```
FlorianPommerening, line 68: 😱

driver/call.py
roeger: I added a doc string.

roeger: I changed this to the suggested name.

driver/run_components.py
roeger: Very true. I changed this.

src/translate/build_model.py
roeger: I reordered the imports.

src/translate/pddl/pddl_types.py
roeger: It means "In the distant future, resolve cyclic import differently."

FlorianPommerening: I didn't mean this as a request to do anything now. I just shuddered when I saw the import in the middle of a function. It's not related to this issue at all, but it's a bad code smell, isn't it? Feels like the time we had an `#include "some_more_code.cc"` in the middle of the open list code.

src/translate/build_model.py
FlorianPommerening: They are still not sorted alphabetically but they weren't before as well.

roeger: How exactly should they be sorted? Full lines lexicographically (from ... first), by package name (again from ... first) or by the imported thing (current order)?

FlorianPommerening: Good question. I was thinking by package name but I didn't think about other options. I looked into this and there is a PEP8 recommendation for it but it mostly talks about how to group imports and the order of the groups. I tried isort (`pip install isort`) and it orders this case like this:
```Python
import itertools
import sys
from functools import reduce

from translate import pddl, timers
```

so I guess full lines lexicographically within a block.

src/translate/options.py
```diff
@@ -1,9 +1,31 @@
...
+def infer_prog():
+    main_mod = sys.modules['__main__']
+    spec = getattr(main_mod, '__spec__', None)
+
+    if spec is not None:
+        # Invoked via `python -m ...`
+        module_name = spec.name
+        if module_name.endswith('.__main__'):
+            module_name = module_name.rsplit('.', 1)[0]
+        python_exec = os.path.basename(sys.executable)
```
FlorianPommerening:
```suggestion
        python_exec = Path(sys.executable).name
```
maybe slightly more readable (also change `import os` to `from pathlib import Path`).

src/translate/build_model.py
haz: Why not run [black](https://black.readthedocs.io/en/stable/) or similar to let the auto-tools make the call?

src/translate/options.py
```diff
@@ -1,9 +1,31 @@
...
+        return f"{python_exec} -m {module_name}"
+    else:
+        # Invoked as script directly.
+        return os.path.basename(sys.argv[0])
```
FlorianPommerening:
```suggestion
        return Path(__file__).name
```
(not sure if `__file__` works. it might have to stay `sys.argv[0]`)

src/translate/build_model.py
FlorianPommerening: That's [another story](https://github.com/aibasel/downward/pull/270/files). I think it would make sense to use a tool for formatting the Python code as well but that is clearly out of scope for this issue.

roeger: Black would go for:
```
from functools import reduce
import itertools
import sys
````

I change it to this order (but only in this file, I do not want to have too many drive-by changes that are not related to the issue).

File: src/translate/options.py (hunk 5 above)
Comment (roeger) [reply src/translate/options.py]: Will change this.

File: src/translate/options.py (hunk 6 above)
Comment (roeger) [reply src/translate/options.py]: `__file__` would be `options.py`, this is not what we want. I will make the switch from `os.path` to `pathlib`.

**Comment by FlorianPommerening** [comment]
Looks good to me

File: src/translate/options.py [hunk 7]
```diff
@@ -60,18 +83,17 @@ def parse_args():
...
+    if options is None:
+        options = parse_args()
+    return options
 
 
-setup()
+def use_default_options():
+    args = "<dummy domain> <dummy problem>"
+    global options
+    options = parse_args(args)
```
maltehelmert: I don't think this works, `args` must be a list, such as `["<dummy domain>", "<dummy problem>"]`. Arguments are split into a list already by the shell, before the program sees them. But also, I now think there is a better way to do this: since the responsibility of not setting the filenames properly ultimately rests with the user, I think *they* should set the dummy filenames.

So I would go with your original design of a `setup` action, call it `set_options`, and define it like this:

```
def set_options(arguments=None):
    global options
    options = parse_args(arguments)
```
    
Then the external code can explicitly call this as `set_options(["<dummy>", "<dummy>"])` to make clear they are not providing filenames, and our own code can use the default `set_options()`. We can then simplify `get_options()` to `def get_options(): return options` (of course with a line break), and `use_default_options` can go away.

We can also go one step further and integrate the two functions `parse_args` and `set_options` into one.

maltehelmert: (And then of course this also means that external code can set other options than the default, which is nice.)

roeger: Should I then raise an exception if `get_options` is called but options have not been set? This might be a good place to direct the user to the right way to do this.

roeger: I added the exception.

FlorianPommerening:
I didn't look at the changes that Malte already looked at but the remaining changes look good to me.

PR #273, opened 2025-07-30
[issue1189] Check style with clang-format instead of uncrustify -- Mergable

remochristen:
Changes merged into `main` caused conflicts with the original pull request for issue1198 at #270 and we couldn't find a way to resolve them without having our new style checks fail in the pull request. In order to not loose the conversation from the original pull request, I'm creating this new pull request which has been rebased on the main branch and thus does not contain the same conflicts.


.github/workflows/style.yml
```diff
@@ -25,23 +25,12 @@ jobs:
...
-          tar xzf uncrustify-0.72.0.tar.gz
-          cd uncrustify-uncrustify-0.72.0
-          mkdir build
-          cd build
-          cmake ../
-          make -j2
-          mv uncrustify ../../
-          cd ../../
-          rm -rf uncrustify-0.72.0.tar.gz uncrustify-uncrustify-0.72.0
+          sudo apt-get -y install clang-format
```
TravisRiveraPetit, line 30: 🔥

misc/style/run-all-style-checks.py
```diff
@@ -51,7 +51,7 @@ def check_cc_files():
 
 
 def check_cplusplus_style():
```
TravisRiveraPetit, line 61: Is this function even called?

misc/style/run-all-style-checks.py
```diff
@@ -51,7 +51,7 @@ def check_cc_files():
...
-    return subprocess.call(["./run-uncrustify.py"], cwd=DIR) == 0
+    return subprocess.call(["./run-clang-format.py"], cwd=DIR) == 0
 
 
 def main():
```
TravisRiveraPetit, line 67: Okay check_cplusplus_style is called but in a strange way.

misc/style/run-clang-format.py
```diff
@@ -42,21 +42,24 @@ def main():
...
+    cmd = [
+        executable, flag, "--Werror", f"--style=file:{config_file}"
+    ] + src_files
     try:
         # Hide clean files printed on stdout.
         returncode = subprocess.call(cmd, stdout=subprocess.PIPE)
-    except FileNotFoundError:
-        sys.exit(f"Error: {executable} not found. Is it on the PATH?")
+    except FileNotFoundError as not_found:
+        src_error_str = f"ERROR: Did not find file: '{not_found.filename}'."
```
TravisRiveraPetit: ERROR or Error? Inconsistent with line 48. Suggestion: use Error to be consistent with previous commit.

TravisRiveraPetit (changes requested):
Python test are set up in a strange way. Minor naming inconsistencies. Otherwise LGTM.

.github/workflows/style.yml
SimonDold: thx

misc/style/run-all-style-checks.py
SimonDold: I made the call more explicit (hopefully less strange).

PR #283, opened 2026-02-10
[issue1201] Check whether a configuration guarantees completeness

grucla:


src/search/open_lists/alternation_open_list.cc
```diff
@@ -124,6 +125,16 @@ bool AlternationOpenList<Entry>::is_reliable_dead_end(
...
 
+template<class Entry>
+bool AlternationOpenList<Entry>::pruning_is_safe() const {
+    for (const auto &sublist: open_lists) {
+        if (sublist->pruning_is_safe()) {
+            return true;
+        }
+    }
+    return false;
+}
```
jendrikseipp, line 139: This computes a logical OR, where we want an AND, I think. I recommend to use std::all_of here and in some places below.

maltehelmert: I would expect "or" to be correct for alternation.

grucla: Tanja and I were not sure about this but went with OR because we thought that for the completeness of the configuration it is enough if one of the open lists is "safe", i.e. ensures that all non-dead-ends are considered.

src/search/search_algorithms/eager_search.cc
```diff
@@ -318,6 +318,16 @@ void EagerSearch::dump_search_space() const {
...
+bool EagerSearch::is_complete() const {
+    if (!open_list->pruning_is_safe()) {
+        return false;
+    }
+    if (!pruning_method->is_safe()) {
+        return false;
+    }
+    return true;
+}
+
```
jendrikseipp, line 323:
```suggestion
return open_list->pruning_is_safe() && pruning_method->is_safe();
```

src/search/open_list.h
```diff
@@ -122,6 +122,9 @@ class OpenList {
     virtual bool is_dead_end(EvaluationContext &eval_context) const = 0;
     virtual bool is_reliable_dead_end(
         EvaluationContext &eval_context) const = 0;
+    virtual bool pruning_is_safe() const {
+      return false;
+    }
```
jendrikseipp: Make this a pure virtual function, like the ones above?

src/search/planner.cc
```diff
@@ -53,9 +53,14 @@ int main(int argc, const char **argv) {
...
-                                ? ExitCode::SUCCESS
-                                : ExitCode::SEARCH_UNSOLVED_INCOMPLETE;
+        ExitCode exitcode;
+        if (search_algorithm->found_solution()) {
+            exitcode = ExitCode::SUCCESS;
+        } else if (search_algorithm->is_complete()) {
+            exitcode = ExitCode::SEARCH_UNSOLVABLE;
+        } else {
+            exitcode = ExitCode::SEARCH_UNSOLVED_INCOMPLETE;
+        }
```
jendrikseipp, line 94: Can't it also be that the search ran out of time?

src/search/search_algorithm.h
```diff
@@ -77,6 +77,9 @@ class SearchAlgorithm {
     virtual ~SearchAlgorithm();
     virtual void print_statistics() const = 0;
     virtual void save_plan_if_necessary();
+    virtual bool is_complete() const {
+        return false;
+    }
```
jendrikseipp: This should also become a pure virtual function eventually, but I guess you want to minimize the diff until more algorithms are supported.

src/search/open_lists/alternation_open_list.cc
jendrikseipp: You're right. My knee-jerk reaction was wrong and I should've thought about this more carefully. But I think this method deserves a comment :) 

I had some other minor comments while looking over this, see below.

src/search/open_lists/tiebreaking_open_list.cc
```diff
@@ -140,6 +141,30 @@ bool TieBreakingOpenList<Entry>::is_reliable_dead_end(
...
+    assert(!evaluators.empty());
+    if (evaluators[0]->dead_ends_are_reliable()) {
+        return true;
+    }
+    // At this point we know that the first evaluator is unsafe.
+    if (allow_unsafe_pruning) {
+        return false;
+    }
+    // Even if the first evaluator is unsafe we can still say that
+    // pruning_is_safe is true if (allow_unsafe_pruning is false and) at least
```
jendrikseipp:
```suggestion
    // pruning is safe if (allow_unsafe_pruning is false and) at least
```

src/search/search_algorithm.h
grucla: I think we could also keep it as it is to have a default value for future search algorithms but I do not have a strong opinion on this. For the moment I would definitely keep it like this because the completeness check is not yet implemented for all search algorithms.

src/search/open_list.h
grucla: Same discussion as with is_complete of search_algorithm.h. For the moment I would keep it like this because the completeness check is not yet implemented for all open lists.

src/search/planner.cc
grucla: Looks like it but I'm not completely sure. In that case it should be enough to extend the check to something like `search_algorithm->is_complete() && search_finished_in_time`, right? (I haven't yet looked into how to do the second part of the check.)

jendrikseipp: Yes, I think so.

src/search/open_lists/tiebreaking_open_list.cc
```diff
@@ -140,6 +141,30 @@ bool TieBreakingOpenList<Entry>::is_reliable_dead_end(
...
+        return true;
+    }
+    // At this point we know that the first evaluator is unsafe.
+    if (allow_unsafe_pruning) {
+        return false;
+    }
+    // Even if the first evaluator is unsafe we can still say that
+    // pruning is safe if (allow_unsafe_pruning is false and) at least
+    // one other evaluator is safe.
+    for (auto eval : evaluators) {
```
SimonDold:
```suggestion
    for (const shared_ptr<Evaluator> &evaluator : evaluators)
```
to be consistent with e.g. `get_path_dependent_evaluators` in this file
and to have less updates on the reference counter of the smart pointers in `evaluators`.

src/search/open_lists/alternation_open_list.cc
```diff
@@ -124,6 +125,18 @@ bool AlternationOpenList<Entry>::is_reliable_dead_end(
     return false;
 }
 
+template<class Entry>
+bool AlternationOpenList<Entry>::pruning_is_safe() const {
+    // If at least one of the sub-lists ensures that no solvable state is
+    // pruned we know that this also holds for AlternationOpenList.
```
SimonDold: https://www.fast-downward.org/latest/for-developers/style-conventions/#comments

SimonDold: https://www.fast-downward.org/latest/for-developers/style-conventions/#comments

src/search/planner.cc
grucla: I ran a blind search with `--overall-time-limit 10s` and also with `--search-time-limit 10s`. In both cases I got exit code 23 `SEARCH_OUT_OF_TIME` which makes me believe that this code line is not reached if the search runs out of time. Are there other ways to restrict the time such that this code line could be reachable?

maltehelmert: The time limits you mention are enforced externally by the fast-downward.py script, which terminates the search process from the outside when the time limit is exceeded.

I think Jendrik meant an independent mechanism in the search algorithms that uses the "max_time" parameter that all algorithms have. We document this on fast-downward.org as follows:

```max_time (double): maximum time in seconds the search is allowed to run for. The timeout is only checked after each complete search step (usually a node expansion), so the actual runtime can be arbitrarily longer. Therefore, this parameter should not be used for time-limiting experiments. Timed-out searches are treated as failed searches, just like incomplete search algorithms that exhaust their search space.```

For the reasons described in the documentation I think we basically never use this parameter (and I'm not so sure it's good that it exists). The documentation says that these should be treated as failed searches like incomplete searches, but we probably should discuss what we really want to do in terms of whether we should treat search algorithms with a finite max_time value as incomplete and whether we should signal running out of time in this way with SEARCH_OUT_OF_TIME. Right now we don't, we return SEARCH_UNSOLVED_INCOMPLETE. You can test it like this, adapting the benchmark path:

```./fast-downward.py ~/downward/benchmarks/gripper/prob10.pddl --search "astar(blind(), max_time=1)"; echo $?```

This prints the numeric exit code at the end, and you can see the symbolic name for it here:
https://www.fast-downward.org/latest/documentation/exit-codes/

File: src/search/evaluators/combining_evaluator.cc [hunk 9]
```diff
@@ -17,11 +17,11 @@ CombiningEvaluator::CombiningEvaluator(
...
     all_dead_ends_are_reliable = true;
     for (const shared_ptr<Evaluator> &subevaluator : subevaluators)
-        if (!subevaluator->dead_ends_are_reliable())
+        if (!subevaluator->is_safe())
             all_dead_ends_are_reliable = false;
 }
 
-bool CombiningEvaluator::dead_ends_are_reliable() const {
+bool CombiningEvaluator::is_safe() const {
     return all_dead_ends_are_reliable;
```
SimonDold: should `all_dead_ends_are_reliable` be renamed to `all_subevaluators_are_safe` then?

src/search/open_list.h
```diff
@@ -122,6 +122,12 @@ class OpenList {
     virtual bool is_dead_end(EvaluationContext &eval_context) const = 0;
     virtual bool is_reliable_dead_end(
         EvaluationContext &eval_context) const = 0;
+    /*
+      is_complete returns true if the open list cannot "overlook" any solvable
+      state. Thus, if the open list is "complete" then it can be used to detect
```
SimonDold: I like the comment from the pareto open list more

How about
```
/* is_complete returns true if no solvable state
       is pruned.  */ ```

File: src/search/search_algorithm.h [hunk 11]
```diff
@@ -77,6 +77,11 @@ class SearchAlgorithm {
     virtual ~SearchAlgorithm();
     virtual void print_statistics() const = 0;
     virtual void save_plan_if_necessary();
+    /*
+      is_complete returns true if the search algorithm cannot "overlook" any
+      solvable state.
```
SimonDold: I don't think it is about solvable states but alive states (e.g. solvable AND reachable)

`astar(blind(), bound=0)` does this search only overlook unsolvable states?
What about a binary counter task where each state is solvable?

SimonDold (commented):
In some for-loops you use auto in others you mention the type of the elements in the loop.
It is consistent within the file but I do not know if there is one suggestion over the other.

As mentioned on discord "astar(blind(), bound=0)" produces results that are misleading for the user.

The rest looks fine to me but I am not sure about right semantics for the different open lists.

src/search/open_lists/alternation_open_list.cc
```diff
@@ -37,6 +38,7 @@ class AlternationOpenList : public OpenList<Entry> {
     virtual bool is_dead_end(EvaluationContext &eval_context) const override;
     virtual bool is_reliable_dead_end(
```
gutodelazeri, line 39: Shouldn't this be renamed to `is_safe()`?

SimonDold: the idea is that `is_safe` is a global property of a heuristic.
But `is_reliable_dead_end` is a local property.
This is more apparent in `tiebreaking_open_list.cc`. The `is_reliable_dead_end` check asks "is there a heuristic that (1) is safe and (2) thinks the state is unsolvable?"

from `open_list.h`:
```
    /*
      is_dead_end and is_reliable_dead_end return true if the state
      associated with the passed-in evaluation context is deemed a
      dead end by the open list.

      The difference between the two methods is that
      is_reliable_dead_end must guarantee that the associated state is
      actually unsolvable, i.e., it must not believe the claims of
      unsafe heuristics.

      Like OpenList::insert, the methods usually evaluate heuristic
      values, which are then cached in eval_context as a side effect.
    */
```

src/search/search_algorithm.h
```diff
@@ -77,8 +80,14 @@ class SearchAlgorithm {
     virtual ~SearchAlgorithm();
     virtual void print_statistics() const = 0;
     virtual void save_plan_if_necessary();
+    /*
+      is_complete_within_bound returns true if the search algorithm finds
+      a plan within the bound if a plan exists.
```
tanjaschindler: Should this comment say explicitly that the bound is exclusive?

src/search/evaluators/combining_evaluator.h
```diff
@@ -24,19 +24,18 @@ class CombiningEvaluator : public Evaluator {
         const std::string &description, utils::Verbosity verbosity);
 
     /*
-      Note: dead_ends_are_reliable() is a state-independent method, so
-      it only returns true if all subevaluators report dead ends reliably.
+      Note: is_safe() is a state-independent method, so
+      it only returns true if all subevaluators are safe.
 
       Note that we could get more fine-grained information when
       considering of reliability for a given evaluated state. For
```
tanjaschindler:
```suggestion
      considering reliability for a given evaluated state. For
```

src/search/open_list.h
```diff
@@ -122,6 +122,11 @@ class OpenList {
     virtual bool is_dead_end(EvaluationContext &eval_context) const = 0;
     virtual bool is_reliable_dead_end(
         EvaluationContext &eval_context) const = 0;
+    /*
+      If for some solvable task no reachable goal state is ever inserted
+      into the open list then it is not complete, i.e. is_complete is false.
+    */
```
tanjaschindler: I would prefer this to be phrased in a positive way, like "is_complete is true if for all solvable tasks it is guaranteed that some reachable goal state is eventually inserted into the open list"

src/search/landmarks/landmark_sum_heuristic.cc
```diff
@@ -103,7 +103,7 @@ int LandmarkSumHeuristic::get_heuristic_value(const State &ancestor_state) {
     return h;
 }
 
-bool LandmarkSumHeuristic::dead_ends_are_reliable() const {
+bool LandmarkSumHeuristic::is_safe() const {
     return dead_ends_reliable;
```
tanjaschindler: The documentation says it is safe except on tasks with conditional effects when using a LandmarkFactory not supporting them but dead_ends_reliable is also set to false if the task has axioms. Which one is correct?

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
```diff
@@ -78,7 +78,7 @@ int LandmarkCostPartitioningHeuristic::get_heuristic_value(
     }
 }
 
-bool LandmarkCostPartitioningHeuristic::dead_ends_are_reliable() const {
+bool LandmarkCostPartitioningHeuristic::is_safe() const {
```
tanjaschindler, line 80: Is the plan still to remove the default value of is_safe? Otherwise, the default can be used here.

src/search/heuristics/relaxation_heuristic.cc
```diff
@@ -110,7 +110,7 @@ RelaxationHeuristic::RelaxationHeuristic(
     }
 }
 
-bool RelaxationHeuristic::dead_ends_are_reliable() const {
+bool RelaxationHeuristic::is_safe() const {
     return !task_properties::has_axioms(task_proxy);
```
tanjaschindler: The documentation of hmax and hadd says safe even with axioms. What is correct?

src/search/pruning_method.h
```diff
@@ -37,6 +37,9 @@ class PruningMethod {
     virtual void initialize(const std::shared_ptr<AbstractTask> &task);
     void prune_operators(const State &state, std::vector<OperatorID> &op_ids);
     virtual void print_statistics() const;
+
+    // is_safe returns true if no solvable state can be pruned.
```
tanjaschindler: Is this really the intended definition? With stubborn sets, if we have two actions that can be applied in arbitrary order, wouldn't it be possible that one of the intermediate states would be pruned, even if it is solvable?

src/search/pruning/stubborn_sets.h
```diff
@@ -58,6 +58,9 @@ class StubbornSets : public PruningMethod {
 public:
     explicit StubbornSets(utils::Verbosity verbosity);
     virtual void initialize(const std::shared_ptr<AbstractTask> &task) override;
+    virtual bool is_safe() const override {
+        return true;
```
tanjaschindler, line 63: See comment in pruning_method.h

src/search/landmarks/landmark_sum_heuristic.cc
SimonDold: This inconsistency was introduced by a trivial commit 
https://github.com/aibasel/downward/commit/036fffa1167cea530efb2f67aca6f82bb005a2c9#diff-7328f2732dbc19ea95eef2e419a59f77a73c4fae57fd13b15fc33fb3e5c68c61

which was a follow-up on the commit related to issue454
https://github.com/aibasel/downward/commit/f75018fd06a3ad476ffff794fa92edabd578e8ab

@roeger you are involved in this issue as well as in issue454. So I want to forward the question to you.

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
SimonDold: I don't think there is a consensus on that.

I would like to ignore this drive-by clean-up (I added a comment in issue1207 to remove it in the case we want to keep the default for `is_safe`).
The other options are:
- removing it now as a drive-by clean-up (given the main we are merging it to)
- waiting for a consensus on issue1207

What is you suggestion?

src/search/open_list.h
SimonDold: Those are not the same statements.

If we are unsure of the mathematical guarantees we should report `true` for the is_complete check to not violate your statement (correct or false-positive).

For the other statement we can report `false` for the is_complete check when we are unsure (correct or false-negative).

Having false-positives for the completeness check is much worse than false-negatives.

I think we actually have to rephrase the other comments too.

src/search/pruning_method.h
grucla: I adjusted the definition based on our discussion yesterday. What do you think about the new one?

SimonDold: I think the `(optimal)` is more distracting than helpful, as we are not reporting with this boolean whether or not this is something useful for deriving the optimality but only for deriving completeness.

I think this is still the wrong direction of the implication.

suggestion:
```
    // is_safe returns true only if, for each solvable non-goal state, at least one
    // action is kept that starts a plan from that state.
```

grucla: I'm fine with removing the `(optimal)` to make the comment easier to understand but I do not see how the direction of the implication is wrong.

SimonDold: The way 
```
    // is_safe returns true if for each solvable non-goal state at least on
    // action is kept that starts an (optimal) plan from that state.
```
reads to me as the implication:
`(is_safe == true )  <== PROPERTY`

If I receive the information `is_safe == true` I can't deduce anything, the property could hold or not.

I think we should have it the other direction:
`(is_safe == true) ==> PROPERTY`

If I receive the information `is_safe == true` now I can deduce that the property holds.

grucla: I overlooked the "suggestion:" part of your first comment and thought that was how I phrased it in the code. In that case you are of course right, it should be "is_safe returns true **only** if [...]" instead of "is_safe returns true if [...]".

grucla: Fixed.

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
grucla: I think we can ignore this for now and wait for a consensus on issue1207.

src/search/evaluator.h
```diff
@@ -30,12 +30,12 @@ class Evaluator : public components::TaskSpecificComponent {
         utils::Verbosity verbosity);
 
     /*
-      dead_ends_are_reliable should return true if the evaluator is
-      "safe", i.e., infinite estimates can be trusted.
+      is_safe returns true only if the evaluator reports dead ends reliably,
+      i.e., if its infinite estimates can be trusted.
```
ClemensBuechner:
```suggestion
      i.e., if states with infinite heuristic value can safely be pruned.
```

src/search/evaluator.cc
```diff
@@ -20,7 +20,7 @@ Evaluator::Evaluator(
       log(utils::get_log_for_verbosity(verbosity)) {
 }
 
-bool Evaluator::dead_ends_are_reliable() const {
+bool Evaluator::is_safe() const {
     return true;
```
ClemensBuechner, line 24: During discussions, it kept coming up that we're generally conservative in the sense that if we're unsure whether a heuristic is safe, then we prefer to say `false` rather than potentially wrongly claiming `true`. In that sense, wouldn't it be more conservative to `return false` in the base class of `Evaluator`?

src/search/search_algorithms/iterated_search.cc
```diff
@@ -112,10 +113,41 @@ shared_ptr<SearchAlgorithm> IteratedSearch::bind_current_search() {
...
+    SearchStatus current_search_status =
+        task_specific_search->get_finished_search_status();
+    assert(current_search_status != SearchStatus::IN_PROGRESS);
+
+    switch (current_search_status) {
+    case SearchStatus::SOLVED:
+    case SearchStatus::UNSOLVABLE:
+        best_status = current_search_status;
+        break;
+
```
ClemensBuechner: Does clang support/enforce these empty lines? They look weird to me...

src/search/search_algorithms/iterated_search.cc
```diff
@@ -112,10 +113,41 @@ shared_ptr<SearchAlgorithm> IteratedSearch::bind_current_search() {
...
+    case SearchStatus::UNSOLVABLE_WITHIN_BOUND:
+        // We ignore inner bounds if the search did not check up to the outer
+        // bound.
```
ClemensBuechner:
```suggestion
        /* We ignore inner bounds if the search did not check up to the outer
           bound. */
```

src/search/pruning_method.h
```diff
@@ -38,6 +38,10 @@ class PruningMethod : public components::TaskSpecificComponent {
     virtual void initialize(const std::shared_ptr<AbstractTask> &task);
     void prune_operators(const State &state, std::vector<OperatorID> &op_ids);
     virtual void print_statistics() const;
+
+    // is_safe returns true only if for each solvable non-goal state at least
+    // one action is kept that starts a plan from that state.
```
ClemensBuechner:
```suggestion
    /* is_safe returns true only if for each solvable non-goal state at least
       one action is kept that starts a plan from that state. */
```

src/search/landmarks/landmark_sum_heuristic.cc
```diff
@@ -15,7 +15,7 @@
 using namespace std;
 
 namespace landmarks {
-static bool are_dead_ends_reliable(
+static bool compute_safe(
```
ClemensBuechner, line 18: There is some repetition of the same rules for safety here and in the cost partitioning landmark heuristic. It seems somewhat reasonable to duplicate this because the heuristics could behave differently, so this is not a suggestion to reduce duplication by defining this in one common place. I nevertheless wanted to share the thought.

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
```diff
@@ -77,7 +77,9 @@ int LandmarkCostPartitioningHeuristic::get_heuristic_value(
...
+    /* Since check_unsupported_features prevents instantiating this heuristic
+       on tasks where it could be unsafe, we can return true here. */
```
ClemensBuechner, line 82: Should we assert these requirements here, then?

src/search/heuristics/relaxation_heuristic.cc
```diff
@@ -109,10 +109,6 @@ RelaxationHeuristic::RelaxationHeuristic(
     }
 }
 
-bool RelaxationHeuristic::dead_ends_are_reliable() const {
-    return !task_properties::has_axioms(task_proxy);
```
ClemensBuechner, line 113: Removing this here changes the semantics, right? Because now the base version of the function `is_safe()` would be used which always returns `true`. If you're confident this holds, you can simply ignore this comment.

src/search/open_lists/alternation_open_list.cc
ClemensBuechner: I just had a discussion about this with Claudia because I stumbled upon it again during my review just now. I only now see this discussion but agree with Jendrik that this deserves a comment. :)

My main confusion was as follows: if all heuristics were safe, a single one evaluating to infinity would be enough to prune a state. So my uninformed mental modal was that the search might prune states which are wrongly declared dead ends by some unsafe heuristic. If this were the case, then I would be surprised if the heuristic is safe as long as any of the heuristics is safe. After looking at the function `AlternationOpenList<Entry>::is_dead_end`, I assume that my mental model was too simplistic and I can see why `any_of` makes sense here.

ClemensBuechner (approved):
I generally approve of the changes but have a few non-critical comments directly in the code. Additionally, one pattern that occurs often documenting comments is the following:
```
// <function name> returns <value> if <description>.
<type> <function name> ...
```
I suggest to abbreviate the documentation in these cases to
```
// Returns <value> if <description>.
```
because it should be clear that it concerns the function that follows below the documentation. Also, I don't like sentences starting with lowercase letters and our function names typically do start with lowercase letters.

src/search/evaluator.cc
grucla: This is what we want to discuss in one of the next Fast Downward meetings, see also issue1207.

src/search/landmarks/landmark_cost_partitioning_heuristic.cc
grucla: That was our original plan but we cannot access the LandmarkFactory outside the constructor which we would need for the assert.

src/search/pruning_method.h
grucla: Done.

src/search/search_algorithms/iterated_search.cc
grucla: Done.

src/search/search_algorithms/iterated_search.cc
grucla: Changed.

src/search/evaluator.h
grucla: We rephrased it but slightly differently than your suggestion.

PR #297, opened 2026-07-22
[issue1222] Add translator option for representing disjunctive formulas using derived variables

speckdavid:
The translator has now an option `condition-normalization-strategy` which allows to choice between the following two options.
- `dnf` (default): The previous strategy to handle disjunctions in conditions. It produces a dnf and splits the condition. For the goal it may substitute at the highest level using axioms.
- `axioms_based` (new): Decompose disjunctive formulas using derived predicates. It replaces each disjunctive subformula with a derived predicate and add a corresponding axiom defining that predicate. The transformation proceeds bottom-up so nested disjunctions are eliminated first.

### DIFF
--- a/src/translate/main.py
+++ b/src/translate/main.py
@@ -695,8 +695,8 @@ def main():
         task = pddl_parser.open(
             domain_filename=get_options().domain, problem_filename=get_options().problem)
 
-    with timers.timing("Normalizing task"):
-        normalize.normalize(task)
+    with timers.timing(f"Normalizing task with strategy {get_options().condition_normalization_strategy}"):
+        normalize.normalize(task, get_options().condition_normalization_strategy)
 
     if get_options().generate_relaxed_task:
         # Remove delete effects.
--- a/src/translate/main.py
+++ b/src/translate/main.py
@@ -696,7 +696,7 @@ def main():
             domain_filename=get_options().domain, problem_filename=get_options().problem)
 
     with timers.timing(f"Normalizing task with strategy {get_options().condition_normalization_strategy}"):
-        normalize.normalize(task, get_options().condition_normalization_strategy)
+        normalize.normalize(task)
 
     if get_options().generate_relaxed_task:
         # Remove delete effects.
--- a/src/translate/main.py
+++ b/src/translate/main.py
@@ -695,7 +695,7 @@ def main():
         task = pddl_parser.open(
             domain_filename=get_options().domain, problem_filename=get_options().problem)
 
-    with timers.timing(f"Normalizing task with strategy {get_options().condition_normalization_strategy}"):
+    with timers.timing("Normalizing task"):
         normalize.normalize(task)
 
     if get_options().generate_relaxed_task:
--- a/src/translate/normalize.py
+++ b/src/translate/normalize.py
@@ -4,6 +4,8 @@ import copy
 from typing import Sequence
 
 from translate import pddl
+from translate.options import get_options
+
 
 class ConditionProxy:
     def clone_owner(self):
@@ -370,16 +372,14 @@ def substitute_complicated_goal(task):
 # Combine Steps [1], [2], [3], [4], [5] and do some additional verification
 # that the task makes sense.
 
-
-def normalize(task, condition_strategy):
+def normalize(task):
     remove_universal_quantifiers(task)
-    if condition_strategy == "dnf":
-        substitute_complicated_goal(task)
-        build_DNF(task)
-    elif condition_strategy == "axiom_based":
-        substitute_conditions_with_axioms(task)
-    else:
-        raise ValueError("Unknown condition normalization strategy: %s" % condition_strategy)
+    match get_options().condition_normalization_strategy:
+        case "dnf":
+            substitute_complicated_goal(task)
+            build_DNF(task)
+        case "axiom_based":
+            substitute_conditions_with_axioms(task)
     split_disjunctions(task)
     move_existential_quantifiers(task)
     eliminate_existential_quantifiers_from_axioms(task)
@@ -463,9 +463,9 @@ def condition_to_rule_body(parameters: Sequence[pddl.TypedObject],
 
 if __name__ == "__main__":
     from translate import pddl_parser
-    from translate.options import get_options, set_options
+    from translate.options import set_options
 
     set_options() # use command line options
     task = pddl_parser.open()
-    normalize(task, get_options().condition_normalization_strategy)
+    normalize(task)
     task.dump()
--- a/src/translate/normalize.py
+++ b/src/translate/normalize.py
@@ -374,12 +374,12 @@ def substitute_complicated_goal(task):
 
 def normalize(task):
     remove_universal_quantifiers(task)
-    match get_options().condition_normalization_strategy:
-        case "dnf":
-            substitute_complicated_goal(task)
-            build_DNF(task)
-        case "axiom_based":
-            substitute_conditions_with_axioms(task)
+    condition_strategy = get_options().condition_normalization_strategy
+    if condition_strategy == "dnf":
+        substitute_complicated_goal(task)
+        build_DNF(task)
+    elif condition_strategy == "axiom_based":
+        substitute_conditions_with_axioms(task)
     split_disjunctions(task)
     move_existential_quantifiers(task)
     eliminate_existential_quantifiers_from_axioms(task)
--- a/src/translate/normalize.py
+++ b/src/translate/normalize.py
@@ -226,6 +226,36 @@ def build_DNF(task):
         if proxy.condition.has_disjunction():
             proxy.set(recurse(proxy.condition).simplified())
 
+# [2 Alternative] Decompose disjunctive formulas using derived predicates.
+# Replace each disjunctive subformula with a derived predicate and add a
+# corresponding axiom defining that predicate. The transformation proceeds
+# bottom-up so nested disjunctions are eliminated first.
+def substitute_conditions_with_axioms(task):
+    def recurse(condition, type_map):
+        if isinstance(condition, (pddl.Literal, pddl.Truth, pddl.Falsity)):
+            return condition
+        elif isinstance(condition, (pddl.Conjunction, pddl.ExistentialCondition)):
+            new_parts = [recurse(part, type_map) for part in condition.parts]
+            condition = condition.change_parts(new_parts)
+            return condition
+        parameters = sorted(condition.free_variables())
+        typed_parameters = tuple(pddl.TypedObject(v, type_map[v]) for v in parameters)
+        key = (condition, typed_parameters)
+        axiom = new_axioms_by_condition.get(key)
+        if not axiom:
+            new_parts = [recurse(part, type_map) for part in condition.parts]
+            condition = condition.change_parts(new_parts)
+            axiom = task.add_axiom(list(typed_parameters), condition)
+            new_axioms_by_condition[key] = axiom
+        return pddl.Atom(axiom.name, parameters)
+
+    new_axioms_by_condition = {}
+
+    for proxy in tuple(all_conditions(task)):
+        # Cannot use generator because we add new axioms on the fly.
+        type_map = proxy.get_type_map()
+        proxy.set(recurse(proxy.condition, type_map))
+
 # [3] Split conditions at the outermost disjunction.
 def split_disjunctions(task):
     for proxy in tuple(all_conditions(task)):
@@ -340,10 +370,16 @@ def substitute_complicated_goal(task):
 # Combine Steps [1], [2], [3], [4], [5] and do some additional verification
 # that the task makes sense.
 
-def normalize(task):
+
+def normalize(task, condition_strategy):
     remove_universal_quantifiers(task)
-    substitute_complicated_goal(task)
-    build_DNF(task)
+    if condition_strategy == "dnf":
+        substitute_complicated_goal(task)
+        build_DNF(task)
+    elif condition_strategy == "axiom_based":
+        substitute_conditions_with_axioms(task)
+    else:
+        raise ValueError("Unknown condition normalization strategy: %s" % condition_strategy)
     split_disjunctions(task)
     move_existential_quantifiers(task)
     eliminate_existential_quantifiers_from_axioms(task)
@@ -427,9 +463,9 @@ def condition_to_rule_body(parameters: Sequence[pddl.TypedObject],
 
 if __name__ == "__main__":
     from translate import pddl_parser
-    from translate.options import set_options
+    from translate.options import get_options, set_options
 
     set_options() # use command line options
     task = pddl_parser.open()
-    normalize(task)
+    normalize(task, get_options().condition_normalization_strategy)
     task.dump()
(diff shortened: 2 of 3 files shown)


src/translate/normalize.py
```diff
@@ -427,9 +463,9 @@ def condition_to_rule_body(parameters: Sequence[pddl.TypedObject],
 
 if __name__ == "__main__":
     from translate import pddl_parser
-    from translate.options import set_options
+    from translate.options import get_options, set_options
 
     set_options() # use command line options
     task = pddl_parser.open()
-    normalize(task)
+    normalize(task, get_options().condition_normalization_strategy)
```
roeger: Is there a reason why you make this an explicit parameter of `normalize` instead of using the options there? (see comments there)

src/translate/normalize.py
```diff
@@ -340,10 +370,16 @@ def substitute_complicated_goal(task):
...
     remove_universal_quantifiers(task)
-    substitute_complicated_goal(task)
-    build_DNF(task)
+    if condition_strategy == "dnf":
+        substitute_complicated_goal(task)
+        build_DNF(task)
+    elif condition_strategy == "axiom_based":
+        substitute_conditions_with_axioms(task)
+    else:
+        raise ValueError("Unknown condition normalization strategy: %s" % condition_strategy)
```
roeger: I don't think there should be a string parameter of normalize. Instead add
```from translate.options import get_options```
at the top of the file and use
```if get_options().condition_normalization_strategy == ...```
here. Or maybe even clearer with a `match` statement instead of the if ... elif ... else.
```
condition_strategy = get_options().condition_normalization_strategy
match condition_strategy:
    case "dnf":
        ...
    case "axiom_based":
        ...
    case _:
        raise ValueError(f"Unknown condition normalization strategy: {condition_strategy}")
```

I am also not sure whether we need the last case because the code would never reach it (argparse fixes the set of values). If you want to keep it, we could discuss whether it really should be a `ValueError`.

File: src/translate/normalize.py [hunk 3]
```diff
@@ -226,6 +226,36 @@ def recurse(condition):
...
+            new_axioms_by_condition[key] = axiom
+        return pddl.Atom(axiom.name, parameters)
+
+    new_axioms_by_condition = {}
+
+    for proxy in tuple(all_conditions(task)):
+        # Cannot use generator because we add new axioms on the fly.
+        type_map = proxy.get_type_map()
+        proxy.set(recurse(proxy.condition, type_map))
+
```
roeger, line 315: This looks correct. Have you also considered other variants of avoiding "duplicate" axioms? I wonder whether using canonical parameter names might be helpful.

roeger: Thinking more about this, I think canonical parameter names might be too expensive because we would have to propagate them down through the entire condition (again and again).

src/translate/normalize.py
speckdavid: Good point. I fixed it.

src/translate/normalize.py
speckdavid: I adjusted it accordingly.

src/translate/normalize.py
speckdavid: I thought about it for a bit, but it wasn't obvious how to do it efficiently. I also wondered whether we could encounter formulas like `(a OR b OR a)`, which would result in different axioms than `(a OR b)`. Then I realized that it would probably make more sense to think about simplifying lifted formulas already at this stage of the translator. So, I decided to not do smarter things for now.

src/translate/normalize.py
speckdavid: I tried using match, but it failed on our Windows tests, which was actually very good. Apparently, `match`  is only supported starting with `Python 3.10`. So I switched back to using if/else.

jendrikseipp: Python 3.9 is EOL, so I think we could just bump the Python version for Windows to 3.14, same as for macOS.

roeger: Maybe we should switch to a newer windows runner (2022 is not quite yesterday...).

FlorianPommerening: We tried going to a newer Windows build before and failed to run our tests there:
https://github.com/aibasel/downward/blob/main/.github/workflows/windows.yml#L29-L36

For the compiler it didn't matter back then, the "old" windows runner still used the latest compiler version. I don't know about the Python version, though.

src/translate/normalize.py
```diff
@@ -226,6 +228,36 @@ def recurse(condition):
         if proxy.condition.has_disjunction():
             proxy.set(recurse(proxy.condition).simplified())
 
+# [2 Alternative] Decompose disjunctive formulas using derived predicates.
+# Replace each disjunctive subformula with a derived predicate and add a
+# corresponding axiom defining that predicate. The transformation proceeds
+# bottom-up so nested disjunctions are eliminated first.
+def substitute_conditions_with_axioms(task):
```
SimonDold: is this the right function name?
Do you substitute conditions with axioms?
Don't you substitute them with derived predicates (and introduce the corresponding axioms for them)?

And the way i got it, it is not for all conditions, only disjunctions. So maybe call the function something like `eliminate_disjunctive_conditions`.

About the comment: Is it really only replacing disjunctive SUBformulas? Will a condition that is a disjunction of conjunctions remain as it has no disjunctive subformulas, only the outer most formula is a disjunction?

Do we need to call [3] `split_disjunctions` after we called `substitute_conditions_with_axioms`, i.e.
are the still disjunctions left that we have to split?

src/translate/normalize.py
SimonDold: This comment seems outdated and the change makes it more apparent.

What `normalize` used to do was 
[1]
substitute_complicated_goal
[2]
[3]
[4]
[5a]
[5b]
[5c]
verify_axiom_predicates

now it does either that or
[1]
[2ALTERNATIVE]
[3]
[4]
[5a]
[5b]
[5c]
verify_axiom_predicates

maybe this would be cleaner by implementing it this way
```
# Combine Steps [1], [2], [3], [4], [5a] [5b] [5c] and do some additional verification
# that the task makes sense.
def normalize(task)
    remove_universal_quantifiers(task)
    normalize_conditions(task) # this does the options check internally
    split_disjunctions(task)
    move_existential_quantifiers(task)
    eliminate_existential_quantifiers_from_axioms(task)
    eliminate_existential_quantifiers_from_preconditions(task)
    eliminate_existential_quantifiers_from_conditional_effects(task)

    verify_axiom_predicates(task)
```
and have `normalize_conditions(task)` as the new [2]
```
# [2] normalize the conditions of the task according to the chosen strategy
def normalize_conditions(task):
    condition_strategy = get_options().condition_normalization_strategy
    if condition_strategy == "dnf":
        substitute_complicated_goal(task)
        build_DNF(task)
    elif condition_strategy == "axiom_based":
        substitute_conditions_with_axioms(task)
```

In my opinion that would keep more of the beauty the that the implementation of `normalize` had.

src/translate/normalize.py
speckdavid: Thanks. I have renamed the function and also the option itself with a more speaking name. 

[3] split_disjuncitons is still needed because it splits the axioms with disjunctive bodies of literals ("base case"). This could probably also be done in place but it would make the code more complex. So, I decided to leave it as it is for now.

src/translate/normalize.py
speckdavid: Thanks. This does indeed make it cleaner. I also extracted steps 5a–5c into a separate function, so normalize now only calls the high-level steps.

src/translate/normalize.py
SimonDold: Well, if the new function does leave some disjunctive conditions that `split_disjunctions` has to split, then I do not recommend to call the function `eliminate_disjunctive_conditions`.
This name communicates to me that it would do more than what it actually does.

src/translate/normalize.py
SimonDold: Looks great! I think the extracted steps 5a-5c are a nice drive-by clean-up here :)

src/translate/normalize.py
speckdavid: We gave the function a meaningful name and decided to only consider axioms for splitting afterwards.

src/translate/normalize.py
speckdavid: It is now cleaned up even further to only have 3 meaningfull steps.

src/translate/main.py
```diff
@@ -695,7 +695,7 @@ def main():
         task = pddl_parser.open(
             domain_filename=get_options().domain, task_filename=get_options().task)
 
-    with timers.timing("Normalizing task"):
+    with timers.timing(f"Normalizing task with strategy {get_options().condition_normalization_strategy}"):
```
roeger: Did you consider the effects of this on lab?
It will lead to two separate attributes
`translator_time_normalizing_task_with_strategy_axiom_based`
and `translator_time_normalizing_task_with_strategy_dnf`.

maltehelmert: Indeed, I'd strongly recommend not touching the `with timers.timing(...)` for this reason.

src/translate/normalize.py
```diff
@@ -175,8 +181,28 @@ def recurse(condition):
             type_map = proxy.get_type_map()
             proxy.set(recurse(proxy.condition))
 
+# [2] Normalize conditions according to the selected strategy.
+def eliminate_disjunctive_conditions(task):
+    condition_strategy = get_options().condition_normalization_strategy
+    if condition_strategy == "dnf":
+        substitute_complicated_goal(task)
```
roeger: I am not convinced that this should be encapsulated in something called `eliminate_disjunctive_conditions`. It's not only about this. For example, it would also replace an existential condition with a (derived) atom.

Your strategy would not transform such a goal and it seems to be handled fine by a later stage of the normalization. But in this case, we probably also should restrict `substitute_complicated_goal` to disjunctive formulas. Are there other corner cases of goal formulas that need to be considered?

src/translate/main.py
speckdavid: We reverted the changed output.

src/translate/normalize.py
speckdavid: Very good point. 

We (@SimonDold  and I) looked into it again and realized that the existential quantifiers were not handled as expected. In fact, they are also a form of "or" and later in the pipeline pulled out where the variables then are used as parameters. This process is not in the spirit of the axiom_based approach. Furthermore, this is not possible for the goal.

So, we changed the method to not only move disjunctions but also existential quantifiers into axioms.

Accordingly, we renamed `eliminate_disjunctive_conditions` to `simplyfy_conditions` because after the process only "simple conditions", i.e., conjuncitons and existential conditions, remain (+ truth values, literals).

maltehelmert: Top-level existential quantifiers of operators and axioms behave the same way semantically as parameters. For axioms, they have exactly the same semantics; for operators, the only difference is whether the parameters are part of the action name or not. Is there a good argument for compiling them into axioms? It potentially means that grounding cannot be as efficient any more because we have a smaller space of choices for the join tree decomposition, and we get axiom-based tasks in what could be a STRIPS/SAS+ tasks.

(For the goal it's different because the goal cannot have parameters.)

Perhaps something we should discuss live?

src/translate/normalize.py
```diff
@@ -226,18 +272,54 @@ def recurse(condition):
...
-            for part in proxy.condition.parts:
-                new_proxy = proxy.clone_owner()
-                new_proxy.set(part)
-                new_proxy.register_owner(task)
-            proxy.delete_owner(task)
+# [2 Option 2] Decompose disjunctive and existential formulas using derived
+# predicates. Replace each disjunctive (sub)formula with a derived predicate
+# and add a corresponding axiom defining that predicate. The transformation
+# proceeds bottom-up so nested disjunctions are eliminated first.
+def substitute_complicated_conditions(task, condition_strategy):
```
roeger: The `condition_strategy` should not be handed over as a string (it is no longer tied to the options object here). Instead use a Boolean `axiomatize_existential` (or similar) that defines whether existential conditions should also be replaced with derived predicates.

src/translate/normalize.py
```diff
@@ -226,18 +272,54 @@ def recurse(condition):
...
-        if isinstance(proxy.condition, pddl.Disjunction):
-            for part in proxy.condition.parts:
-                new_proxy = proxy.clone_owner()
-                new_proxy.set(part)
-                new_proxy.register_owner(task)
-            proxy.delete_owner(task)
+# [2 Option 2] Decompose disjunctive and existential formulas using derived
+# predicates. Replace each disjunctive (sub)formula with a derived predicate
+# and add a corresponding axiom defining that predicate. The transformation
+# proceeds bottom-up so nested disjunctions are eliminated first.
```
roeger: The comment only properly reflects the disjunctions and is unclear about the existential conditions.

src/translate/normalize.py
```diff
@@ -226,18 +272,54 @@ def recurse(condition):
...
+        elif not isinstance(condition, replace_types):
+            new_parts = [recurse(part, type_map, replace_types) for part in condition.parts]
+            condition = condition.change_parts(new_parts)
+            return condition
+        assert isinstance(condition, replace_types)
+        parameters = sorted(condition.free_variables())
+        typed_parameters = tuple(pddl.TypedObject(v, type_map[v]) for v in parameters)
+        key = (condition, typed_parameters)
+        axiom = new_axioms_by_condition.get(key)
+        if not axiom:
```
roeger, line 293: More out of curiosity: have you run experiments to get a sense of how often the caching helps and how often it doesn't?

src/translate/normalize.py
```diff
@@ -175,8 +180,49 @@ def recurse(condition):
...
+        substitute_complicated_goal(task)
+
+    if condition_strategy == "dnf":
+        build_DNF(task)
+    elif condition_strategy in ("axiomatize_disjunctions", "axiomatize_disjunctions_existentials"):
+        substitute_complicated_conditions(task, condition_strategy)
+    else:
+        raise SystemExit("error: unknown condition normalization strategy: %s" % condition_strategy)
+
+    split_disjunctions(task)
```
SimonDold, line 200: For me it would be easier to read if it is split up into the different modes.
But i am also fine as is.

```
def simplify_conditions(task):
    condition_strategy = get_options().condition_normalization_strategy
    if condition_strategy == "dnf":
        substitute_complicated_goal(task)
        build_DNF(task)
    elif condition_strategy == "axiomatize_disjunctions":
        substitute_complicated_goal(task)
        substitute_complicated_conditions(task, condition_strategy)
    elif condition_strategy == "axiomatize_disjunctions_existentials":
        substitute_complicated_conditions(task, condition_strategy)
    else:
        raise SystemExit("error: unknown condition normalization strategy: %s" % condition_strategy)
    split_disjunctions(task)
```

speckdavid: Done. It is now even clearer with Gabis suggestion about the boolean flag for `substitute_complicated_conditions` .

src/translate/normalize.py
speckdavid: I checked it locally for a handcrafted instance with a lot of repetition. There it did work. For the smaller instances in the standard benchmark (which I tested locally), it did not really find duplicates. However, this may changes for larger or "worse" modeled problems. Definitely something we can run an experiment for curiosity :)

src/translate/normalize.py
speckdavid: Thanks, fixed.

src/translate/normalize.py
speckdavid: Thanks, fixed.

PR #299, opened 2026-07-27
[Issue 1224] Introduce one-sided constraints to the LP interface.

gutodelazeri:
This PR changes the LP interface from two-sided constraints

    lb <= a^Tx <= ub

to one-sided constraints and equalities:

    a^T x <= rhs
    a^T x >= rhs
    a^T x  = rhs

`LPConstraint` now stores a `Sense` (`LE`, `GE`, or `EQ`) and one right-hand side.

This simplifies LP backends. It is also preparatory work for adding further LP solvers in issue 1199 (https://issues.fast-downward.org/issue1199).

## Main changes

- Replace `LPConstraint`'s lower and upper bounds with a constraint sense and right-hand side.
- Replace `set_constraint_lower_bound()` and `set_constraint_upper_bound()` with `set_constraint_rhs()` and `set_constraint_sense()`.
- Update the CPLEX backend:
  - map constraint senses directly to CPLEX senses;
  - remove ranged-row data and conversion logic;
  - remove bookkeeping that was only needed for inconsistent two-sided constraints.
- Update the SoPlex backend:
  - convert each sense/RHS pair to SoPlex's two-sided representation;
  - remember constraint senses so dynamically changing the RHS updates the correct side;
- Update all existing LP constraint generators to the new interface.

## Important notes

The old interface contained two places where two-sided constraints were created, but one side was redundant:
- In landmark cost partitioning, the lower bound was implied by nonnegative landmark-cost variables.
- In the Rankooh-Rintanen formulation, the upper bound of constraint family (3) was implied by the bounds on `f_q` and `f_{p,a}`. In the paper, the authors do not mention an upper bound. 

The new interface does not add these redundant constraints. The generated LP models, however, are equivalent, i.e., their feasible region and objective function remain unchanged. 

## Validation

We compared the new interface with baseline Fast Downward on the full optimal benchmark suite using CPLEX and SoPlex across 11 solver/configuration combinations.

Overall:
- coverage increased or remained the same for most configurations;
- memory usage generally decreased;
- every task solved by both revisions had identical expansions, evaluations, plan cost, and plan length.

We separately tested restoring the redundant Rankooh-Rintanen upper bound as an additional one-sided constraint. This did not recover the behavior of baseline Fast Downward and slightly reduced coverage, so the redundant constraint is not included in this PR.

Experiment details and discussion:
https://issues.fast-downward.org/issue1224


src/search/lp/lp_solver.h
```diff
@@ -31,13 +31,21 @@ std::tuple<LPSolverType> get_lp_solver_arguments_from_options(
 
 class LinearProgram;
 
+enum class Sense {
+    GE, // ax >= b
+    LE, // ax <= b
+    EQ // ax = b
```
SimonDold: I think those should be spelled out. `GREATER_EQUAL, LESS_EQUAL, EQUAL`.

src/search/lp/lp_solver.h
```diff
@@ -46,17 +54,20 @@ class LPConstraint {
...
+    Sense get_sense() const {
+        return sense;
     }
-    double get_upper_bound() const {
-        return upper_bound;
+
+    void set_right_hand_side(double rhs) {
+        right_hand_side = rhs;
     }
-    void set_upper_bound(double ub) {
```
SimonDold, line 58: I think the interface with `set_upper_bound` and `set_lower_bound` is beneficial, because it allows me to not think about the way you write down your constraints.
Is it `x+y>=1` or `1<=x+y`, they are the same semantics.

if i have the expression `x+y` I would enjoy to just state a semantic wish "please restrict this with an upper bound of 1", and you take care of writing it down.

I don't think it is very useful to modify the sense or the rhs on its own. When would you ever do this?

src/search/lp/cplex_solver_interface.cc
```diff
@@ -67,30 +67,32 @@ static void freeProblem(CPXENVptr env, CPXLPptr *problem) {
...
-static int sense_to_cplex_sense(LPObjectiveSense sense) {
+static int model_sense_to_cplex_sense(LPObjectiveSense sense) {
     if (sense == LPObjectiveSense::MINIMIZE) {
         return CPX_MIN;
     } else {
         return CPX_MAX;
     }
 }
 
+static char constraint_sense_to_cplex_sense(lp::Sense constraint_sense) {
```
FlorianPommerening: Do we need `lp::` here? We should be inside the scope of the namespace already.

FlorianPommerening: Also, to be consistent with `LPObjectiveSense`, maybe we should rename `Sense` to `LPConstraintSense` or `ConstraintSense` and the function `model_sense_to_cplex_sense` to `objective_sense_to_cplex_sense`.

src/search/lp/cplex_solver_interface.cc
```diff
@@ -67,30 +67,32 @@ static void freeProblem(CPXENVptr env, CPXLPptr *problem) {
...
+    case lp::Sense::GE:
+        return 'G';
+        break;
+    case lp::Sense::EQ:
+        return 'E';
+        break;
+    default:
+        cerr << "Unsupported constraint sense: "
+             << static_cast<int>(constraint_sense) << endl;
+        utils::exit_with(utils::ExitCode::SEARCH_CRITICAL_ERROR);
```
FlorianPommerening:
```suggestion
        ABORT("Unsupported constraint sense: " + 
            static_cast<int>(constraint_sense));
```

We usually use `ABORT` for "this should not be possible", i.e., errors targeted to developers rather than users.

src/search/lp/lp_solver.h
FlorianPommerening: We discussed this in person and decided to keep the functions as they are. The main reasons are
* We do not have an expression class, so we cannot represent `x+y` by itself as in the example above. Without a class representing expressions, and only a class representing a one-sided constraint, the functions imply a functionality that we do not support.
* with two set-bound functions, it is possible to create a trivially unsolvable constraint `1 <= EXPRESSION <= 0` which cannot be represented in CPLEX. This pull request gets rid of the complexity that was necessary to deal with this situation and having both set-bounds functions would re-introduce it.

src/search/lp/soplex_solver_interface.cc
```diff
@@ -23,7 +23,6 @@ static LPRowSetReal constraints_to_row_set(
     for (const LPConstraint &constraint : constraints) {
         num_nonzeros += constraint.get_coefficients().size();
     }
-
```
FlorianPommerening, line 26: Was this intentional? I liked the newline there to separate the block dealing with nonzeros from the block dealing with constraints. If you prefer otherwise, that is also fine with me.

src/search/lp/soplex_solver_interface.cc
```diff
@@ -33,9 +32,28 @@ static LPRowSetReal constraints_to_row_set(
         for (int i = 0; i < num_entries; ++i) {
             entries.add(variables[i], coefficients[i]);
         }
-        rows.add(
-            constraint.get_lower_bound(), entries,
-            constraint.get_upper_bound());
+        const double b = constraint.get_right_hand_side();
+        const lp::Sense s = constraint.get_sense();
```
FlorianPommerening:
```suggestion
        const double rhs = constraint.get_right_hand_side();
        const lp::Sense sense = constraint.get_sense();
```

src/search/lp/soplex_solver_interface.cc
```diff
@@ -33,9 +32,28 @@ static LPRowSetReal constraints_to_row_set(
...
+            rhs = soplex::infinity;
+            break;
+        case lp::Sense::EQ:
+            lhs = b;
+            rhs = b;
+            break;
+        default:
+            cerr << "Invalid constraint sense code: " << static_cast<int>(s)
+                 << endl;
+            utils::exit_with(utils::ExitCode::SEARCH_CRITICAL_ERROR);
```
FlorianPommerening:
```suggestion
            ABORT("Invalid constraint sense code: " + static_cast<int>(s));
```

src/search/lp/soplex_solver_interface.cc
```diff
@@ -71,19 +89,28 @@ void SoPlexSolverInterface::load_problem(const LinearProgram &lp) {
...
+        constraint_senses.push_back(constraint.get_sense());
+    }
 }
 
 void SoPlexSolverInterface::add_temporary_constraints(
     const named_vector::NamedVector<LPConstraint> &constraints) {
     soplex.addRowsReal(constraints_to_row_set(constraints));
-    num_temporary_constraints = constraints.size();
+    num_temporary_constraints += constraints.size();
+    for (int i = 0; i < (int)constraints.size(); ++i) {
```
FlorianPommerening:
```suggestion
    int num_constraints = constraints.size();
    for (int i = 0; i < num_constraints; ++i) {
```

`static_cast<int>(x)` would be better than the C-style cast `(int) x` but we often use a local variable for this.

src/search/lp/soplex_solver_interface.cc
```diff
@@ -71,19 +89,28 @@ void SoPlexSolverInterface::load_problem(const LinearProgram &lp) {
...
+        const auto &c = constraints[i];
```
FlorianPommerening:
```suggestion
        const LPConstraint &c = constraints[i];
```

src/search/lp/soplex_solver_interface.cc
```diff
@@ -240,4 +257,63 @@ bool SoPlexSolverInterface::has_temporary_constraints() const {
 void SoPlexSolverInterface::print_statistics() const {
     soplex.printStatistics(cout);
 }
+
+void SoPlexSolverInterface::set_constraint_rhs(int index, double b) {
+    const lp::Sense sense = constraint_senses[index];
+
+    if (sense == lp::Sense::GE) {
```
FlorianPommerening: could be a switch-case to be more consistent with other uses.

src/search/lp/soplex_solver_interface.cc
```diff
@@ -240,4 +257,63 @@ bool SoPlexSolverInterface::has_temporary_constraints() const {
...
+        soplex.changeLhsReal(index, b);
+    } else if (sense == lp::Sense::LE) {
+        soplex.changeRhsReal(index, b);
+    } else if (sense == lp::Sense::EQ) {
+        soplex.changeLhsReal(index, b);
+        soplex.changeRhsReal(index, b);
+    } else {
+        cerr << "Invalid constraint sense code: " << static_cast<int>(sense)
```
FlorianPommerening: `ABORT`, see above

src/search/lp/soplex_solver_interface.cc
```diff
@@ -240,4 +257,63 @@ bool SoPlexSolverInterface::has_temporary_constraints() const {
...
+             << endl;
+        utils::exit_with(utils::ExitCode::SEARCH_CRITICAL_ERROR);
+    }
+}
+
+void SoPlexSolverInterface::set_constraint_sense(int index, lp::Sense sense) {
+    const double lhs = soplex.lhsReal(index);
+    const double rhs = soplex.rhsReal(index);
+
+    const bool lhs_is_neginf = (lhs == -infinity);
```
FlorianPommerening: You used `soplex::infinity` earlier. If there is a local variable, then probably cases that use `soplex::infinity` should use the local variable instead.

src/search/lp/soplex_solver_interface.cc
```diff
@@ -240,4 +257,63 @@ bool SoPlexSolverInterface::has_temporary_constraints() const {
...
+
+    double b;
+    if (lhs_is_neginf && !rhs_is_posinf) {
+        b = rhs;
+    } else if (!lhs_is_neginf && rhs_is_posinf) {
+        b = lhs;
+    } else if (!lhs_is_neginf && !rhs_is_posinf && lhs == rhs) {
+        b = rhs;
+    } else {
+        b = rhs; // keep compiler happy
```
FlorianPommerening:
```suggestion
        utils::unused_variable(b)
```
Needs an import of `../utils/language.h` (I think)

src/search/lp/soplex_solver_interface.cc
```diff
@@ -240,4 +257,63 @@ bool SoPlexSolverInterface::has_temporary_constraints() const {
...
+        utils::exit_with(utils::ExitCode::SEARCH_CRITICAL_ERROR);
+    }
+}
+
+void SoPlexSolverInterface::set_constraint_sense(int index, lp::Sense sense) {
+    const double lhs = soplex.lhsReal(index);
+    const double rhs = soplex.rhsReal(index);
+
+    const bool lhs_is_neginf = (lhs == -infinity);
+    const bool rhs_is_posinf = (rhs == infinity);
```
FlorianPommerening: I'd avoid the abbreviation `posinf` and `neginf` in the variable name. Maybe `lhs_unbounded`/`rhs_unbounded`? Maybe the logic below also gets clearer if we flip this and say `lhs_finite = (lhs > - infinity)` instead.

src/search/lp/soplex_solver_interface.cc
```diff
@@ -240,4 +257,63 @@ bool SoPlexSolverInterface::has_temporary_constraints() const {
...
+    case lp::Sense::GE:
+        soplex.changeLhsReal(index, b);
+        soplex.changeRhsReal(index, infinity);
+        break;
+    case lp::Sense::EQ:
+        soplex.changeLhsReal(index, b);
+        soplex.changeRhsReal(index, b);
+        break;
+    default:
+        cerr << "Invalid constraint sense code: " << static_cast<int>(sense)
```
FlorianPommerening: `ABORT`

src/search/operator_counting/delete_relaxation_rr_constraints.cc
```diff
@@ -543,22 +538,23 @@ bool DeleteRelaxationRRConstraints::update_constraints(
     int con_id;
     for (FactPair f : last_state) {
         con_id = get_constraint_id(f);
-        lp_solver.set_constraint_lower_bound(con_id, 0);
-        lp_solver.set_constraint_upper_bound(con_id, 0);
+        lp_solver.set_constraint_sense(con_id, lp::Sense::EQ);
```
FlorianPommerening: Do we have to set the sense here (and in line 548)? If all these constraints start as equality constraints and are never changed, then not changing the sense here might be ok and more efficient.

src/search/lp/lp_solver.cc
```diff
@@ -137,6 +139,7 @@ LPSolver::LPSolver(LPSolverType solver_type) {
         missing_solver = "SoPlex";
 #endif
         break;
+
```
FlorianPommerening: There is no newline before the other cases. I'd not add one here.

src/search/lp/lp_solver.cc
```diff
@@ -69,11 +75,7 @@ ostream &LPConstraint::dump(
         }
         stream << coefficients[i] << " * " << variable_name;
     }
-    if (upper_bound != infinity) {
-        stream << " <= " << upper_bound;
-    } else if (lower_bound == -infinity) {
-        stream << " <= infinity";
-    }
+    stream << get_sense() << get_right_hand_side();
```
FlorianPommerening: I think `get_right_hand_side()` could still return +/- infinity as a solver-internal constant, which would need to be translated to strings `infinity` and `-infinity`.

src/search/lp/cplex_solver_interface.cc
```diff
@@ -204,53 +204,31 @@ void CplexSolverInterface::CplexColumnsInfo::assign(
...
     sense.clear();
-    range_values.clear();
-    range_indices.clear();
-
     int num_rows = constraints.size();
-    sense.resize(num_rows);
-    rhs.resize(num_rows);
-    if (dense_range_values) {
-        range_values.resize(num_rows, 0);
+    if (num_rows > 0) {
```
FlorianPommerening: Does this `if` make a difference here?

src/search/lp/cplex_solver_interface.cc
```diff
@@ -204,53 +204,31 @@ void CplexSolverInterface::CplexColumnsInfo::assign(
...
     for (int row_index = 0; row_index < num_rows; ++row_index) {
         const LPConstraint &constraint = constraints[row_index];
-        double lb = constraint.get_lower_bound();
-        double ub = constraint.get_upper_bound();
-        const auto &[sense_value, rhs_value, range_value] =
-            bounds_to_sense_rhs_range(lb, ub);
-        sense[row_index] = sense_value;
+        double rhs_value = constraint.get_right_hand_side();
+        lp::Sense constraint_sense = constraint.get_sense();
+        char cplex_sense = constraint_sense_to_cplex_sense(constraint_sense);
```
FlorianPommerening: if the variable is only needed in the next line, I'd directly assign to the array without a temporary variable.

FlorianPommerening (commented):
Looks good to me. I left comments but mostly they are about superficial things (renaming variables, etc.) that do not require re-running experiments. The only comment that could affect performance is about removing a `set_sense` in the delete-relaxation constraints.

src/search/lp/soplex_solver_interface.h
```diff
@@ -27,6 +27,7 @@ class SoPlexSolverInterface : public SolverInterface {
     mutable soplex::SoPlex soplex;
     int num_permanent_constraints;
     int num_temporary_constraints;
+    std::vector<lp::LPConstraintSense> constraint_senses;
```
FlorianPommerening: Does this need `lp::`?

src/search/lp/soplex_solver_interface.h
```diff
@@ -40,8 +41,9 @@ class SoPlexSolverInterface : public SolverInterface {
         const std::vector<double> &coefficients) override;
     virtual void set_objective_coefficient(
         int index, double coefficient) override;
-    virtual void set_constraint_lower_bound(int index, double bound) override;
-    virtual void set_constraint_upper_bound(int index, double bound) override;
+    virtual void set_constraint_rhs(int index, double right_hand_side) override;
+    virtual void set_constraint_sense(
+        int index, lp::LPConstraintSense sense) override;
```
FlorianPommerening: Does this need `lp::`?

src/search/lp/soplex_solver_interface.cc
```diff
@@ -71,19 +88,29 @@ void SoPlexSolverInterface::load_problem(const LinearProgram &lp) {
...
+        constraint_senses.push_back(constraint.get_sense());
+    }
 }
 
 void SoPlexSolverInterface::add_temporary_constraints(
     const named_vector::NamedVector<LPConstraint> &constraints) {
     soplex.addRowsReal(constraints_to_row_set(constraints));
-    num_temporary_constraints = constraints.size();
+    num_temporary_constraints += constraints.size();
+    int num_constraints = constraints.size();
```
FlorianPommerening:
```suggestion
    int num_constraints = constraints.size();
    num_temporary_constraints += num_constraints;
```

src/search/lp/soplex_solver_interface.cc
```diff
@@ -33,9 +33,26 @@ static LPRowSetReal constraints_to_row_set(
         for (int i = 0; i < num_entries; ++i) {
             entries.add(variables[i], coefficients[i]);
         }
-        rows.add(
-            constraint.get_lower_bound(), entries,
-            constraint.get_upper_bound());
+        const double lp_rhs = constraint.get_right_hand_side();
```
FlorianPommerening: This used to be `b` and I said to change it to `rhs` but I didn't see that we already have a `rhs` variable in this context. In this case the original name of `b` is probably less confusing than `lp_rhs`

src/search/lp/cplex_solver_interface.h
```diff
@@ -261,8 +208,9 @@ class CplexSolverInterface : public SolverInterface {
         const std::vector<double> &coefficients) override;
     virtual void set_objective_coefficient(
         int index, double coefficient) override;
-    virtual void set_constraint_lower_bound(int index, double bound) override;
-    virtual void set_constraint_upper_bound(int index, double bound) override;
+    virtual void set_constraint_rhs(int index, double right_hand_side) override;
+    virtual void set_constraint_sense(
+        int index, lp::LPConstraintSense sense) override;
```
FlorianPommerening: Do we need `'lp::`?

FlorianPommerening (commented):
found a coulpe more small things

FlorianPommerening (approved):
Looks good to me

PR #300, opened 2026-07-27
Issue1232: Call translator problem file input "problem", not "task".

roeger:
We split this up from #298. The changes have already been reviewed there.

### DIFF
--- a/src/translate/options.py
+++ b/src/translate/options.py
@@ -31,9 +31,9 @@ def get_arg_parser():
     # process (issue1217).
     argparser = argparse.ArgumentParser(prog=infer_prog())
     argparser.add_argument(
-        "domain", help="path to domain pddl file")
+        "domain", metavar="DOMAIN", help="path to domain pddl file")
     argparser.add_argument(
-        "task", help="path to task pddl file")
+        "task", metavar="TASK", help="path to task pddl file")
     argparser.add_argument(
         "--relaxed", dest="generate_relaxed_task", action="store_true",
         help="output relaxed task (no delete effects)")
--- a/src/translate/options.py
+++ b/src/translate/options.py
@@ -33,7 +33,7 @@ def get_arg_parser():
     argparser.add_argument(
         "domain", metavar="DOMAIN", help="path to domain pddl file")
     argparser.add_argument(
-        "task", metavar="TASK", help="path to task pddl file")
+        "problem", metavar="PROBLEM", help="path to problem pddl file")
     argparser.add_argument(
         "--relaxed", dest="generate_relaxed_task", action="store_true",
         help="output relaxed task (no delete effects)")
--- a/src/translate/options.py
+++ b/src/translate/options.py
@@ -31,9 +31,9 @@ def get_arg_parser():
     # process (issue1217).
     argparser = argparse.ArgumentParser(prog=infer_prog())
     argparser.add_argument(
-        "domain", metavar="DOMAIN", help="path to domain pddl file")
+        "domain", metavar="DOMAIN", help="path to domain PDDL file")
     argparser.add_argument(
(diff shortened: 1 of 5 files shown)


src/translate/options.py
```diff
@@ -31,9 +31,9 @@ def get_arg_parser():
     # process (issue1217).
     argparser = argparse.ArgumentParser(prog=infer_prog())
     argparser.add_argument(
-        "domain", help="path to domain pddl file")
+        "domain", metavar="DOMAIN", help="path to domain pddl file")
```
maltehelmert: While we're cleaning this argument up, I'd change "pddl" => "PDDL" in the help string here and in the next changed line.

src/translate/options.py
maltehelmert: This was already correctly "problem", but can be fix the inconsistency between `<domain file>` (space) and `<problem_file>` (underscore) while we're looking at the occurrences of task vs. problem? I would probably use the version with the underscore. LLM concurs.

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -619,25 +619,25 @@ def parse_init(context, alist, predicate_dict, term_names):
...
         predicate_dict, functions, actions, axioms = parse_domain_pddl(context, domain_pddl)
-    if not isinstance(task_pddl, list):
-        context.error("Invalid definition of a PDDL task.")
-    task_name, task_domain_name, task_requirements, objects, init, goal, \
-        use_metric = parse_task_pddl(context, task_pddl, type_dict,
+    if not isinstance(problem_pddl, list):
+        context.error("Invalid definition of a PDDL problem.")
+    problem_name, problem_domain_name, problem_requirements, objects, init, goal, \
+        use_metric = parse_problem_pddl(context, problem_pddl, type_dict,
                                      predicate_dict, {c.name for c in constants})
```
maltehelmert: We need to reindent line 632 here because "problem" is longer than "task". More generally, I think there's a tool that will take care of this for us, but I don't remember off-hand what it is.

src/translate/pddl_parser/parsing_functions.py
```diff
@@ -619,25 +619,25 @@ def parse_init(context, alist, predicate_dict, term_names):
...
 
-    if domain_name != task_domain_name:
+    if domain_name != problem_domain_name:
         context.error(f"The domain name specified by the task "
```
maltehelmert: I would say "...by the problem file " rather than "...by the task ". This also matches "by the domain file" two lines down.

src/translate/pddl_parser/parsing_functions.py
maltehelmert: I would change "TASK" to "PROBLEM" in the two lines above (and then also where this is used).

maltehelmert: Task => Problem

src/translate/options.py
grucla: Done.

src/translate/pddl_parser/parsing_functions.py
grucla: Done (manually for now).

src/translate/pddl_parser/parsing_functions.py
grucla: I agree that it should be "...by the problem file " and changed it to this.

src/translate/pddl_parser/parsing_functions.py
grucla: Done.

grucla: Done.

src/translate/options.py
grucla: I changed `<domain file>` to `<domain_file>`.

PR #301, opened 2026-07-28
[issue1228] alternative precondition choice functions for LM-Cut

PLauerRocks:
This is a pull request for [issue1228](https://issues.fast-downward.org/issue1228).

Open questions:

- [x] Did we reproduce the increase in coverage from Max and my paper?
- [x] Should we implement the [TODO](https://github.com/PLauerRocks/downward-pcf-lmcut/blob/af26add5b5b52b95ff9c9250d2ad117158e54d6d/src/search/operator_counting/lm_cut_constraints.cc#L22) to introduce a flag for operator counting constraints based on LM-Cut.
- [x] In the comments I refer to "goal zone" as in Malte and Carmel's paper. But the functions in the code actually say "goal plateau". Should this be adjusted? If yes, in what way?
- [x] Should the options be enabled by default?

### DIFF
--- a/src/search/heuristics/lm_cut_heuristic.h
+++ b/src/search/heuristics/lm_cut_heuristic.h
@@ -15,7 +15,8 @@ class LandmarkCutHeuristic : public Heuristic {
 public:
     LandmarkCutHeuristic(
         const std::shared_ptr<AbstractTask> &task, bool cache_estimates,
-        const std::string &description, utils::Verbosity verbosity);
+        const std::string &description, utils::Verbosity verbosity,
+        bool use_goal_zone_detection, bool use_border_detection);
 };
 }
 
--- a/src/search/heuristics/lm_cut_heuristic.h
+++ b/src/search/heuristics/lm_cut_heuristic.h
@@ -14,9 +14,9 @@ class LandmarkCutHeuristic : public Heuristic {
     virtual int compute_heuristic(const State &ancestor_state) override;
 public:
     LandmarkCutHeuristic(
-        const std::shared_ptr<AbstractTask> &task, bool cache_estimates,
-        const std::string &description, utils::Verbosity verbosity,
-        bool use_goal_zone_detection, bool use_border_detection);
+        const std::shared_ptr<AbstractTask> &task, bool use_goal_zone_detection,
+        bool use_border_detection, bool cache_estimates,
+        const std::string &description, utils::Verbosity verbosity);
 };
 }
 
--- a/src/search/heuristics/lm_cut_landmarks.cc
+++ b/src/search/heuristics/lm_cut_landmarks.cc
@@ -263,15 +263,12 @@ void LandmarkCutLandmarks::break_supporter_ties(RelaxedOperator *op) {
 }
 
 void LandmarkCutLandmarks::mark_goal_plateau(RelaxedProposition *subgoal) {
-    // NOTE: subgoal can be null if we got here via recursion through
-    // a zero-cost action that is relaxed unreachable. (This can only
-    // happen in domains which have zero-cost actions to start with.)
-/*
-  NOTE: subgoal can be null if we got here via recursion through
-  a zero-cost action that is relaxed unreachable. (This can only
-  happen in domains which have zero-cost actions to start with.)
-  For example, this happens in pegsol-strips #01.
-*/
+    /*
+      NOTE: subgoal can be null if we got here via recursion through
+      a zero-cost action that is relaxed unreachable. (This can only
+      happen in domains which have zero-cost actions to start with.)
+      For example, this happens in pegsol-strips #01.
+    */
     if (subgoal && subgoal->status != GOAL_ZONE) {
         subgoal->status = GOAL_ZONE;
         for (RelaxedOperator *achiever : subgoal->effect_of)
@@ -284,13 +281,11 @@ void LandmarkCutLandmarks::mark_goal_plateau(RelaxedProposition *subgoal) {
 
 void LandmarkCutLandmarks::validate_h_max() const {
 #ifndef NDEBUG
-    // Using conditional compilation to avoid complaints about unused
-    // variables when using NDEBUG. This whole code does nothing useful
-/*
-  Using conditional compilation to avoid complaints about unused
-  variables when using NDEBUG. This whole code does nothing useful
-  when assertions are switched off anyway.
-*/
+    /*
+      Using conditional compilation to avoid complaints about unused
+      variables when using NDEBUG. This whole code does nothing useful
+      when assertions are switched off anyway.
+    */
     for (const RelaxedOperator &op : relaxed_operators) {
         if (op.unsatisfied_preconditions) {
             bool reachable = true;
@@ -321,15 +316,12 @@ bool LandmarkCutLandmarks::compute_landmarks(
     for (RelaxedOperator &op : relaxed_operators) {
         op.cost = op.base_cost;
     }
-    // The following three variables could be declared inside the loop
-    // ("second_exploration_queue" even inside second_exploration),
-    // but having them here saves reallocations and hence provides a
-/*
-  The following three variables could be declared inside the loop
-  ("second_exploration_queue" even inside second_exploration),
-  but having them here saves reallocations and hence provides a
-  measurable speed boost.
-*/
+    /*
+      The following three variables could be declared inside the loop
+      ("second_exploration_queue" even inside second_exploration),
+      but having them here saves reallocations and hence provides a
+      measurable speed boost.
+    */
     vector<RelaxedOperator *> cut;
     Landmark landmark;
     vector<RelaxedProposition *> second_exploration_queue;
--- a/src/search/heuristics/lm_cut_landmarks.cc
+++ b/src/search/heuristics/lm_cut_landmarks.cc
@@ -234,11 +234,7 @@ void LandmarkCutLandmarks::second_exploration(
     }
 }
 
-/*
-  A proposition is on the border of the goal zone if it can not be achieved by
-  operators with zero cost.
-*/
-static bool is_border(const RelaxedProposition *prop) {
+static bool has_no_zero_cost_achiever(const RelaxedProposition *prop) {
     for (const RelaxedOperator *op : prop->effect_of) {
         if (op->cost == 0) {
             return false;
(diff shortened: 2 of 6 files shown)


src/search/heuristics/lm_cut_heuristic.h
```diff
@@ -15,7 +15,8 @@ class LandmarkCutHeuristic : public Heuristic {
 public:
     LandmarkCutHeuristic(
         const std::shared_ptr<AbstractTask> &task, bool cache_estimates,
-        const std::string &description, utils::Verbosity verbosity);
+        const std::string &description, utils::Verbosity verbosity,
+        bool use_goal_zone_detection, bool use_border_detection);
```
SimonDold: minor inconsistency with the rest of the code base 

parameter order for components should be 
1st task,
2nd parameters specific for this component
3rd parameters for this component type

In you case this should be
```LandmarkCutHeuristic(
        const std::shared_ptr<AbstractTask> &task, 
        bool use_goal_zone_detection, bool use_border_detection,
        bool cache_estimates, const std::string &description, utils::Verbosity verbosity);
```

src/search/heuristics/lm_cut_landmarks.cc
```diff
@@ -227,6 +231,35 @@ void LandmarkCutLandmarks::second_exploration(
     }
 }
 
+/*
+  A proposition is on the border of the goal zone if it can not be achieved by
+  operators with zero cost.
+*/
+static bool is_border(const RelaxedProposition *prop) {
```
SimonDold: `is_border` sounds like you answer whether or not the proposition is a/the border.
But the comment says it is answering whether it is *on* the border. So i'd suggest to call the function `is_on_border`.

ClemensBuechner: I'm going one step further and have a question about the comment itself. What's the connection to the goal zone? In the implementation, I see none... It just checks whether all operators achieving `prop` have a non-zero cost or not.
Also, regarding the comment itself: it's not clear to me how `prop` being an effect of only non-zero cost operators prevents it from being in the goal zone. Doesn't the goal zone contain all propositions which may reach goal propositions through only zero-cost operators? So why does it matter how `prop` is reached when the goal zone definition cares about what can be reached from `prop` at zero cost.

src/search/heuristics/lm_cut_heuristic.h
PLauerRocks: This is fixed now, thanks. :)

src/search/heuristics/lm_cut_landmarks.cc
PLauerRocks: I adjusted the function name to `has_no_zero_cost_achiever`, which is probably more suitable to match what the function actually checks. 

Re if this marks a border: All `prop` for which we call `has_no_zero_cost_achiever` from `mark_goal_plateau` are in the goal zone.  The reason is that we call `tie_break` on zero-cost actions while regressing over goal_zone nodes.

The name border comes from the fact that prop is the last node on the regressed path g → ... → prop that is still marked as being in the goal zone.
In Max's and my paper, we adapted the definition so that all incoming edges of `prop` end up in the cut. From that perspective, I still think border is an appropriate name. However, you are right that this is not necessarily true in Fast Downward which follows the cut definition of the original LM-Cut paper. In principle, all actions achieving `prop` could still lie in the goal zone and no edge ends up in the cut. There is actually an example in [the paper](https://fai.cs.uni-saarland.de/lauer/papers/hsdip2020.pdf) where this happens (Figure 1).

Though, since the name now only appears as a command-line option (and corresponding bool flags), I would simply keep it and blame the authors (Max and me). That is the terminology we used in the paper, so keeping the option name consistent seems reasonable to me.

src/search/heuristics/lm_cut_landmarks.cc
```diff
@@ -330,4 +367,27 @@ bool LandmarkCutLandmarks::compute_landmarks(
...
+        "marking the goal zone, break ties in favor of a precondition "
+        "that already belongs to the goal zone.",
+        "true");
+    feature.add_option<bool>(
+        "border_detection",
+        "When choosing the h^max supporter of a zero-cost operator while "
+        "marking the goal zone, break ties in favor of a precondition "
+        "on the border of the goal zone (a proposition reachable only "
+        "through operators with positive cost).",
+        "true");
```
ClemensBuechner, line 390: This is more a meta-comment regarding the concepts behind these two options rather than a review comment. These sound like similar ideas but executed in opposing directions, or do I misunderstand something? On the one hand, goal zone detection prefers to stay inside the goal zone while possible. On the other hand, border detection prefers preconditions at the border, necessarily leaving the goal zone in the next step. I'm surprised these two work so well when combined, even though they follow opposing objectives.

src/search/heuristics/lm_cut_landmarks.cc
```diff
@@ -227,6 +234,34 @@ void LandmarkCutLandmarks::second_exploration(
...
+static bool has_no_zero_cost_achiever(const RelaxedProposition *prop) {
+    for (const RelaxedOperator *op : prop->effect_of) {
+        if (op->cost == 0) {
+            return false;
+        }
+    }
+    return true;
+}
+
+void LandmarkCutLandmarks::tie_break_supporter(RelaxedOperator *op) {
```
ClemensBuechner: Reading this function name, at first glance my mind interprets it as a noun: "the tie-break supporter". If I understand the function semantics correctly, you mean it as a verb in the sense of "to break ties among supporters", right? Maybe we can find a clearer name for this function. The idea I liked most so far, but am still not fully convinced of, is `break_supporter_ties`.

src/search/heuristics/lm_cut_landmarks.cc
ClemensBuechner:
```suggestion
/*
  NOTE: subgoal can be null if we got here via recursion through
  a zero-cost action that is relaxed unreachable. (This can only
  happen in domains which have zero-cost actions to start with.)
  For example, this happens in pegsol-strips #01.
*/
```
Consider formatting this comment as a multi-line comment (as we usually do) as drive-by cleanup.

ClemensBuechner:
```suggestion
/*
  Using conditional compilation to avoid complaints about unused
  variables when using NDEBUG. This whole code does nothing useful
  when assertions are switched off anyway.
*/
```

ClemensBuechner:
```suggestion
/*
  The following three variables could be declared inside the loop
  ("second_exploration_queue" even inside second_exploration),
  but having them here saves reallocations and hence provides a
  measurable speed boost.
*/
```

src/search/heuristics/lm_cut_landmarks.cc
```diff
@@ -227,6 +234,34 @@ void LandmarkCutLandmarks::second_exploration(
...
+            continue;
+        }
+        if (use_border_detection && has_no_zero_cost_achiever(pre)) {
+            op->h_max_supporter = pre;
+            break;
+        }
+        if (use_goal_zone_detection && pre->status == GOAL_ZONE) {
+            op->h_max_supporter = pre;
+            break;
+        }
```
ClemensBuechner, line 261: This implementation prefers to break ties based on `border_detection` over `goal_zone_detection` if both respective flags are active at the same time. Is there a particular reason for this (theoretical or empirical)? If so, maybe a comment documenting this would be useful. Otherwise, I'd be curious about the difference it makes, but I don't see this critical for merging the changes.

ClemensBuechner (approved):
It is impressive to me how elegant and simple this change is, given its effectiveness in terms of coverage. I certainly approve of the changes. My biggest issue is the function name `tie_break_supporter`, see one of the comments below. I further suggest to do some drive-by cleanup in formatting the code comments, but this is not important at all.

SimonDold:
Since Clemens opened to door for drive-by clean-ups I want to suggest putting not only the paper by Pascal and Max in the synopsis. This relates to [issue1200](https://issues.fast-downward.org/issue1200). In the list there it says we would like to have helmert-domshlak-icaps2009 as an reference.

But this should not hinder merging. Ignore it if it does.

src/search/heuristics/lm_cut_landmarks.cc
PLauerRocks: I wouldn't say it's necessarily opposing. Both were designed with the idea to keep the goal zone small (limit the regression in mark_goal_plateau as much as possible). Goal Zone Detection takes this to an extreme by not exploring nodes that would also be explored otherwise. Border Detection limits this recursion after exploring the next node.

So how I see it is that GZD stops the recursion at depth+0. And BD stops the recursion at depth+1. But, I am happy to discuss this further offline. :) Looks like you have a very different view on this. Maybe we can come up with further ideas.

src/search/heuristics/lm_cut_landmarks.cc
PLauerRocks: Goal Zone Detection allows to limit the exploration to nodes that would also be explored otherwise. At least intuitively this seems like the best thing you can do. If I remember correctly, we also tested both variants and preferring GZD over BD was better. (I believe this was the same thing with all other combinations we tested. Preferring GZD was always better.)

PLauerRocks:
> Since Clemens opened to door for drive-by clean-ups I want to suggest putting not only the paper by Pascal and Max in the synopsis. This relates to [issue1200](https://issues.fast-downward.org/issue1200). In the list there it says we would like to have helmert-domshlak-icaps2009 as an reference.
> 
> But this should not hinder merging. Ignore it if it does.

I will add this to the follow up issue. I was happy to get this reference right without anybody complaining. :)

src/search/heuristics/lm_cut_landmarks.cc
ClemensBuechner: Your explanation addresses my question to a degree that I don't think I have more to contribute. The connection through limiting recursion depth was a missing piece of the puzzle in my head. :)

src/search/heuristics/lm_cut_landmarks.cc
ClemensBuechner: But this implementation prefers BD because that's the if-clause that comes first. So if BD is active and there is something at the border, we will break before going into the GZD. So if we would rather want to prefer GZD, then the order should be swapped!

PR #302, opened 2026-07-28
[issue1231] add early stopping to iterated search.

SimonDold:
Instead of having the user pass a `continue_on_fail` flag for the iterated search, we use the completeness checks to detect whether or not continuing the search makes sense, otherwise we stop early.

Existing configs that call the iterated search with `continue_on_fail=<whatever>` should remove this  part as the keyword argument `continue_on_fail` is no longer known to the iterated search.

### DIFF
--- a/driver/aliases.py
+++ b/driver/aliases.py
@@ -47,7 +47,7 @@ eager(alt([tiebreaking([sum([g(),weight(hAdd,10)]),hAdd]),
            tiebreaking([sum([g(),weight(hgc,10)]),hgc],pref_only=true)],
           boost=500),
       preferred=[hcea,hgc],reopen_closed=true,cost_type=normal)
-],repeat_last=true,continue_on_fail=true))))))"""]
+],repeat_last=true))))))"""]
 
 ALIASES["seq-sat-fd-autotune-2"] = [
     "--search",
@@ -87,7 +87,7 @@ lazy(alt([single(sum([g(),weight(hff,2)])),
           single(sum([g(),weight(hgc,2)]),pref_only=true)],
          boost=1000),
      preferred=[hcea,hgc],reopen_closed=true,cost_type=one)
-],repeat_last=true,continue_on_fail=true)))))"""]
+],repeat_last=true)))))"""]
 
 def _get_lama(pref):
     return [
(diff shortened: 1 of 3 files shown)


driver/aliases.py
```diff
@@ -47,7 +47,7 @@
            tiebreaking([sum([g(),weight(hgc,10)]),hgc],pref_only=true)],
           boost=500),
       preferred=[hcea,hgc],reopen_closed=true,cost_type=normal)
-],repeat_last=true,continue_on_fail=true))))))"""]
```
TravisRiveraPetit, line 50: We now have `continue_on_solve` instead of `continue_on_fail`.
Shouldn't we add `continue_on_solve=true` to the alias?

SimonDold: `continue_on_solve` is a different flag, not a replacement.
It is implicitly set here to its default.

gutodelazeri (commented):
lgtm

roeger:
From looking at the diff I would agree that this can be merged. Did anyone run at least some local experiments to make sure that we did not miss something? The commit message in the squash should include the information what updates need to be made to an existing command line config of an iterated search. If I am not mistaken `continue_on_fail=<whatever>` simply needs to be removed, right?

SimonDold:
What kind of local experiments do you have in mind?
The github actions do some tests and they failed when i forgot to update the aliases
https://github.com/aibasel/downward/actions/runs/30352466050/job/90254838494

Now that I updated the aliases they pass.

roeger:
Well, did you ever run an iterated search with a bound that can trigger the new behavior?

I am also confused to see the early stop included in this issue1231, which in the issue tracker only is "Remove continue_on_fail option in iterated search".

SimonDold:
I checked locally against main.
if you have an unsolvable problem and you run 
`"iterated([astar(cg()),astar(blind()),astar(add())], continue_on_fail=false)"`
it will stop after the search with cg abd report "Search stopped without finding a solution."

on main with
`"iterated([astar(cg()),astar(blind()),astar(add())], continue_on_fail=true)"`
it will run all three and report "Task is provably unsolvable."

in the PR if you run
`"iterated([astar(cg()),astar(blind()),astar(add())])"`
it will only run cg aand blind but not add and reports "Task is provably unsolvable."


in issue1231 we said we want to take care of the comment
```
      Once we can implement this function we also want to change IteratedSearch
      such it automatically stops once a complete search finds no solution.
      In that case the continue_on_fail option would not really be useful
      anymore and should be removed.
```
We remove the flag by adding the early stopping.

roeger:
I meant the aspect where some search returns UNSOLVABLE_WITHIN_BOUND.

SimonDold:
again unsolvable problem, pdb detects the first state as unsolvable:
on main:
`iterated([astar(cg()),astar(blind(), bound=5),astar(pdb())], bound=3, pass_bound=false, continue_on_fail=true)`
does 3 searches "Task is provably unsolvable."
("Task is provably unsolvable within the given bound." if pass_bound=true)

`iterated([astar(cg()),astar(blind(), bound=5),astar(pdb())], bound=3, pass_bound=false, continue_on_fail=false)`
does only the first "Search stopped without finding a solution."
(same if pass_bound=true)

in the PR
`iterated([astar(cg()),astar(blind(), bound=5),astar(pdb())], bound=3, pass_bound=false)`
does the first two and reports "Task is provably unsolvable within the given bound."
(same if pass_bound=true)

SimonDold:
Gabi and I talked offline and she gave her okay to merge.

PR #303, opened 2026-07-28
[issue1233] Report unsolvability for bounded search algos if the bound was not used.

SimonDold:


TravisRiveraPetit (approved):

roeger:
I see one disadvantage of this, namely that this functionality is easy to miss when one implements a new search algorithm, which can lead to wrong return codes. Alternatively, `bound_was_used` could be initialized to `true` in the constructor of `SearchAlgorithm`, and all algorithms who correctly use it need to update it to `false` in their constructor. Not beautiful, but safer. What do others think?

roeger:
Don't you also have to support the flag in the iterated search? I think currently you could wrongly report UNSOLVABLE.

SimonDold:
I agree with your safety concerns.
I am happy with your suggestion.

Another option could be to have a function like 
```
bool SearchAlgorithm::is_beyond_bound(int x) {
    if (x >= bound) { // greater or equal because the bound is given exclusively
          bound_was_used=true;
          return true;
    }
    return false;
}
```

And use this function for the checks with the bound.

We can make the field `bound` private in the base class and remove the getter and setter for it. Only iterated search uses this and there we could deal with a function like
```
void SearchAlgorithm::shrink_bound(int x) {
    if (x < bound) {
        bound = x;
    }
}
``` 
and replace in the iterated search
```
    if (pass_bound && best_bound < task_specific_search->get_bound()) {
        task_specific_search->set_bound(best_bound);
    }
```
with
```
if (pass_bound) {
    task_specific_search->shrink_bound(best_bound);
}
```

The only remaining case for the field `bound` to be used is in the log strings. 
Like 
```
    log << "Conducting best first search"
        << (reopen_closed_nodes ? " with" : " without")
        << " reopening closed nodes, (real) bound = " << bound << endl;
```
But for that we could provide a public or protected function like
```
string SearchAlgorithm::get_bound_string() {
     return to_string(bound);  //fantasy syntax
}
```

Sure you could read the number out of the string to prune based on the bound without updating that the bound was used :D but that would be noticed in a code review.

SimonDold:
Yes, you are right. This is already a great example for your safety concerns.

here we would have to update
```
void IteratedSearch::update_best_status(task_specific_search) {
//...
        if (task_specific_search->get_bound() >= bound) {
            best_status = current_search_status;
        }
//...
```

This points out a mistake in my earlier comment, where i said `get_bound()` was only used at one place (i searched in the wrong local branch) for this we could have a function like
```
int SearchAlgorithm::get_final_bound() {
    if (status == SearchStatus::IN_PROGRESS) {
        ERROR
    }
    return bound;
}
```

and change the update of the best status to
```
void IteratedSearch::update_best_status(task_specific_search) {
//...
        if (is_beyond_bound(task_specific_search->get_final_bound())) {
            best_status = current_search_status;
        }
//...
```

otherwise it should be 
 ```
void IteratedSearch::update_best_status(task_specific_search) {
//...
        if (ask_specific_search->get_bound() >= bound) {
            bound_was_used = true;
            best_status = current_search_status;
        }
//...
```

**Comment by roeger** [comment]
@SimonDold Can you just add the proposed changes to the PR, so that we can see them as a diff? Regarding the proposed method `is_beyond_bound`, I would use a name that makes the test very clear (whether it is x > bound or x >= bound). Maybe something like `is_greater_or_equal_to_bound`?

File: src/search/search_algorithm.h [hunk 1]
```diff
@@ -65,6 +66,8 @@ class SearchAlgorithm : public components::TaskSpecificComponent {
     void set_plan(const Plan &plan);
     bool check_goal_and_set_plan(const State &state);
     int get_adjusted_cost(const OperatorProxy &op) const;
+    // Sets bound_was_used to true if this check returns true.
```
grucla:
```suggestion
    // Set bound_was_used to true if this check returns true.
```

src/search/search_algorithm.h
```diff
@@ -53,10 +54,10 @@ class SearchAlgorithm : public components::TaskSpecificComponent {
     SearchSpace search_space;
     SearchProgress search_progress;
     SearchStatistics statistics;
-    int bound;
     OperatorCost cost_type;
     bool is_unit_cost;
     double max_time;
+    bool bound_was_used;
```
grucla: Without context the meaning of `bound_was_used` was not clear to me (how do you "use" a bound?). Maybe rename it to something like `found_g_value_larger_or_equal_to_bound`, or `went_beyond_bound`? We could also invert it and call it something like `remained_within_bound`, or `remained_below_bound`.

src/search/search_algorithms/iterated_search.cc
```diff
@@ -147,8 +148,8 @@ SearchStatus IteratedSearch::step() {
     if (!task_specific_search) {
         return best_status;
     }
-    if (pass_bound && best_bound < task_specific_search->get_bound()) {
-        task_specific_search->set_bound(best_bound);
+    if (pass_bound) {
+        task_specific_search->shrink_bound(best_bound);
```
grucla, line 152: `best_bound` was not introduced by this PR but I tripped over it here because I forgot when one bound is "better" / "best" compared to another bound. Maybe we could rename it to something like `lowest_known_bound` as a drive-by improvement?

grucla (commented):
The changes look good to me. I think some renamings / rephrasings would be nice (see comments) but if others prefer the current names I'm fine with merging it as is.

src/search/search_algorithm.h
roeger: @grucla Would `bound_was_used_for_pruning` be clearer? Your suggestions are very specific and do not fit all search algorithms.

grucla: I do not understand how for example `went_beyond_bound` is more specific than `bound_was_used_for_pruning`. As far as I understand the flag can also be set to true in `IteratedSearch::update_best_status` where no state is pruned.

roeger: `went_beyond_bound` is very unclear to me. What went beyond the bound? If you do not like my earlier suggestion, can I win your support for `bound_was_relevant`?

SimonDold: Claudia and i talked offline, we are happy with `bound_was_relevant`.

src/search/search_algorithms/iterated_search.cc
SimonDold: Claudia and I talked offline and said that it is hard to find a good name.
One view is that this int represents the lowest know plan cost, but this does not match the variable on initialization.

It would maybe be something like "the upper bound to plan costs we still care about". We decided to keep it as is for now.

roeger:
After some consideration, I think that the implementation works correctly for the iterated search (medium confidence). I also like the safer design after the revision (and just am confident that no-one will extract the bound from the string). If everyone else is happy, this can be merged from my side.

roeger:
Ah, hm, we did not run any experiments for this PR, right? I wonder whether having a call for the bound test for every generated state makes a difference for the runtime.

PR #305, opened 2026-07-29
issue1201: Handle new exit code in portfolios.

grucla:
issue1201 added the new exit code `UNSOLVABLE_WITHIN_BOUND` that requires separate handling when running a portfolio.

### DIFF
--- a/driver/portfolio_runner.py
+++ b/driver/portfolio_runner.py
@@ -117,7 +117,10 @@ def run_sat(configs, executable, sas_file, plan_manager, final_config,
                 continue
 
             yield exitcode
-            if exitcode == returncodes.SEARCH_UNSOLVABLE:
+            if (exitcode == returncodes.SEARCH_UNSOLVABLE or
+                exitcode == returncodes.SEARCH_UNSOLVABLE_WITHIN_BOUND):
+                # SEARCH_UNSOLVABLE and SEARCH_UNSOLVABLE_WITHIN_BOUND are
+                # equivalent because the bound never increases.
                 return
 
             if exitcode == returncodes.SUCCESS:
@@ -137,7 +140,10 @@ def run_sat(configs, executable, sas_file, plan_manager, final_config,
                         return
 
                     yield exitcode
-                    if exitcode == returncodes.SEARCH_UNSOLVABLE:
+                    if (exitcode == returncodes.SEARCH_UNSOLVABLE or
+                        exitcode == returncodes.SEARCH_UNSOLVABLE_WITHIN_BOUND):
+                        # SEARCH_UNSOLVABLE and SEARCH_UNSOLVABLE_WITHIN_BOUND
+                        # are equivalent because the bound never increases.
                         return
                 if final_config_builder:
                     print("Build final config.")
@@ -168,6 +174,9 @@ def run_opt(configs, executable, sas_file, plan_manager, timeout, memory):
         exitcode = run_search(executable, args, sas_file, plan_manager,
                               run_time, memory)
         yield exitcode
+
+        assert exitcode != returncodes.SEARCH_UNSOLVABLE_WITHIN_BOUND
+        # Portfolios for optimal planning should not be used with bounds.
 
         if exitcode in [returncodes.SUCCESS, returncodes.SEARCH_UNSOLVABLE]:
             break
--- a/driver/returncodes.py
+++ b/driver/returncodes.py
@@ -63,6 +63,7 @@ def generate_portfolio_exitcode(exitcodes):
     There are multiple types of unexpected exit codes -> SEARCH_CRITICAL_ERROR.
     [..., SUCCESS, ...] -> SUCCESS
     [..., SEARCH_UNSOLVABLE, ...] -> SEARCH_UNSOLVABLE
+    [..., SEARCH_UNSOLVABLE_WITHIN_BOUND, ...] -> SEARCH_UNSOLVABLE_WITHIN_BOUND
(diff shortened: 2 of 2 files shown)


driver/portfolio_runner.py
```diff
@@ -137,7 +140,10 @@ def run_sat(configs, executable, sas_file, plan_manager, final_config,
                         return
 
                     yield exitcode
-                    if exitcode == returncodes.SEARCH_UNSOLVABLE:
+                    if (exitcode == returncodes.SEARCH_UNSOLVABLE or
+                        exitcode == returncodes.SEARCH_UNSOLVABLE_WITHIN_BOUND):
+                        # SEARCH_UNSOLVABLE and SEARCH_UNSOLVABLE_WITHIN_BOUND
+                        # are equivalent because the bound never increases.
```
roeger, line 146: Same as above. Do not return (stop everything) on SEARCH_UNSOLVABLE_INCOMPLETE because another portfolio component might still find a plan.

driver/portfolio_runner.py
```diff
@@ -117,7 +117,10 @@ def run_sat(configs, executable, sas_file, plan_manager, final_config,
                 continue
 
             yield exitcode
-            if exitcode == returncodes.SEARCH_UNSOLVABLE:
+            if (exitcode == returncodes.SEARCH_UNSOLVABLE or
+                exitcode == returncodes.SEARCH_UNSOLVABLE_WITHIN_BOUND):
+                # SEARCH_UNSOLVABLE and SEARCH_UNSOLVABLE_WITHIN_BOUND are
+                # equivalent because the bound never increases.
```
roeger, line 123: Couldn't it be the case that a portfolio has a component that runs with a bound and other components don't? This would be analogue to SEARCH_UNSOLVABLE_INCOMPLETE (another component might find a plan), not to SEARCH_UNSOLVABLE (there is not hope that another component finds a plan). If you revert this change, the code will do the right thing (continuing, but not adding the config for the next round).

roeger (changes requested):

driver/portfolio_runner.py
roeger: The "above" means the comment on lines 120ff.

roeger:
Since at the end there is almost no change necessary for the portfolios (except for the one assertion), I would rather call this "Handle new exit code in driver" in the squash commit message.

driver/portfolio_runner.py
roeger: Ah, I see that the portfolios always use decreasing bounds that are managed by the portfolio_runner. Then this makes sense.

driver/portfolio_runner.py
roeger: Ah, sorry, I mistyped INCOMPLETE when I meant UNSOLVABLE_WITHIN_BOUND. But I think this is resolved together with the other comment in this file.

driver/portfolio_runner.py
grucla: Just wanted to write an explanation for this but you already found out on your own.
So, this means that you agree with the current state of the pull request?

roeger:
It took me some time to understand that the code in `generate_portfolio_exitcode` makes sense. The position of `SEARCH_UNSOLVABLE_WITHIN_BOUND` within the list is crucial and correct (very subtle!). The documentation of this function is somewhere between cryptic and misleading, but this is nothing that has been newly introduced in this PR. So from my side, this can be merged.

grucla:
I also only really understood the documentation after I understood the code. What do you think about changing the first sentence "A portfolio's exitcode is determined as follows:" to the following as a drive-by improvement?
"A portfolio's exitcode is determined by applying the following rules in that order:"

roeger:
> I also only really understood the documentation after I understood the code. What do you think about changing the first sentence "A portfolio's exitcode is determined as follows:" to the following as a drive-by improvement? "A portfolio's exitcode is determined by applying the following rules in that order:"

It still does not quite fit because of the different variants of success. This would require three more lines.

grucla:
How about this?

```
A portfolio's exitcode is determined by applying the following rules in that order:

There is exactly one type of unexpected exit code -> use it.
There are multiple types of unexpected exit codes -> SEARCH_CRITICAL_ERROR.

SUCCESS, SEARCH_OUT_OF_MEMORY, and SEARCH_OUT_OF_TIME in exitcodes -> SEARCH_PLAN_FOUND_AND_OUT_OF_MEMORY_AND_TIME
SUCCESS and SEARCH_OUT_OF_MEMORY in exitcodes -> SEARCH_PLAN_FOUND_AND_OUT_OF_MEMORY
SUCCESS and SEARCH_OUT_OF_TIME in exitcodes -> SEARCH_PLAN_FOUND_AND_OUT_OF_TIME
SUCCESS in exitcodes -> SUCCESS

SEARCH_UNSOLVABLE in exitcodes -> SEARCH_UNSOLVABLE
SEARCH_UNSOLVABLE_WITHIN_BOUND in exitcodes -> SEARCH_UNSOLVABLE_WITHIN_BOUND
SEARCH_UNSOLVED_INCOMPLETE in exitcodes -> SEARCH_UNSOLVED_INCOMPLETE

SEARCH_OUT_OF_MEMORY and SEARCH_OUT_OF_TIME in exitcodes -> SEARCH_OUT_OF_MEMORY_AND_TIME
SEARCH_OUT_OF_TIME in exitcodes -> SEARCH_OUT_OF_TIME
SEARCH_OUT_OF_MEMORY in exitcodes -> SEARCH_OUT_OF_MEMORY
```

roeger:
Much clearer!

PR #308, opened 2026-08-05
[issue1240] Keep axioms stratifiable in normalization edge case

grucla:
The normalization of the translator makes the axioms of a task unstratifiable if they contain an axiom that mentions its head predicate, possibly recursively, in its body under the scope of a universal quantifier. We fix this by using a different universal quantifier elimination for such axioms.


grucla:
I ran some experiments where the output of the translator unexpectedly changed on some domains although they are no such edge cases. The names of some variables introduced by the translator change in output.sas. This is avoidable but would make the code less clear which is why I did not implement it. See my comment on the issue tracker (issue1240) for details.

PR #310, opened 2026-08-12
Reduce unused code

SimonDold:
> **This is a Placeholder**  
Replace the content of this field with a meaningful description of the changes made within this pull request. It will be suggested as the commit message when merging the pull request.

For the pull request title, summarize the changes in imperative mood and prepend "[issueX]" where X is the issue number in our issue tracker (issues.fast-downward.org). The title will be suggested as the commit subject.

The commit message (i.e., this field) should highlight how the changes affect planner behavior:
 - Do they add functionality?
 - Does the command line syntax change?
 - Is planner performance affected according to the experiment?

Find our guidelines for commit messages in the Fast Downward Wiki (https://www.fast-downward.org/latest/for-developers/git/#git_conventions).

### DIFF
--- a/src/search/landmarks/landmark_factory_reasonable_orders_hps.cc
+++ b/src/search/landmarks/landmark_factory_reasonable_orders_hps.cc
@@ -112,7 +112,6 @@ void LandmarkFactoryReasonableOrdersHPS::insert_reasonable_orderings(
 */
 void LandmarkFactoryReasonableOrdersHPS::approximate_reasonable_orderings(
     const TaskProxy &task_proxy) {
-    State initial_state = task_proxy.get_initial_state();
     for (const auto &node : *landmark_graph) {
         const Landmark &landmark = node->get_landmark();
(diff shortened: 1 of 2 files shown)


maltehelmert:
Was this created by accident? You did not replace the placeholder with the required content.

Regarding the pull request, the forward declarations are not "unused code" but there intentionally. We thought it was clearer to forward-declare everything but think about what does or doesn't need forward declarations here. It also serves as an overview of the large collection of classes that work in a coordinated way. So this part is not something we should merge.

The "get_initial_state" looks like it should go (but I didn't check context thoroughly).

SimonDold:
I put it as a draft to discuss it at some point. Should have been a PR to my own clone though....
But your reaction already answered the questions I was going to ask.

For the  "get_initial_state" the mechanic is that it calls a simple getter of the AbstractTask, to  create a state with it via the "create_state" method, which calls the constructor of the State class.
There are no side effects beside some asserts and building smart pointers to the vector that stores the state values, which should be deleted right after.

